PR #3, opened 2016-03-15
Opus 2015 3dms

jhunkeler:
Implement change from 2015.3


jhunkeler:
@stsci-hack This merge needs help

PR #9, opened 2016-08-15
Deprecate astrolib

jhunkeler:
Bonus: Introduces a `.gitignore` to help prevent accidental inclusion of generated files.


pllim:
About time!

PR #11, opened 2016-11-12
Document formula for sigma when combining with 'minmed'

mcara:


mcara:
@stsci-hack Please review changes

mcara:
Due to formatting differences it is difficult to see actual change made to the text. Therefore, I am posting bellow the only non-formatting difference:

```
The value of sigma is computed as :math:`\sigma = \sqrt(D + S + R^2)`,
where *D* is image data, *S* is the value of the subtracted sky, and
*R* is the value of the readout noise.
``

**Comment by philhodge** [comment]
This assumes that the gain is 1. Is that the case?

**Comment by mcara** [comment]
@philhodge Hm, let me dig dipper... Thanks!

**Comment by mcara** [comment]
@philhodge @stsci-hack So, I have modified previous statement to include specification for units (which are electrons so gain is already taken into account):

```
    The value of sigma is computed as :math:`\sigma = \sqrt(M + S + R^2)`,
    where *M* is the median image data (in electrons), *S* is the value of the
    subtracted sky (in electrons), and *R* is the value of the readout noise
    (in electrons).
```

Issue #12, opened 2016-11-14
Subtracted sky is not converted to electrons when computing sigma for combine_type "minmed"


mcara:
Based on my analysis of the code (done when working on https://github.com/spacetelescope/drizzlepac/pull/11), it *seems* that ``createMedian`` step does not convert subtracted sky value to electrons (rather it is left in its original units, i.e., ``e/s``, ``counts/s``, ``counts``, ``e`` - so sometimes - when input FLT images are already in electrons - the formula for computing sigma works correctly).

See: https://github.com/spacetelescope/drizzlepac/blob/master/lib/drizzlepac/createMedian.py#L265 
and
https://github.com/spacetelescope/drizzlepac/blob/master/lib/drizzlepac/minmed.py#L178

Subtracted sky value comes directly from the value of ``MDRIZSKY`` in input image headers:
https://github.com/spacetelescope/drizzlepac/blob/master/lib/drizzlepac/sky.py#L354

mcara:
Solved by https://github.com/spacetelescope/drizzlepac/pull/13

PR #13, opened 2016-11-15
Convert sky background from native input units to electrons for 'minmed' combine type

mcara:
This PR fixes the issue described in https://github.com/spacetelescope/drizzlepac/issues/12 due to which sky background values are not converted to electrons from native units (as recorded in input image headers) leading to an underestimation of sigma that results in a more aggressive cosmic ray rejection, possibly flagging many good pixels as cosmic rays.

@stsci-hack


stsci-hack:
If this code actually runs as expected, this PR can be merged.

Issue #23, opened 2017-01-24
TypeError: buffer is too small for requested array


pllim:
Example test: https://ssb.stsci.edu/pandokia/pandokia.cgi?query=detail&key_id=36462995

```
Trailer file written to:  stis_01_1visit

  File "/Users/shared/iraf_conda/bldtmp/porcelain.phEUcEl55g/porcelain/envs/dev/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/Users/shared/iraf_conda/bldtmp/porcelain.phEUcEl55g/porcelain/envs/dev/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/iraf/rt/betadrizzle/stis/stis_01/stis_01.py", line 72, in stis_01
    astrodrizzle.AstroDrizzle('o*flt.fits',runfile=testname,configobj=parObj)
  File "/Users/shared/iraf_conda/bldtmp/porcelain.phEUcEl55g/porcelain/envs/dev/lib/python2.7/site-packages/drizzlepac/astrodrizzle.py", line 123, in AstroDrizzle
    run(configObj, wcsmap=wcsmap)
  File "/Users/shared/iraf_conda/bldtmp/porcelain.phEUcEl55g/porcelain/envs/dev/lib/python2.7/site-packages/drizzlepac/util.py", line 229, in wrapper
    raise errorobj
TypeError('buffer is too small for requested array',)
```

mcara:
I have investigated the cause of this crash and it is due to:
1. a previous crash left a corrupted ``STISFUV-MAMA_1024x1024_1_staticMask.fits`` file in the test directory (it has too little data compared to what is expected);
2. a combination of software changes that made the code susceptible to the presence of corrupt static mask files in the working directory: AND
   * In the [Changeset 7537](https://trac.stsci.edu/ssb/stsci_python/changeset/7537/development/trunk/BigBlackBox/lib/staticMask.py) code was made to overwrite existing static mask files *only* ``if not(fileutil.checkFileExists(filename))``, that is, never (CC: @sosey ).
   * In the [Changeset 32128](https://trac.stsci.edu/ssb/stsci_python/changeset/32128/drizzlepac/trunk/lib/drizzlepac/staticMask.py) in which I have removed randomization of static mask file names (at INS request). This change in itself is not harmful except when combined with previously mentioned code that does not permit overwriting existing mask files (an important detail that I missed).

[...]

PR #24, opened 2017-02-02
Allow existing static masks to be overwritten

mcara:
This PR modifies code so that if there is an existing static mask in the working directory, this mask will be overwritten when code runs. This addresses the issue https://github.com/spacetelescope/drizzlepac/issues/23


mcara:
@stsci-hack @sosey Please review this PR.

PR #26, opened 2017-02-02
Implement local rotateCD

mcara:
Due to change of the definition of ``rotateCD`` made in https://github.com/astropy/astropy/pull/5189 as well as its deprecation in https://github.com/astropy/astropy/pull/5240, it is necessary that our code has its own ``rotateCD`` that implements previous (to PR https://github.com/astropy/astropy/pull/5189) version of the code.

@stsci-hack @nden please review this PR

### DIFF
--- a/lib/drizzlepac/wcs_functions.py
+++ b/lib/drizzlepac/wcs_functions.py
@@ -521,6 +521,16 @@ def restoreDefaultWCS(imageObjectList, output_wcs):
     updateImageWCS(imageObjectList, output_wcs)
 
 
+def _rotateCD(cd, theta):
+    # This is the old (pre-astropy PR #5189 -
+    # see https://github.com/astropy/astropy/pull/5189) as this the version
+    # of the rotateCD expected in 'mergeWCS' below
+    theta = np.deg2rad(theta)
+    cth = np.cos(theta)
+    sth = np.sin(theta)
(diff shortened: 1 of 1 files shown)


nden:
This should recover the original state from before the change in astropy. (BTW, there is a statement in the astropy discussion that our code does not use this method so changing it won't affect anything.)
:+1: for merging especcialy if it explains some of the failing pandokia tests.

nden:
Speaking of tests apparently Travis runs on drizzlepac without having a formal `.travis.yml` file or any tests. This had me baffled for a while until I figured out why. But this is obviously a different discussion.

mcara:
Reply to @nden https://github.com/spacetelescope/drizzlepac/pull/26#issuecomment-277273065

> BTW, there is a statement in the astropy discussion that our code does not use this method so changing it won't affect anything.

I know. I missed this somehow... Sad.

> 👍 for merging especcialy if it explains some of the failing pandokia tests.

It seems to not affect tests at all. It seems that the CD matrix that is being rotated is diagonal so that CD==CD.T and therefore this makes no difference in the current tests (but maybe our tests do not test all possible situations so this may still affect something). Regardless, my PR implements a "local" version of ``rotateCD`` identical (I hope=>please verify) to what was before.

Issue #15, opened 2017-01-24
AstropyDeprecationWarning: "clobber" was deprecated in version 1.3


pllim:
Example test: https://ssb.stsci.edu/pandokia/pandokia.cgi?query=detail&key_id=36464376

```
[WARNING ] /srv/iraf/bldtmp/porcelain.g23AA24Tm8/porcelain/envs/dev/lib/python2.7/site-packages/astropy/utils/decorators.py:470:
AstropyDeprecationWarning: "clobber" was deprecated in version 1.3
and will be removed in a future version. Use argument "overwrite" instead.
  AstropyDeprecationWarning)
```

mcara:
While some of the warnings are coming from ``drizzlepac`` itself, others are coming from multiple dependent packages such as ``stsci.tools``, ``skymatch``, etc. I will start updating these packages and I will post back here each change.

mcara:
So, I have created several PRs addressing the problem of deprecated ``clobber`` parameter in ``astropy.io.fits``:

- ``stsci.tools``: https://github.com/spacetelescope/stsci.tools/pull/20
- ``drizzlepac``: https://github.com/spacetelescope/drizzlepac/pull/28
- ``stsci.skypac``: https://github.com/spacetelescope/stsci.skypac/pull/2

@nden @stsci-hack However, the ``headerlet`` part of [``stwcs``](https://github.com/spacetelescope/stwcs) package needs to be updated as well as ``drizzlepac`` uses the headerlet facility. I have opened an issue https://github.com/spacetelescope/stwcs/issues/23 on ``stwcs``.

pllim:
`stsci.tools`: It is merged and pushed to AstroConda via astroconda/astroconda-contrib#159

mcara:
``stsci.skypac`` merged in https://github.com/spacetelescope/stsci.skypac/pull/2
``drizzlepac`` merged in https://github.com/spacetelescope/drizzlepac/pull/28

Only one more package left: https://github.com/spacetelescope/stwcs/issues/23

mcara:
STWCS was updated in https://github.com/spacetelescope/stwcs/pull/24
Closing this issue.

PR #28, opened 2017-02-03
Fix deprecated 'clobber' parameter in astropy.io.fits write functions

mcara:
The ``clobber`` parameter in ``astropy.io.fits`` write functions has been deprecated and it has been replaced with ``overwrite``: see https://github.com/astropy/astropy/pull/5171.

This has caused our code (see, e.g., ``drizzlepac`` - https://github.com/spacetelescope/drizzlepac/issues/15) to issue numerous deprecation warnings.

This PR updates the code in ``stsci.tools`` so that it uses ``overwrite`` for versions of astropy  >= 1.3 and ``clobber`` for verions < 1.3 (for backward compatibility).

This PR addresses issue described in https://github.com/spacetelescope/drizzlepac/issues/15

@stsci-hack Please review this PR.

### DIFF
--- a/lib/drizzlepac/catalogs.py
+++ b/lib/drizzlepac/catalogs.py
@@ -10,6 +10,7 @@ import copy
 
 import numpy as np
 #import pywcs
+import astropy
 from astropy import wcs as pywcs
 import astropy.coordinates as coords
 from astropy import units as u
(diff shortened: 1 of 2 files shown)


mcara:
Even though there was no comment on this PR, a similar PR (see spacetelescope/stsci.tools#20) has had extensive comments and there were no objections to this way of handling the clobber issue => merging this PR.

Issue #22, opened 2017-01-24
RuntimeWarning: invalid value encountered


pllim:
Example test: https://ssb.stsci.edu/pandokia/pandokia.cgi?query=detail&key_id=36464242

```
[WARNING ] /srv/iraf/bldtmp/porcelain.g23AA24Tm8/porcelain/envs/dev/lib/python2.7/site-packages/drizzlepac/minmed.py:220:
RuntimeWarning: invalid value encountered in sqrt
  __rms_file = np.sqrt(__median_file_weighted + __bkgd_file + __readnoise_file)

[WARNING ] /srv/iraf/bldtmp/porcelain.g23AA24Tm8/porcelain/envs/dev/lib/python2.7/site-packages/drizzlepac/minmed.py:247:
RuntimeWarning: invalid value encountered in less
  __minimum_flag_file = np.where(np.less(__minimum_file_weighted,__median_rms_file),1,0)

[WARNING ] /srv/iraf/bldtmp/porcelain.g23AA24Tm8/porcelain/envs/dev/lib/python2.7/site-packages/drizzlepac/minmed.py:301:
RuntimeWarning: invalid value encountered in less
  self.combArrObj = np.where(np.less(__minimum_file_weighted,__median_rms2_file),
```

pllim:
Seems to have disappeared as of Feb 5, on dev.

pllim:
Ops, sorry for the noise. It is still there.

mcara:
The funny part is that this issue was foreseen since the beginning: https://github.com/spacetelescope/drizzlepac/blob/master/lib/drizzlepac/minmed.py#L219

mcara:
I created a PR that fixes this issue: https://github.com/spacetelescope/drizzlepac/pull/33

However, because I expect this PR to break multiple tests, I would want to have existing failing tests to be OKified first before this PR can be merged.

CC: @stsci-hack

mcara:
Solved by https://github.com/spacetelescope/drizzlepac/pull/33

PR #33, opened 2017-02-07
Zero negative values before computing image's Poisson errors

mcara:
This PR solves the issue of negative values under square root that produce warnings reported in https://github.com/spacetelescope/drizzlepac/issues/22. Fundamentally, the code now will set to zero any negative data values before computing Poisson noise.

This PR also removed unnecessary double underscores from variable names + removes extra new lines reported in https://github.com/spacetelescope/stsci.tools/issues/22#issuecomment-277808856


mcara:
@stsci-hack I expect this PR to break multiple tests. Because we already have a couple of tests that are failing for a long time, we need to OKify those first before this PR should be merged. Otherwise it will be even more difficult to disentangle the causes of failures.

Also, @stsci-hack please review this PR.

Issue #31, opened 2017-02-06
Make new releases


pllim:
New releases of `drizzlepac` and its affected dependencies are needed to propagate the fixes of deprecation warnings to our public build on AstroConda. (So, really the process is GitHub releases + AstroConda recipe updates.)

EDIT: Deprecation warning is not as urgent as critical bug fix, so for packages with only fixes to these warnings, the release can wait but should be done before the deprecated codes are removed completely from their dependencies (e.g., Astropy).

p.s. If you maintain PyPi etc, that needs updating too but I don't care about that.

larrybradley:
Please include `fitsblender` in the new releases to include a fix I made there.

mcara:
I do not understand the need for this "issue". Of course we will make a release once some other issues are taken care of. at this moment I am not ready to make a release.

@larrybradley @jhunkeler Isn't each package "independent", at least with regard to releases? Why realeasing of ``fitsblender`` is even mentioned here?

larrybradley:
I mentioned it here only ecause it's an "affected dependency" and if you're going to make new releases to include it.  I would do it myself, but I don't know if you have any other pending changes.

pllim:
If `fitsblender` is used by `drizzlepac` and will cause the latter to emit such warning, then it is counted as one of the "affected dependencies". I opened this issue because there are so many packages involved, and I am loosing track, so this serves as a reminder.

mcara:
@larrybradley 
> ... I would do it myself, but I don't know if you have any other pending changes.
I do not have any pending changes for ``fitsblender``.

@pllim 
> ... I am loosing track, so this serves as a reminder.
I promise not to forget all the packages affected by recent changes (while working on latest ~2 weeks ago issues).

mcara:
@jhunkeler Please make releases for:
stsci.convolve
stsci.skypac
stsci.tools (maybe pending your investigation of https://github.com/spacetelescope/stsci.tools/issues/22)
stwcs (if @nden agrees)
pysynphot
reftools
fitsblender

Please hold off releasing ``drizzlepac``: I am still working on two issues (unless @pllim add more as we speak 😄 ).

pllim:
EDIT: I cross out pysynphot in the list above. The changes made has no effect on user-facing API.

nden:
Are we supposed to make a public release after every change in master now? There's one change in stwcs since the last release and it's to hide this DeprecationWarning, it's not even a bug fix release.
I don't understand ...

[...]

PR #37, opened 2017-02-09
Bumped version to 2.1.8

jhunkeler:
See also: https://github.com/spacetelescope/drizzlepac/issues/31#issuecomment-278527868


mcara:
@jhunkeler Thanks!

Issue #51, opened 2017-03-28
Error message has bad syntax


rjavila:
A user was trying to run some WFPC2 images through astrodrizzle and got the message below. They complained that the syntax in the error message was wrong (missing quotes in the updatewcs call). The message should also be updated to use astropy.io.fits instead of iraf tasks. Also, typing 'c' gives you a message that 'c' is not defined. 

```
A 'DGEOFILE' keyword is present in the primary header but 'NPOLFILE' keyword was not found.
            This version of the software uses a new format for the residual distortion DGEO files.
            Please consult the instrument web pages for which reference files to download.
            A small (new style) dgeofile is needed ('_npl.fits' extension) and possibly a
            detector to image correction file ('_d2i.fits' extension).
            The names of these files must be added to the primary header either using the task XXXX
            or manually, for example:

            hedit u33z0401t_c0m.fits[0] npolfile fname_npl.fits add+
            hedit u33z0401t_c0m.fits[0] d2imfile fname_d2i.fits add+

            where fname_npl.fits is the name of the new style dgeo file and fname_d2i.fits is
            the name of the detector to image correction. After adding these keywords to the
            primary header, updatewcs must be run to update the science files:

            from stwcs import updatewcs
            updatewcs.updatewcs(u33z0401t_c0m.fits)

            Alternatively you may choose to run astrodrizzle without DGEO and detector to image correction.

            To stop astrodrizzle and update the dgeo files, type 'q'.
            To continue running astrodrizzle without the non-polynomial distortion correction, type 'c':
```

mcara:
@rjavila This problem should be fixed in https://github.com/spacetelescope/drizzlepac/pull/53

If this is not an emergency, the fix will be made available with the next release of the ``drizzlepac``.

Thanks for reporting the problem.

mcara:
A bug-fix release for drizzlepac was made today and should be available for download tomorrow. Closing this issue...

PR #54, opened 2017-03-30
Fix incorrect indent introduced when fixing the 'input' bug

mcara:


mcara:
This fixes a bug (incorrect indent) introduced in https://github.com/spacetelescope/drizzlepac/pull/53 when fixing https://github.com/spacetelescope/drizzlepac/issues/51

PR #55, opened 2017-04-04
Fix multiple bugs in 'getflat' method of 'STISInputImage'

mcara:
Multiple bugs are affecting the ``getflat()`` method of the ``STISInputImage`` class preventing processing of STIS imaging data.


mcara:
CC: @stsci-hack

PR #60, opened 2017-04-27
Suppress info messages

nden:
This supresses log.INFO messages comping from astropy.wcs in cases which I believe are benign and are safe to ignore. My local testing shows it produces a clean log as desired by some.


nden:
@stsci-hack This is ready for a review.

stsci-hack:
These changes seem to be very local and specifically address the issues we have been seeing without adding any undue changes on the rest of the code.  This safe set of changes also highlights, and works around, the unusual logging implemented for drizzlepac.  As a result, I see no reason to hold up this merge, even if there are still concerns about what is triggering the excessive messages in the first place.

Issue #81, opened 2017-08-28
TweakReg is ignoring units of user-provided catalogs


mcara:
TweakReg is ignoring units (``xyunits`` parameter) of user-provided (through the ``catfile`` parameter) catalogs. Specifically, user-provided catalogs are assumed to be in "pixels" even when user sets ``xyunits`` to ``"degrees"``.

Even though the support for catalogs in degrees is documented in ``TweakReg``'s documentation, the support for degrees has never been implemented.

This bug was first reported by Will Waldron (UAH).

mcara:
Fixed in https://github.com/spacetelescope/drizzlepac/pull/82

PR #82, opened 2017-08-28
Add support for user-provided image catalogs in degrees

mcara:
TweakReg is ignoring units of user-provided catalogs. Specifically, user-provided catalogs are assumed to be in "pixels" **regardless** of the value of the ``xyunits`` parameters.

This PR adds support for image catalogs with coordinates in degrees.

This bug was first reported by Will Waldron (UAH) who also proposed a solution partially implemented in this PR.

This PR addresses issue https://github.com/spacetelescope/drizzlepac/issues/81

### DIFF
--- a/doc/source/release_notes.rst
+++ b/doc/source/release_notes.rst
@@ -17,13 +17,15 @@ of the list).
 Drizzlepac v2.1.18 (not released)
 =================================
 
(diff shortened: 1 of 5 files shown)


mcara:
@stsci-hack Thank you! I just need to update version date before I make a release.

Issue #94, opened 2017-10-31
Possible logical/computational error in createMedia.py


mcara:
CC: @stsci-hack 

The code in a couple of `IF` blocks is never executed and so I wonder if it should be removed or whether this is a mistake and for some "units" the *desired* code is not executed... See https://github.com/spacetelescope/drizzlepac/blob/master/lib/drizzlepac/createMedian.py#L169:

```python
        if lthresh is not None:
            if proc_units.lower() == 'native':
                if native_units.lower() == "counts":
                    lthresh = lthresh * det_gain
                    if native_units.lower() == "counts/s":
                        lthresh = lthresh * img_exptime
        if hthresh is not None:
            if proc_units.lower() == 'native':
                if native_units.lower() == "counts":
                    hthresh = hthresh * det_gain
                    if native_units.lower() == "counts/s":
                        hthresh = hthresh * img_exptime
```
Code that follows `if native_units.lower() == "counts/s"` is never executed because `native_units` cannot be both `"counts"` AND `"counts/s"` at the same time. I wonder what was the intention here and what is the correct thing to do...

mcara:
Yeah, the more I look at this code the more it seems that this is a bug. When user specifies either `combine_lthresh` or `combine_hthresh` in the `CREATE MEDIAN IMAGE` step, these limits/thresholds are not converted correctly to electrons (processing unit).

1. When `native_units` are `'COUNTS/S'` threshold does not get converted to electrons because of indentation problem in the code;
2. When `native_units` are `'ELECTRONS/S'` (e.g., WFC3) threshold does not get converted to electrons because this option is missing.

stsci-hack:
This error would only impact WFPC2, STIS and NICMOS data, as both ACS and WFC3 have native_units of ELECTRONS or ELECTRONS/S.  However, I believe this logic should be revised to un-indent the second if-statement in each block to replicate if-elif logic.

mcara:
@stsci-hack What about WFC3 when `native_units` are `'ELECTRONS/S'`? This code appears to skip converting  `e/s` to `e`.

stsci-hack:
The data coming into this step should always be rate images anyway, so for those cases where the native units are already e/s, there should be no need to perform any additional transformation (unless I am not thinking clearly, which is always a possibility lately...).

mcara:
> _"... for those cases where the native units are already e/s, there should be no need to perform any additional transformation"_

[...]

PR #95, opened 2017-11-17
Fix unit conversion for combine thresholds

mcara:
This PR should address the issue of unit conversion for either `combine_lthresh` or `combine_hthresh` parameters from the `CREATE MEDIAN IMAGE` step - see https://github.com/spacetelescope/drizzlepac/issues/94 for more details.

### DIFF
--- a/lib/drizzlepac/createMedian.py
+++ b/lib/drizzlepac/createMedian.py
@@ -169,22 +169,22 @@ def _median(imageObjectList, paramDict):
 
         if lthresh is not None and proc_units.lower() == 'native':
             if native_units.lower() == 'counts':
-                lthresh = lthresh * det_gain
+                lthresh *= det_gain
             elif native_units.lower() == 'counts/s':
-                lthresh = lthresh * det_gain * img_exptime
+                lthresh *= det_gain * img_exptime
             elif native_units.lower() == 'electrons/s':
-                lthresh = lthresh * img_exptime
+                lthresh *= img_exptime
             else:
                 raise ValueError("Unexpected native units: '{}'"
                                  .format(native_units))
 
         if hthresh is not None and proc_units.lower() == 'native':
             if native_units.lower().startswith('counts'):
-                hthresh = hthresh * det_gain
+                hthresh *= det_gain
             elif native_units.lower() == 'counts/s':
-                hthresh = hthresh * det_gain * img_exptime
+                hthresh *= det_gain * img_exptime
             elif native_units.lower() == 'electrons/s':
-                hthresh = hthresh * img_exptime
+                hthresh *= img_exptime
             else:
                 raise ValueError("Unexpected native units: '{}'"
                                  .format(native_units))
--- a/lib/drizzlepac/createMedian.py
+++ b/lib/drizzlepac/createMedian.py
@@ -168,16 +168,22 @@ def _median(imageObjectList, paramDict):
         native_units = image.native_units
         if lthresh is not None:
             if proc_units.lower() == 'native':
-                if native_units.lower() == "counts":
+                if native_units.lower() == 'counts':
                     lthresh = lthresh * det_gain
-                    if native_units.lower() == "counts/s":
-                        lthresh = lthresh * img_exptime
+                elif native_units.lower() in ['counts/s', 'electrons/s']:
+                    lthresh = lthresh * img_exptime
+                else:
+                    raise ValueError("Unexpected native units: '{}'"
+                                     .format(native_units))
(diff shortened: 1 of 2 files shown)


lib/drizzlepac/createMedian.py
```diff
@@ -168,16 +168,22 @@ def _median(imageObjectList, paramDict):
         native_units = image.native_units
         if lthresh is not None:
             if proc_units.lower() == 'native':
-                if native_units.lower() == "counts":
+                if native_units.lower() == 'counts':
```
philhodge: I would suggest changing this to

    if native_units.lower().startswith('counts'):

because regardless of whether the units are 'counts/s' or 'counts', you should multiply by the gain.  The same comment applies to line 180.

lib/drizzlepac/createMedian.py
```diff
@@ -168,16 +168,22 @@ def _median(imageObjectList, paramDict):
...
                     lthresh = lthresh * det_gain
-                    if native_units.lower() == "counts/s":
-                        lthresh = lthresh * img_exptime
+                elif native_units.lower() in ['counts/s', 'electrons/s']:
```
philhodge: I would suggest changing this to

    if native_units.lower() in ['counts/s', 'electrons/s']

lib/drizzlepac/createMedian.py
```diff
@@ -168,16 +168,22 @@ def _median(imageObjectList, paramDict):
...
+                    lthresh = lthresh * img_exptime
+                else:
```
philhodge: This could be a completely separate test:

    if native_units.lower() not in ['counts', 'electrons', 'counts/s', 'electrons/s']:

lib/drizzlepac/createMedian.py
mcara: Yes, we have to deal with "counts" separately (i.e., multiply by gain) but I think I still want to have one long if..elif...else instead of splitting it into two blocks because I want to get to the last `else` and raise an exception when `native_units` are not recognized as it happened when `WFC3` image class was created with `native_units` set to `'ELECTRONS/S'` but this code was not updated with these new units (because it never crashed).

lib/drizzlepac/createMedian.py
```diff
@@ -169,22 +169,22 @@ def _median(imageObjectList, paramDict):
...
             if native_units.lower() == 'counts':
-                lthresh = lthresh * det_gain
+                lthresh *= det_gain
             elif native_units.lower() == 'counts/s':
-                lthresh = lthresh * det_gain * img_exptime
+                lthresh *= det_gain * img_exptime
             elif native_units.lower() == 'electrons/s':
-                lthresh = lthresh * img_exptime
+                lthresh *= img_exptime
             else:
```
philhodge: What if `native_units.lower()` were equal to 'electrons'?

lib/drizzlepac/createMedian.py
```diff
@@ -169,22 +169,22 @@ def _median(imageObjectList, paramDict):
...
                 raise ValueError("Unexpected native units: '{}'"
                                  .format(native_units))
 
         if hthresh is not None and proc_units.lower() == 'native':
             if native_units.lower().startswith('counts'):
```
philhodge: Using `startswith('counts')` here means that the test on 'counts/s' will never be reached.

mcara:
@philhodge Please review latest commit/change to this PR.

lib/drizzlepac/createMedian.py
```diff
@@ -166,18 +166,25 @@ def _median(imageObjectList, paramDict):
...
+        native_units_lc = native_units.lower()
+
+        if proc_units.lower() == 'native':
+            if native_units_lc not in ['counts', 'electrons', 'counts/s',
+                                       'electrons/s']:
+                raise ValueError("Unexpected native units: '{}'"
+                                     .format(native_units))
+
+            if lthresh is not None:
+                if native_units_lc.starswith('counts'):
```
philhodge: typo:  change `starswith` to `startswith`

lib/drizzlepac/createMedian.py
```diff
@@ -166,18 +166,25 @@ def _median(imageObjectList, paramDict):
...
+                    lthresh *= det_gain
+                if native_units_lc.endswith('/s'):
+                    lthresh *= img_exptime
+
+            if hthresh is not None:
+                if native_units_lc.starswith('counts'):
```
philhodge: typo:  change `starswith` to `startswith`

Issue #85, opened 2017-09-26
TweakReg figures not closed? (when interactive mode turned off)


wenlong2:
Hi I came across a problem when make a plot after run the TweakReg task:
When plot a new figure, say, a scatter plot using pylab, which is not related to TweakReg, the TweakReg data (vector plot/hist plot) go on top of the new figure. I was wondering is it possible turn the matplotlib settings back to normal at the end of TweakReg task? Many thanks!

example code:
```
from astropy.io import ascii
import pylab as pl
tweakreg.TweakReg(files=fs_fits, configobj=f_cfg, refimage=f_ref, catfile='mtxy.lst',
                              refcat=self.f_cat, updatehdr=updatehdr, wcsname=wcsname)
xys = ascii.read(self.outdir+'shifts.txt', data_start=1)
pl.errorbar(xys['col2'], xys['col3'], xerr=xys['col6'], yerr=xys['col7'], fmt='o', alpha=0.8)
pl.savefig('test.png')
pl.close()
```
The ``test.png`` was supposed to be scatter plot of the frame shifts, however, it turns out be a vector plot of TweakReg task on top of the scatter plot.

mcara:
CC: @stsci-hack I think this would be down your alley...

stsci-hack:
PR #96 has been submitted to correct this issue.  Once it has been tested, it should solve the user's problem.

PR #96, opened 2017-12-01
Changes to isolate tweakreg plots from user environment

stsci-hack:
These changes are intended to address the plot interaction plot raised in issue #85.  The changes isolate the interaction with matplotlib inside tweakreg by resetting the matplotlib interactive setting to the starting user value while also creating a new figure upon finishing the plotting.

This combination of changes allowed, under interactive testing using the commands provided in issue #85, to properly render the user plots after tweakreg has completed, regardless of whether tweakreg had 'interactive' turned on or not. 

Additional changes were implemented to simply keep references to the pyplot library consistent within the package.

### DIFF
--- a/lib/drizzlepac/catalogs.py
+++ b/lib/drizzlepac/catalogs.py
@@ -357,12 +357,12 @@ class Catalog(object):
 
         """
         try:
-            from matplotlib import pyplot as pl
+            from matplotlib import pyplot as plt
         except:
-            pl = None
+            plt = None
 
-        if pl is not None: # If the pyplot package could be loaded...
-            pl.clf()
+        if plt is not None: # If the pyplot package could be loaded...
+            plt.clf()
             pars = kwargs.copy()
 
             if 'marker' not in pars:
@@ -382,8 +382,9 @@ class Catalog(object):
                 pl_vmax = pars['vmax']
                 del pars['vmax']
 
-            pl.imshow(self.source,cmap=pl_cmap,vmin=pl_vmin,vmax=pl_vmax)
-            pl.plot(self.xypos[0]-1,self.xypos[1]-1,pars['marker'])
+            plt.imshow(self.source,cmap=pl_cmap,vmin=pl_vmin,vmax=pl_vmax)
+            plt.plot(self.xypos[0]-1,self.xypos[1]-1,pars['marker'])
+            plt.figure() # create new figure for subsequent plots (by user?)
 
     def writeXYCatalog(self,filename):
         """ Write out the X,Y catalog to a file
@@ -740,13 +741,15 @@ class UserCatalog(Catalog):
 
         """
         try:
-            from matplotlib import pyplot as pl
+            from matplotlib import pyplot as plt
         except:
-            pl = None
+            plt = None
 
-        if pl is not None:
-            pl.clf()
-            pl.plot(self.xypos[0],self.xypos[1],**kwargs)
+        if plt is not None:
+            plt.clf()
+            plt.plot(self.xypos[0],self.xypos[1],**kwargs)
+
+            plt.figure()  # create new figure for subsequent plots (by user?)
 
 
 class RefCatalog(UserCatalog):
--- a/lib/drizzlepac/tweakutils.py
+++ b/lib/drizzlepac/tweakutils.py
@@ -783,7 +783,15 @@ def make_vector_plot(coordfile,columns=[1,2,3,4],data=None,figure_id=None,
             Write out plot to a file with this name if specified.
 
     """
+    import matplotlib
     from matplotlib import pyplot as plt
+    interactive = matplotlib.is_interactive()
+    turn_interactive_off = False
+
+    if not interactive:
+        plt.ion()
+        if not interactive:
+            turn_interactive_off = True
 
     if data is None:
         data = readcols(coordfile,cols=columns)
@@ -828,6 +836,7 @@ def make_vector_plot(coordfile,columns=[1,2,3,4],data=None,figure_id=None,
         write_xy_file(output,[xy1x,xy1y,dx,dy])
 
     fig = plt.figure(num=figure_id)
+
     if not append:
         plt.clf()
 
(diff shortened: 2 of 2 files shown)


lib/drizzlepac/catalogs.py
```diff
@@ -382,8 +382,9 @@ def plotXYCatalog(self, **kwargs):
                 pl_vmax = pars['vmax']
                 del pars['vmax']
 
-            pl.imshow(self.source,cmap=pl_cmap,vmin=pl_vmin,vmax=pl_vmax)
-            pl.plot(self.xypos[0]-1,self.xypos[1]-1,pars['marker'])
+            plt.imshow(self.source,cmap=pl_cmap,vmin=pl_vmin,vmax=pl_vmax)
+            plt.plot(self.xypos[0]-1,self.xypos[1]-1,pars['marker'])
+            plt.figure() # create new figure for subsequent plots (by user?)
```
mcara: Shouldn't `figure()` be called before `plot()`? CC: @sosey

mcara: Shouldn't `figure()` be called before `plot()`? CC: @sosey

lib/drizzlepac/tweakutils.py
```diff
@@ -783,7 +783,15 @@ def make_vector_plot(coordfile,columns=[1,2,3,4],data=None,figure_id=None,
...
 
     """
+    import matplotlib
     from matplotlib import pyplot as plt
+    interactive = matplotlib.is_interactive()
+    turn_interactive_off = False
+
+    if not interactive:
+        plt.ion()
+        if not interactive:
```
mcara: Is this second `if` even necessary? It seems that the first one already selects a branch that corresponds to "non-interactive" mode.

lib/drizzlepac/catalogs.py
mcara: Shouldn't figure() be called before plot()? CC: @sosey

lib/drizzlepac/catalogs.py
```diff
@@ -740,13 +741,15 @@ def plotXYCatalog(self, **kwargs):
...
+            plt = None
 
-        if pl is not None:
-            pl.clf()
-            pl.plot(self.xypos[0],self.xypos[1],**kwargs)
+        if plt is not None:
+            plt.clf()
+            plt.plot(self.xypos[0],self.xypos[1],**kwargs)
+
+            plt.figure()  # create new figure for subsequent plots (by user?)
```
mcara: Shouldn't figure() be called before plot()? CC: @sosey

lib/drizzlepac/tweakutils.py
sosey: The difference between interactive and non-interactive mode is when the plots get updated to the screen. I'm unsure how the plotting functions are being used specifically in this package. If the intention is to generate a bunch of plots that stay on the screen for the user to look at before continuing, then non-interactive mode is commonly used. Using the show() function will throw the plots up together and block until you want to destroy them. But if interactive is already on for some other reason,  you can still do plotting without switching between the two.

lib/drizzlepac/catalogs.py
sosey: arg, I can't believe the comments got deleted :(   attempting to recreate....

lib/drizzlepac/catalogs.py
sosey: It might be better to work with your own figure so that users plots aren't being clobbered. I would recommend considering the following changes:

- move the import of matplotlib to the top of the module, into a try that sets something like `matplotlib_installed = True`; available for both classes
-  add a useful ImportError statement in the except block that tells the user matplotlib couldn't be found and plotting wont happen
- each catalog class can then have an instance figure attribute, I don't believe this will interfere with any other plots that the user has already, or will later, create
- now use the axes class call in matplotlib for creating the plot, this would make plotXYCatalog look something like the following:
```
def plotXYCatalog(self, **kwargs):
        """
        Method which displays the original image and overlays the positions
        of the detected sources from this image's catalog.
        Plotting `kwargs` that can be provided are:
            vmin, vmax, cmap, marker
        Default colormap is `summer`.
        """
 
        if matplotlib_installed:
            self.fig.clf()
            self.fig.add_subplot(111)
            ax = self.fig.gca()
            pars = kwargs.copy()
 
            if 'marker' not in pars:
                pars['marker'] = 'b+'
 
            if 'cmap' in pars:
                pl_cmap = pars['cmap']
                del pars['cmap']
            else:
                pl_cmap = 'summer'
            pl_vmin = None
            pl_vmax = None
            if 'vmin' in pars:
                pl_vmin = pars['vmin']
                del pars['vmin']
            if 'vmax' in pars:
                pl_vmax = pars['vmax']
                del pars['vmax']
 
            ax.imshow(self.source, cmap=pl_cmap, vmin=pl_vmin, vmax=pl_vmax)
            ax.plot(self.xypos[0]-1, self.xypos[1]-1, pars['marker'])
            self.fig.canvas.draw() # or plt.draw()
    else:
        print("Warning: Matplotlib not found, unable to plot")
 ```
draw() will refresh the plot in interactive mode if you are using object method and function calls.

lib/drizzlepac/tweakutils.py
sosey: The difference between interactive and non-interactive mode is when the plots get updated to the screen. I'm unsure how the plotting functions are being used specifically in this package. If the intention is to generate a bunch of plots that stay on the screen for the user to look at before continuing, then non-interactive mode is commonly used. Using the show() function will throw the plots up together and block until you want to destroy them. But if interactive is already on for some other reason, you can still do plotting without switching between the two.

mcara:
@stsci-hack Also, please do not forget to update `.../doc/source/release_notes.rst` file with the changes in this PR (under the "Drizzlepac v2.1.21 (Unreleased)" section).

stsci-hack:
I am more than happy to have someone with real expertise with matplotlib (like @sosey) provide updates to the plotting code used in tweakreg.  Obviously, the current implementation is not adequate (relying on pyplot interface).

PR #125, opened 2018-04-11
Fixes to let WFPC2 data process

stsci-hack:
These fixes were needed in order to allow WFPC2 data from the archive to be processed again.


mcara:
@stsci-hack Just noticed: in the release notes you did not increment version. If you want to wait for more changes before making a release, please change release date to something like "not yet released". Still, please increment drizzlepac version to 2.2.1

PR #132, opened 2018-05-02
Improve setuptools overrides and JenkinsFile

jhunkeler:
OK, a few things are happening in this PR:

1. #142 works, however it fails to generate `.egg` files by default, and does not automatically obtain dependencies listed under `install_requires`. Thanks to Conda, I missed this little gem while testing things. To fix this behavior I've basically implemented to what `setuptools`'s `install` command does internally.

2. `Jenkinfile` now implements a full build matrix (`python` + `numpy` + `astropy`). For as many combinations as there are it still finishes rather quickly... so that's good.

3. All of `drizzlepac`'s modules and scripts are imported or executed post-build to ensure the package is "alive". The only caveat so far is if an import or script fails it kills the job, so if the first import attempt bombs it could take several "fixup" iterations to get back to an "all green" state. (I'll work on a better solution to this issue in a separate PR)

cc @mcara @stsci-hack @sosey


jhunkeler:
Comin' right up! 👨‍🍳

mcara:
I downloaded your branch and run it with `develop` and `install` and it works.

setup.py
```diff
@@ -111,8 +111,10 @@ def run(self):
             self.do_egg_install()
 
         if not os.path.exists(docs_compiled_dest):
-            print('warning: Sphinx "htmlhelp" documentation was '
-                  'NOT bundled!', file=sys.stderr)
+            print('\nwarning: Sphinx "htmlhelp" documentation was NOT bundled!\n'
+                  '         Execute the following then reinstall:\n\n'
```
mcara: Let me understand this: 1) users have to run install to see this message; 2) users need to run `build_sphinx`; 3) users need to run install AGAIN (=re-install)??? Isn't this too much?

PR #134, opened 2018-05-02
Update setup version requirement for stsci.image

mcara:
CC: @jhunkeler 

In Python 2.7 (my env after updating it) `stsci.image` was `2.2.0` but `drizzlepac` requires `2.3.0`.


mcara:
Deleted branch by mistake. Restoring it.

PR #140, opened 2018-05-02
Add current dir last to sys.path so that RDT works

mcara:
This fixes `python setup.py build_sphinx` and likely `RTD` but it may affect `python setup.py develop`.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -33,15 +33,18 @@ def setup(app):
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('../'))
-sys.path.insert(0, os.path.abspath('../../'))
-sys.path.insert(0, os.path.abspath('../../src/'))
 
-
+#sys.path.insert(0, os.path.abspath('../../'))
 #sys.path.insert(0, os.path.abspath('../'))
 #sys.path.insert(0, os.path.abspath('packagename/'))
 #sys.path.insert(0, os.path.abspath('exts/'))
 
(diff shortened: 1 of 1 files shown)


jhunkeler:
Is there a way to pass `python setup.py build_ext --inplace` on RTD? I don't think the path order matters so much as the shared library not existing relative to the local `./drizzlepac` package sphinx is examining/importing from.

```
$ cd drizzlepac
$ python setup.py install
$ python -c 'import drizzlepac'

# Boom

 Coordinate transformation and image resampling library NOT found!

 Please check the installation of this package to insure C code was built successfully.
  File "/Users/jhunk/Downloads/drizzlepac/drizzlepac/ablot.py", line 27, in <module>
    from . import cdriz
ImportError: cannot import name 'cdriz'

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "/Users/jhunk/Downloads/drizzlepac/drizzlepac/__init__.py", line 24, in <module>
    from . import ablot
  File "/Users/jhunk/Downloads/drizzlepac/drizzlepac/ablot.py", line 32, in <module>
    raise ImportError
ImportError
```

Same directory, same everything, but building in-place so the C extensions are injected into the local package:
```
$ python setup.py build_ext --inplace
$ python -c 'import drizzlepac'
# no error
```

mcara:
@jhunkeler I would suggest you try this PR.

sosey:
http://docs.readthedocs.io/en/latest/builds.html suggests they really only use `setup.py intall`

sosey:
why can't the user just use develop or install and then go in docs and `make html` ?

sosey:
@jhunkeler ya lost me a little, if they haven't checked out the repo how will they build the docs locally?

mcara:
I think https://github.com/spacetelescope/drizzlepac/pull/142 may be a more promising approach with fewer potential side-effects so I am withdrawing this PR.

PR #141, opened 2018-05-02
mdriz: Fix broken imports

jhunkeler:
Given the state of this module and what it refers to... can it be deprecated?

This is only being fixed so that my import tests will succeed.

https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fdrizzlepac/detail/PR-132/11/pipeline

### DIFF
--- a/drizzlepac/mdriz.py
+++ b/drizzlepac/mdriz.py
@@ -8,10 +8,11 @@ Main program for running MultiDrizzle from the command line.
 
 """
 from __future__ import absolute_import, division, print_function # confidence high
+import getopt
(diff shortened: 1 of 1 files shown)


stsci-hack:
Given that we have not documented this interface, it should not cause any problems to deprecate this specific module.  You could start by removing it from the setup.py file, then in the following release (if we get no complaints, as expected) we could remove it entirely from the package.   Of course, we could always take the dramatic approach and simply remove all traces of this now, but that may have unexpected consequences with the community...

PR #142, opened 2018-05-02
Sphinx brain

jhunkeler:
- Sphinx
  - `conf.py` adds the `.eggs` directory populated by `setup_requires`. This is useful if compiling the docs outside of Conda.

- Setuptools
  - Implement `BuildSphinx`  wrapper ripped straight from `stsci-jwst/jwst.git`. The overall logic was modified slightly to prevent a chicken/egg scenario when Sphinx wasn't available. (see the `CMDCLASS` dictionary)
  - Implement `InstallCommand` wrapper to ensure C extensions are compiled/available on `sys.path` when importing `drizzlepac` from a path relative to the project. (RTD support)
  - Does not touch `develop` since that build mode already performs `build_ext --inplace` by design
  - Does not change behavior of `cd doc/source && make html`. Doing this manually still might not work unless you have the correct dependencies pre-installed, either manually, provided by `setup_requires`, or a conda recipe, _and_ `python setup.py build_ext --inplace` has been executed beforehand.

If a general user or developer wants to install drizzlepac with the `htmlhelp` docs bundled:
```bash
$ conda install sphinx graphviz
# ^ Graphviz is not a Python package so pip/pypi/virtualenv users are on their own at this point

$ python setup.py build_sphinx
$ python setup.py install
```

If a user or developer wants to install drizzlepac without the `htmlhelp` docs bundled:
```bash
$ python setup.py install
```

RTD executes the following command prior to building the docs. The `InstallCommand` wrapper should trap this, build the extension(s), and continue without failure:
```bash
$ python setup.py install --force
```

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -35,8 +35,8 @@ def setup(app):
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('../'))
 sys.path.insert(0, os.path.abspath('../../'))
+sys.path.insert(0, os.path.abspath('../../.eggs'))
 sys.path.insert(0, os.path.abspath('../../src/'))
-
 
 #sys.path.insert(0, os.path.abspath('../'))
 #sys.path.insert(0, os.path.abspath('packagename/'))
--- a/setup.py
+++ b/setup.py
@@ -90,8 +90,11 @@ docs_compiled_dest = os.path.normpath('{0}/htmlhelp'.format(NAME))
 
 
 class InstallCommand(install):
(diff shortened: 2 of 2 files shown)


jhunkeler:
RTD Experiment URL:

https://drizzlepac-jhunk.readthedocs.io

[Documentation Status](http://drizzlepac-jhunk.readthedocs.io/en/latest/?badge=latest)

mcara:
I have tested a wheel from @jhunkeler and this approach seems to work well and hopefully with fewer side-effects compared to https://github.com/spacetelescope/drizzlepac/pull/140.

PR #149, opened 2018-05-08
Add jenkins rt

stsci-hack:
This PR includes all the updates necessary to add Jenkins-based regression tests which rely on Artifactory (or local directory) for inputs.  The initial tests included are all the tests for ACS.  Tests for other instruments will be included later.


pllim:
I don't see any modifications done to `Jenkinsfile`. Therefore, the current CI report tied to this PR is misleading, because they are not tested in this PR. I only see install and doc build tests.

What is the plan here? Do you only want to run these on a nightly basis? Or do you want to run a subset or all of these per push commit? I would prefer the latter if feasible (representative subset is an acceptable compromise if not all of them).

conftest.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # Add option to use big data sets
+    parser.addoption(
+        "--bigdata",
+        action="store_true",
+        help="use big data sets (intranet)"
+    )
+
+    parser.addoption(
+        "--env",
```
pllim: Is this really necessary? What is this used for?

drizzlepac/helpers/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .io import *
```
pllim: Why the `*` imports?

drizzlepac/helpers/io.py
```diff
@@ -0,0 +1,119 @@
...
+    elif check_url(src):
+        download(src, dest)
+
+    else:
+        raise BigdataError('Failed to retrieve data: {}'.format(src))
+
+    return dest
+    
+    
+def upload_results(**kwargs):
```
pllim: This function named scared me a little, as I thought it blindly allows people to upload files. Perhaps `generate_upload_json()` is a bit less scary? :sweat_smile:

drizzlepac/helpers/utils.py
```diff
@@ -0,0 +1,245 @@
...
+RE_URL = re.compile('\w+://\S+')
+
+default_compare = dict(
+    ignore_keywords=['DATE', 'CAL_VER', 'CAL_VCS', 'CRDS_VER', 'CRDS_CTX'],
+    keys=['primary', 'sci', 'dq'],
+    rtol=0.000001,
+)
+
+
+def cmp_fitshdr(left, right, **kwargs):
```
pllim: Why not just use http://astropy.readthedocs.io/en/latest/io/fits/api/diff.html#astropy.io.fits.HeaderDiff ?

`fits.HDUList([a[kw] for kw in keys])` should not be necessary (at least in theory).

drizzlepac/helpers/utils.py
```diff
@@ -0,0 +1,245 @@
...
+                                        rtol=rtol,
+                                        **kwargs)
+
+    if no_assert:
+        return result
+
+    assert result.identical, result.report()
+
+
+def cmp_gen_hdrkeywords(fitsobj, base, keywords, limit=0, start=0):
```
pllim: Again, I am not convinced that this is not something that is already in one of the classes mentioned in http://astropy.readthedocs.io/en/latest/io/fits/usage/misc.html#differs .

drizzlepac/helpers/utils.py
```diff
@@ -0,0 +1,245 @@
...
+    -------
+    match: boolean
+        True if the strings match
+    """
+    words1 = str1.split()
+    words2 = str2.split()
+    if len(words1) != len(words2):
+        return False
+    for w1, w2 in zip(words1, words2):
+        if w1[:length] != w2[:length]:
```
pllim: This is case-sensitive. I am not sure what you use this for, but if you are checking FITS keywords, perhaps you also want to add in an option for case-insensitive comparison?

drizzlepac/helpers/utils.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def download(url, dest):
+    """Simple http/https downloader
+    """
+    dest = os.path.abspath(dest)
+
+    with requests.get(url, stream=True) as r:
+        with open(dest, 'w+b') as data:
+            for chunk in r.iter_content(chunk_size=0x4000):
```
pllim: Is chunk size a magic number?

drizzlepac/tests/acs/test_acs_kernels.py
```diff
@@ -0,0 +1,61 @@
+import os
+
+from stsci.tools import teal
+import drizzlepac
+from drizzlepac import astrodrizzle
+from drizzlepac.helpers.mark import require_bigdata
+from drizzlepac.helpers.io import get_bigdata
```
pllim: Why not relative import like L9 below?

drizzlepac/tests/acs/test_acs_kernels.py
```diff
@@ -0,0 +1,61 @@
...
+
+from ..helpers import BaseACS, raw_from_asn
+
+def pytest_generate_tests(metafunc):
```
pllim: Since you are only parametrizing one method below, maybe it is easier to read to use https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions decorator instead? Or do you plan to add more stuff to this file?

drizzlepac/helpers/utils.py
jhunkeler: Those `cmp_*` functions were written to encapsulate the ~10 lines of code used per test in `jwst.git`. If there's a cleaner way to do any of this... we should. I was only basing those functions off what I was seeing at the time.

drizzlepac/helpers/utils.py
jhunkeler: `16384` :grin:

pllim: What does that mean in plain English? Should be documented. :wink:

jhunkeler: All this does is iterate over 16 kilobytes at a time instead of reading the entire file into memory at once. That's all.

stsci-hack:
This PR now includes code to run ALL Drizzlepac tests currently active under Pandokia using input and truth files stored on Artifactory.  The UNIT tests defined in 'acs/test_unit.py' currently do NOT pass, but they at least run to completion.  These failures will need to be investigated and corrected sometime after I return from my vacation (unless someone else wants to investigate these differences sooner...). 

@jhunkeler  I feel these tests can be merged in now (even with failures) to enable immediate Pandokia testing of Drizzlepac to be disabled immediately.

stsci-hack:
@jhunkeler  Actually, I overlooked one set of tests: the drizzlepac.ui tests.  I was not sure how to translate them to run under pytest (at least not in the time I had this week).  In addition, they seemed rather minimal in utility, so if we do get them added in later, it will not be a problem.

jhunkeler:
Superseded by #156

Issue #150, opened 2018-05-09
TST: Existing Jenkins tests have warnings


pllim:
While inspecting #149 results, I noticed:

1. doc build
```
WARNING: LaTeX command 'latex' cannot be run (needed for math display), check the imgmath_latex setting
```

2. install tests (all Python versions)
```
warning: Sphinx "htmlhelp" documentation was NOT bundled!
```

For the doc build, I think adding `-w` option would make the test fail on any warning (optional but good practice).

For the install tests, do you know why it is complaining about "htmlhelp"? Can we safely ignore this?

c/c @jhunkeler @rendinam

pllim:
Also see https://github.com/spacetelescope/drizzlepac/pull/173#issuecomment-420119425

PR #151, opened 2018-05-09
EXP: Add -w to doc build in Jenkins

pllim:
**DO NOT MERGE**

I just want to see if "-w" really works. This does not actually fix the warning reported in #150


pllim:
Ops, nevermind. Astropy uses something extra, which might not be available here.

```
error: option -w not recognized

script returned exit code 1
```

Issue #157, opened 2018-06-11
UnboundLocalError: local variable 'errorobj' referenced before assignment


griffin-h:
When an exception occurs in a function called with `util.WithLogging`, that exception never makes it to the user because of an `UnboundLocalError` in the `wrapper` function. This is because `errorobj` is not defined inside the `finally` statement. It can be fixed by changing lines 219-220 of util.py from
```
            except Exception as errorobj:
                raise
```
to
```
            except Exception as e:
                errorobj = e
                raise
```

mcara:
@griffin-h Thank you for reporting this and proposing a fix. I have implemented your fix in #158.

PR #158, opened 2018-06-11
Fix undefined errorobj in with_logging

mcara:
This PR implements the fix proposed by @griffin-h in #157.


mcara:
@jhunkeler Is there anybody out there???

mcara:
@jhunkeler You are a good man

PR #165, opened 2018-07-18
Dummy PR for testing stscibot integration

drdavella:


drdavella:
It works! But @mcara I think you will need to move the changelog format to `rst` (although you still shouldn't have to add PR numbers to old entries).

mcara:
@drdavella Isn't changelog already in `rst`? https://github.com/spacetelescope/drizzlepac/blob/master/CHANGELOG.rst

PR #173, opened 2018-08-20
Update unit tests input file

stsci-hack:
Update unit tests for Drizzlepac to include the following revisions:

- PEP 8 compliance (remove extraneous imports,  primarily)
- Eliminate use of 'difference' files as 'truth' files, since they are only outputs for the test and not compared
- Switch inputs to old version of file which does NOT have SIP (just like the file used by Pandokia)

This allows these unit tests to once again match the last results from Pandokia testing, and can now serve as the basis for maintaining the code.


pllim:
Example log with warnings: https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fdrizzlepac/detail/PR-173/8/pipeline/29/

tests/acs/test_unit.py
```diff
@@ -512,11 +504,11 @@ def test_blot_with_lan5(self):
         assert(med_diff < 1.0e-6)
         assert(max_diff < 1.0e-5)
 
-    def test_blot_with_image(self):
+    def not_test_blot_with_image(self):
```
pllim: I counted 14 methods in this file but pytest is only running 11 of them, which means it is not collecting those named "not_test*". Is this intentional?

stsci-hack: That is intentional for the time being.  Those which have been renamed 'not_test' are being included but 'skipped' due to longer term issues with those specific test definitions. These are intended to be updated and turned on in a future PR.

pllim: Why not mark them as `xfail`? It would be more obvious that way, no?

stsci-hack: I will evaluate what would be best for these tests, looking to mark them as 'xfail' where possible.

stsci-hack: 1 test was reset to use 'xfail', with possible reason stated, while the other 2 cases you identified were revised so that they no longer needed 'xfail', but would pass.

pllim:
"Immediate interpolation of variables in the 'env_vars' list is not supported and will probably not do what you expect. Please change the double quotes (") to single quotes (') in each value of the 'env_vars' list." -- from Jenkins log. Not sure what this means.

rendinam:
Are the values `BUILD_MATRIX_SUFFIX` and `BUILD_MATRIX_ID` used anywhere in the build? Specifically, are they needed by the commands that run in `build_cmds` or `test_cmds`?

If not, they should be removed to eliminate the "Immediate interpolation" warning. If those variables are needed, some extra thought will need to be put into this due to conflicting goals in the build support machinery regarding how environment variables are set.

stsci-hack:
Unfortunately, I believe the MATRIX_SUFFIX and MATRIX_ID variables are part of the machinery used in labelling the results directories for failing tests on Artifactory.  This was something Joe came up with and I am learning by using.   If this needs to be modified, someone more experienced with Jenkinsfiles will need to weigh in on what should be done and how.

rendinam:
I see. There are safeguards in place to check for the type of variable expansion taking place so that jobs that need to modify an existing environment variable may use the syntax `var=${SHELL_VAR}:/extra/stuff`, which gets expanded in the build environment rather than by the interpreter that parses the Jenkinsfile. This is useful for adding to the PATH var, for example. If local expansions in the Jenkinsfile are allowed for `env_vars` entries, this type of direct environment modification would not be possible.

A different method of string concatenation looks like it might allow a way to do both early expansion as is needed here and late expansion as is needed by jobs that modify PATH and similar. I'll push a modification shortly.

pllim:
Might be worth thinking about changing one of the Python 3.7 build to use Astropy dev to make sure nothing there breaks `drizzlepac` (before it is too late):

https://github.com/spacetelescope/acstools/blob/6ff49129f338f3ae6a9cf82ee843b1376dcd358a/Jenkinsfile#L33

And in case it helps, this enables listing of dependency versions in the test run:

https://github.com/spacetelescope/acstools/blob/6ff49129f338f3ae6a9cf82ee843b1376dcd358a/conftest.py#L14-L37

rendinam:
Build 12 https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fdrizzlepac/detail/PR-173/12/pipeline/39 passed all configurations, but the variables `BUILD_MATRIX_SUFFIX=${matrix_id}_py${python_ver}_np${numpy_ver}_ap${astropy_ver}' and `BUILD_MATRIX_ID=0' were not set correctly.

The environment received the strings `BUILD_MATRIX_SUFFIX=${MATRIX_SUFFIX}` and `BUILD_MATRIX_ID=${matrix_id}` and the shell did not have the right hand side variables to dereference defined within it, so the assignments were null.

The artifacts in the repo for `17Sep2018T` show that build 12 created `drizzlepac-results/acs/17Sep2018T/test_blot_with_image0_jenkins-STScI-drizzlepac-PR-173-12_standalone/`.

Whereas, build 13 https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fdrizzlepac/detail/PR-173/13/pipeline/36 also passed but with an updated method of defining those values within the Jenkinsfile before the environment comes up. Build 13 created a series of variations on `drizzlepac-results/acs/17Sep2018T/test_blot_with_image0_jenkins-STScI-drizzlepac-PR-173-13_0_py36_np114_ap305/`. The `0_py36_np114_ap305` portion is the value that gets propagated from `BUILD_MATRIX_ID` and `BUILD_MATRIX_SUFFIX`. This confirms that the adjustment which triggered build 13 corrects the problem.

PR #182, opened 2018-10-01
Fix rtd failure install numpy

mcara:


drdavella:
@mcara that should be the current behavior: the bot says everything looks good. Or am I missing something?

mcara:
@drdavella See https://github.com/spacetelescope/drizzlepac/pull/182#issuecomment-426095872

mcara:
@drdavella `no-changelog-entry-needed` was added _before_ PR was created.

drdavella:
It should work whether `no-changelog-entry-needed` is set when the PR is created, or afterwards. In any event, I'm confused, because the bot is already demonstrating the behavior you requested. If you remove the label, the bot should update the message to complain, and then if you restore it, the bot should restore the current message.

mcara:
Well the label is here and so is the message. Shouldn't it go away?

drdavella:
No, the message appears in all cases. But in this case it says:

> Everything looks good from my point of view! 👍

mcara:
Oh, I missed _"Everything looks good from my point of view!"_

PR #185, opened 2018-10-26
Add JenkinsfileRT for use in testing against new astropy.

rendinam:
This builds and tests `drizzlepac` in an environment consisting of all stable packages except for
   * `astropy`
   * `stwcs`
which are taken from the latest master snapshot provided by astroconda-dev.

For now, this will be run in a special purpose Jenkins job on the RT Jenkins instance.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@ TEST_DEPS = "pytest pytest-remotedata crds"
 
 // Conda needs explicit dependencies listed
 DEPS = "fitsblender graphviz nictools numpydoc \
-        pytest pytest-remotedata pyregion \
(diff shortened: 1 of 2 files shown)


rendinam:
Not sure what's going on with the `Jenkinsfile`. `DEPS` is defined for conda, but then never used, so the latest change to pin `pytest` to a known-working version to allow the CI tests to pass is ineffective. Since fixing this is beside the original intent of this PR, namely to simply add the JenkinsfileRT for use in meeting a request to test against `astropy` and `stwcs` master, I am merging this and will investigate the underlying issue afterwards. This leaves the tests no worse off than they were before the PR was submitted.

Issue #189, opened 2018-11-01
Incorrect units in CR-cleaned images


mcara:
@stsci-hack I was processing a set of WFC3/IR images and it appears that CR-cleaned images (`'*_crclean.fits'`) have data that appear to be `FLT * EXPTIME` but the header shows that units are `'ELECTRONS/S'`.

mcara:
Eh, based on my reading of the code and some instrument books, almost all instruments that do not have units of `electrons` must be affected by this bug, including instruments such as `WFPC2` that have units of counts (need gain to convert to electrons) or `NICMOS` (units of count-rate).

mcara:
Fixed via #190

PR #190, opened 2018-11-02
Fix incorrect units in CR-cleaned images

mcara:
This PR should fix incorrect units in CR-cleaned images created by ``astrodrizzle`` - see #189. Now CR-cleaned images should have the same units as input images.

### DIFF
--- a/drizzlepac/drizCR.py
+++ b/drizzlepac/drizCR.py
@@ -306,8 +306,8 @@ def _drizCr(sciImage, virtual_outputs, paramDict):
             np.bitwise_and(__dqMask,__crMask,__dqMask)
 
             ####### Create the corr file
-            __corrFile = np.zeros(__inputImage.shape,dtype=__inputImage.dtype)
-            __corrFile = np.where(np.equal(__dqMask,0),__blotData,__inputImage)
+            __corrFile = np.where(__dqMask, __inputImage, __blotData)
+            __corrFile /= scienceChip._conversionFactor
             __corrDQMask = np.where(np.equal(__dqMask,0),
                                     paramDict['crbit'],0).astype(np.uint16)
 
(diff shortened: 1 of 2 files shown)


drizzlepac/drizCR.py
```diff
@@ -306,8 +306,8 @@ def _drizCr(sciImage, virtual_outputs, paramDict):
             np.bitwise_and(__dqMask,__crMask,__dqMask)
 
             ####### Create the corr file
-            __corrFile = np.zeros(__inputImage.shape,dtype=__inputImage.dtype)
-            __corrFile = np.where(np.equal(__dqMask,0),__blotData,__inputImage)
+            __corrFile = np.where(__dqMask, __inputImage, __blotData)
+            __corrFile /= scienceChip._conversionFactor
             __corrDQMask = np.where(np.equal(__dqMask,0),
```
mcara: Eh, now I see that if I looked a little bit past the `__corrFile`, I could have cleaned up `__corrDQMask` as well (no more `np.equal(__dqMask,0)`)

drizzlepac/drizCR.py
```diff
@@ -306,8 +306,8 @@ def _drizCr(sciImage, virtual_outputs, paramDict):
             np.bitwise_and(__dqMask,__crMask,__dqMask)
 
             ####### Create the corr file
-            __corrFile = np.zeros(__inputImage.shape,dtype=__inputImage.dtype)
-            __corrFile = np.where(np.equal(__dqMask,0),__blotData,__inputImage)
+            __corrFile = np.where(__dqMask, __inputImage, __blotData)
+            __corrFile /= scienceChip._conversionFactor
```
mcara: @stsci-hack Eh, now I see that if I looked a little bit past the `__corrFile`, I could have cleaned up `__corrDQMask` as well (no more `np.equal(__dqMask,0)`)

PR #192, opened 2018-11-03
Cleanup CR-cleaning code and remove dependency on stsci.convolve

mcara:
This PR cleans up core `drizCR.py` code (used for cosmic ray cleaning). In addition, it replaces calls to `stsci.convolve` with calls to `scipy.signal.convolve2d`. In light of this, @pllim , maybe https://github.com/spacetelescope/stsci.convolve/issues/1 is doable assuming no other STScI package  depends on `stsci.convolve`.

CC: @stsci-hack

### DIFF
--- a/drizzlepac/drizCR.py
+++ b/drizzlepac/drizCR.py
@@ -158,7 +158,8 @@ def _driz_cr(sciImage, virtual_outputs, paramDict):
                               .format(blot_image_name))
 
             try:
-                blot_image = fits.open(blot_image_name, memmap=False)
+                blot_data = fits.getdata(blot_image_name, ext=0)
+                blot_data *= sci_chip._conversionFactor
             except IOError:
                 print("Problem opening blot images")
                 raise
@@ -170,10 +171,7 @@ def _driz_cr(sciImage, virtual_outputs, paramDict):
         input_image *= sci_chip._conversionFactor
 
         # make the derivative blot image
-        blot_data = blot_image[0].data * sci_chip._conversionFactor
         blot_deriv = quickDeriv.qderiv(blot_data)
-        if not sciImage.inmemory:
-            blot_image.close()
 
         # Boolean mask needs to take into account any crbits values
         # specified by the user to be ignored when converting DQ array.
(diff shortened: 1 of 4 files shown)


pllim:
50% code reduction. I am excited about this! 👏

drizzlepac/drizCR.py
```diff
@@ -7,35 +7,34 @@
 :License: :doc:`LICENSE`
 
 """
-from __future__ import absolute_import, division, print_function # confidence medium
+from __future__ import absolute_import, division, print_function
```
stsci-hack: Simply curious: do we still need these imports if we are only supporting Python 3.x?

mcara: @stsci-hack I am not sure what _"we are only supporting Python 3.x"_ really means since I see `drizzlepac` is still available in `astroconda` with `Python 2.7`. Is it not? If it still is, I suppose it better work. Feel free to investigate this issue in more depth but from my point of view, it just does not hurt to leave it there. The effort to remove these `from __future__` imports should be minimal "in future".

pllim: If you are only supporting Python 3, the code throws an error when you try to **install** in Python 2 **and** it is no longer released in any Python 2 environments.

mcara:
@stsci-hack @pllim This PR is mostly a code refactoring - there are no feature changes or even measurable performance changes. That is why I added `no-changelog-entry-needed` label. @pllim as an expert, what is your opinion on when a changelog entry is needed?

pllim:
@mcara , change log is for users. If you don't need users need to know about this, then no change log is needed. Documentation change also usually don't need change log because it is self-documenting by nature. 😉

mcara:
@pllim @stsci-hack My idea was to wait for more changes before making a "release" (exactly because this PR add no new features). _But_, if @pllim is in hurry to get rid of `stsci.convolve`, e.g., then we can make a "release" and probably we will need to just specify that this is a "code cleanup" release. Just let me know what you prefer.

pllim:
No rush from me. But maybe @jhunkeler or @rendinam would be happy if delivery is made simpler.

mcara:
Nah, I would say let them suffer 😆

SaOgaz:
@mcara did this make it into astroconda? Are we clear to archive stsci.convolve?

Issue #201, opened 2018-12-27
Are fit_xy computed correctly?


mcara:
CC: @stsci-hack 

In #199 and #200 the formula to compute "fitted XY" is:
```python
fit['fit_xy'] = np.dot(xy, fit['fit_matrix'])+center - fit['offset']
```

However, based on the old code, the fix should have been (see [imgclasses.py Line 663](https://github.com/spacetelescope/drizzlepac/pull/200/files#diff-7679c53854b1814a4439225d3315021eL663)):
```python
xy_fit = self.fit['img_coords'] - self.fit['resids'] + self.refWCS.wcs.crpix
```
which, taking into account https://github.com/spacetelescope/drizzlepac/blob/b40842fe67d5835229f44b41852cdb1242e7faa8/drizzlepac/linearfit.py#L278 is equivalent to 
```python
fit['fit_xy'] = np.dot(xy, fit['fit_matrix'])+center + fit['offset']
```
That is, now I am worried that the sign with which the offset is being added is incorrect.

mcara:
Fixed via #202

Issue #199, opened 2018-12-26
Output FIT catalogs report incorrect sky coordinates


stsci-hack:
The optionally generated FIT catalogs, written out with the '*catalog_fit.match' filename, includes multiple columns worth of information as used in the FIT including the RA/DEC of  the input sources after applying the fit.  Unfortunately, this computation is not being done correctly and that information is completely incorrect.  The values reported in the file are generated using the reference image WCS based on the input source positions corrected by the fit residuals [Line 665 from imgclasses.py](https://github.com/spacetelescope/drizzlepac/blob/1492474e2e562e14dd1a060cc4f5171db9272f0f/drizzlepac/imgclasses.py#L665).  The correct sky position should be generated using the WCS (with distortion) for the chip the source was originally measured from.

stsci-hack:
I have identified the basic problem and will be submitting a proposed fix on 27 Dec.

PR #202, opened 2018-12-28
Implement correct formula for fit_xy

mcara:
This PR implements really correct formula for computation of "aligned"/corrected coordinates of sources from the image catalog in the reference WCS. This corrects the incorrect formula from #200.

Also see https://github.com/spacetelescope/drizzlepac/pull/200#issuecomment-450318949, #199, #201.

CC: @stsci-hack


nden:
tests failed?

mcara:
@nden What tests?

mcara:
@nden I am not familiar with the setup of these tests, but locally I can build the master fine and `ablot` which seems to be the common reason for failure. Could it be an issue with test setup? CC: @stsci-hack  @jhunkeler

Issue #186, opened 2018-10-26
WCS._naxis1/2 deprecated


nden:
As a heads up - the next astropy release, 3.1,  deprecates `WCS._naxis1/2` . Instead `WCS.pixel_shape` should be used. `WCS.pixel_shape` is a list of `[NAXIS1, NAXIS2, ...]`.

philhodge:
pixel_shape is not a good name.  Pixels can be rectangular, and they can even be hexagonal.

nden:
`pixel_shape` is defined in the [WCS APE](https://github.com/astropy/astropy-APEs/pull/45). In this context it means the shape of the data array in pixel space. Please comment on the APE PR if you feel strongly about it.

mcara:
I agree with @philhodge Why is it not `image_shape`?

mcara:
or even `domain_shape`

nden:
@mcara Because there's an ambiguity about the order of axes - numpy order vs FITS order. The compromise was to have two attributes - `pixel_shape` (for FITS ) and `array_shape` (for np ordering).

mcara:
Fixed via https://github.com/spacetelescope/drizzlepac/pull/207 (which also drops support for Python 2).

PR #207, opened 2019-01-16
Remove py2 support. Use WCS.pixel_shape instead of naxis

mcara:
This PR deals with three issues:

1. Fixes deprecation warnings due to deprecation of the `WCS._naxis1` - see https://github.com/spacetelescope/drizzlepac/issues/186

2. Remove support for Python 2.

3. Remove last dependency on `stsci.convolve`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,11 @@ of the list).
 DrizzlePac (DEVELOPMENT)
 ========================
 
+- Replaced the use of ``WCS._naxis1`` and ``WCS._naxis2`` with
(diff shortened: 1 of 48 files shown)


mcara:
@stsci-hack reported a segmentation fault in `tweakreg`. I have traced this to not converting results of convolution to `np.float32` before further processing. Also, `np.linalg.inv` does not support `longdouble` so input arrays need to be converted to `float64`. Next commit will fix this bug introduced in https://github.com/spacetelescope/drizzlepac/pull/205

PR #211, opened 2019-01-27
Fix broken dependencies

jhunkeler:
* `acstools` is listed in `install_requires` so it must be present for doc builds and Jenkins jobs.
* `stsci.convolve` was removed from `install_requires` so it must be removed from doc builds and Jenkins jobs.


JenkinsfileRT
```diff
@@ -21,7 +21,8 @@ bc.nodetype = 'linux'
 bc.env_vars = test_env
 bc.name = '3.6'
 bc.conda_channels = ['http://ssb.stsci.edu/astroconda']
-bc.conda_packages = ['numpy',
+bc.conda_packages = ['acstools',
+                     'numpy',
```
stsci-hack: I was wondering whether simply setting this to 'numpy < 1.16' would fix the problems with the testing on Boyle...

mcara:
@jhunkeler Do you plan to merge this?

PR #212, opened 2019-01-31
Remove unnecessary PATH, HOME overrides

rendinam:
and use relative path to invoke `conda` for installation of special-case development packages to prevent the erroneous use of the python interpreter in the base conda environment for execution of applicable `build_cmds` & `test_cmds`, rather than the interpreter provided in the custom environment created for hosting the build.


rendinam:
Jenkins job is not completing due to an unrelated resource issue. Merging to improve the regression test situation.

PR #215, opened 2019-02-06
test with  dependencies  from the master branches

nden:
@jhunkeler or @rendinam Does this look correct? I want to run those tests with all STScI maintained packages installed from master.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -1,31 +1,27 @@
 // Obtain files from source control system.
 if (utils.scm_checkout()) return
 
-
-def test_env = [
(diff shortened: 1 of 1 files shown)


rendinam:
In principle, yes. I had to add a couple dependencies to make the tests happy. I'd also like to expand this RT job to include a run of the suite against the stable package set as well, to support DMS delivery requirements. Currently working through some issues with that. Once it's ready, I'll push the changes here.

rendinam:
The RT job for drizzlepac on Jenkins has been renamed `drizzlepac` (formerly `drizzlepac-astropytesting`).

PR #216, opened 2019-02-06
Initial load of hlapipeline

mdlpstsci:
hlapipeline source code added to the drizzlepac package.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -1,0 +1,1075 @@
+#!/usr/bin/env python
+
+"""This script is a modernized implementation of tweakreg.
+
+"""
+
+import datetime
+from astropy.io import fits
+from astropy.table import Table
+from astropy.coordinates import SkyCoord, Angle
+from astropy import units as u
+from collections import OrderedDict
+from drizzlepac import updatehdr
+import glob
+import math
+import numpy as np
+import os
+import pdb
+import pickle
+from stsci.tools import fileutil
+from stwcs.wcsutil import headerlet
+from stwcs.wcsutil import HSTWCS
+import sys
+import tweakwcs
+try:
+    from drizzlepac.hlautils import astrometric_utils as amutils
+    from drizzlepac.hlautils import astroquery_utils as aqutils
+    from drizzlepac.hlautils import analyze as filter
+    from drizzlepac.hlautils import get_git_rev_info
+except:
+    from drizzlepac.hlautils import astrometric_utils as amutils
+    from drizzlepac.hlautils import astroquery_utils as aqutils
+    from drizzlepac.hlautils import analyze as filter
+    from drizzlepac.hlautils import get_git_rev_info
+
+MIN_CATALOG_THRESHOLD = 3
+MIN_OBSERVABLE_THRESHOLD = 10
+MIN_CROSS_MATCHES = 3
+MIN_FIT_MATCHES = 6
+MAX_FIT_RMS = 10 # RMS now in mas, 1.0
+MAX_FIT_LIMIT = 1000 # Maximum RMS that a result is useful
+MAX_SOURCES_PER_CHIP = 250  # Maximum number of sources per chip to include in source catalog
+
+# Module-level dictionary contains instrument/detector-specific parameters used later on in the script.
+detector_specific_params = {"acs":
+                                {"hrc":
+                                     {"fwhmpsf": 0.073,
+                                      "classify": True,
+                                      "threshold": None},
+                                 "sbc":
+                                     {"fwhmpsf": 0.065,
+                                      "classify": False,
+                                      "threshold": 2.0},
+                                 "wfc":
+                                     {"fwhmpsf": 0.13, #0.076,
+                                      "classify": True,
+                                      "threshold": -1.1}},
+                            "wfc3":
+                                {"ir":
+                                     {"fwhmpsf": 0.14,
+                                      "classify": False,
+                                      "threshold": None},
+                                 "uvis":
+                                     {"fwhmpsf": 0.076,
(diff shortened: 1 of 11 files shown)


drizzlepac/alignimages.py
```diff
@@ -0,0 +1,1075 @@
...
+        fitStatusDict = best_fitStatusDict.copy()
+
+        # Protect the writing of the table within the best_fit_rms
+        info_keys = OrderedDict(imglist[0].meta['tweakwcs_info']).keys()
+        # Update filtered table with number of matched sources and other information
+        for item in imglist:
+            imgname = item.meta['name']
+            index = np.where(filteredTable['imageName'] == imgname)[0][0]
+
+            if item.meta['tweakwcs_info']['status'].startswith("FAILED") != True:
```
mcara: @mdlpstsci This should be written as `if not ...starswith('FAILED')`. `starswith` returns a boolean result.

drizzlepac/alignimages.py
```diff
@@ -0,0 +1,1075 @@
...
+from astropy.table import Table
+from astropy.coordinates import SkyCoord, Angle
+from astropy import units as u
+from collections import OrderedDict
+from drizzlepac import updatehdr
+import glob
+import math
+import numpy as np
+import os
+import pdb
```
mcara: @mdlpstsci `pdb` is imported but not used anywhere.

drizzlepac/alignimages.py
```diff
@@ -0,0 +1,1075 @@
...
+import pickle
+from stsci.tools import fileutil
+from stwcs.wcsutil import headerlet
+from stwcs.wcsutil import HSTWCS
+import sys
+import tweakwcs
+try:
```
mcara: @mdlpstsci This `try-except` block does not do anything. Also why imports below are absolute?

drizzlepac/alignimages.py
```diff
@@ -0,0 +1,1075 @@
...
+    from drizzlepac.hlautils import astrometric_utils as amutils
+    from drizzlepac.hlautils import astroquery_utils as aqutils
+    from drizzlepac.hlautils import analyze as filter
+    from drizzlepac.hlautils import get_git_rev_info
+except:
+    from drizzlepac.hlautils import astrometric_utils as amutils
+    from drizzlepac.hlautils import astroquery_utils as aqutils
+    from drizzlepac.hlautils import analyze as filter
```
mcara: @mdlpstsci Why are you redefining/overwriting built-in keyword `filter`?

drizzlepac/alignimages.py
```diff
@@ -0,0 +1,1075 @@
...
+                except KeyError:
+                    asnid = None
+                if asnid and asnid not in retrieve_list:
+                    filelist = aqutils.retrieve_observation(asnid,**pars)
+                    retrieve_list.append(asnid) # record asn's already retrieved
+
+        if len(filelist) > 0:
+            totalInputList += filelist
+
+    if len(filelist) > 0: totalInputList = sorted(
```
mcara: @mdlpstsci This code crashes when input list is empty because `filelist` is undefined. Also, is it really important to check the length of the last (set in the last iteration of the loop above) `filelist` or should it rather be `totalInputList`.

mcara:
On OS X iPython this PR makes `drizzlepac` crash when trying to import it because it now tries to import `matplotlib`.

Also, `hlautils.updatehdr.py` is not called anywhere. Shouldn't it be removed?

drizzlepac/hlautils/analyze.py
```diff
@@ -0,0 +1,265 @@
...
+        if instrume == 'WFC3':
+            sfilter  = (header_data[FILKEY]).upper()
+        # Concatenate the two ACS filter names together with an underscore
+        # If the filter name is blank, skip it
+        if instrume == 'ACS':
+            for filtname in acsFiltNameList:
+
+                # The filter keyword value could be zero or more blank spaces 
+                # Strip off any leading or trailing blanks
+                if len(header_data[filtname].upper().strip()) > 0:
```
mcara: There is no need to convert a string to upper case in order to check whether it is all-white-space. A simple version would be `if header_data[filtname].strip():`

drizzlepac/hlautils/analyze.py
```diff
@@ -0,0 +1,265 @@
...
+            noProcValue = mtflag 
+
+        # Bostrophidon without or with dwell (WFC3 only)
+        elif any ([scan_typ == 'C', scan_typ == 'D']):
+            noProcKey   = SCNKEY
+            noProcValue = scan_typ
+
+        # Filter which does not begin with: 'F'(F###), 'C'(CLEAR), 'N'(N/A), and is not blank
+        # The sfilter variable may be the concatenation of two filters (F160_CLEAR)
+        elif sfilter[0] != 'F' and sfilter[0] != '' and sfilter[0] != 'C' and sfilter[0] != 'N':
```
mcara: @mdlpstsci A simpler and more readable version: `elif sfilter and not sfilter.startswith(('F', 'C', 'N')):` or `elif sfilter and sfilter[0] not in ['F', 'C', 'N']:`. Also, if `sfilter` can be an empty string then the original code will crash when attempting to get the first character `sfilter[0]`.

mcara:
@mdlpstsci @stsci-hack Please do not use tabs when committing code to this repository.

mdlpstsci:
@stsci-hack @dulude Please see all the comments made by Mihai.  He suggests or had to make some changes to accommodate a modified tweakwcs.   I am concerned because the hlautils/updatehdr.py was not the same as the updatehdr found elsewhere.  Also, in my testing the relative paths did not work.  I had to change paths in my working area to be absolute paths.

PR #220, opened 2019-02-08
Update HLA pipeline code to use tweakwcs 0.4.0 API

mcara:
@stsci-hack @mdlpstsci This PR modifies your code in order to take into account API changes in `tweakwcs v0.4.0` - see https://github.com/spacetelescope/tweakwcs/pull/57

If this looks OK to you, feel free to merge this PR.


mcara:
@mdlpstsci @dulude You should try to install this code and run it. Also, that `updatehdr` that was sitting in `hlautils` was an old version of the `drizzlepac/updatehdr.py` file. There is no benefit of having it there. Relative imports are not the issue - see https://github.com/spacetelescope/drizzlepac/pull/221#issuecomment-461940783

stsci-hack:
This PR was superceded by #234, where these changes along with additional ones necessary for pipeline use were implemented as a single PR based on a more recent version of master.

PR #221, opened 2019-02-08
drizzlepac/alignimages.py: changed relative imports to absolute imports

dulude:


jhunkeler:
Why?

mcara:
@dulude Could you please provide some (even a short one) descriptions to your PRs. This is second PR that has no description.

Also, @dulude and @stsci-hack , have you actually tested the code with relative imports and did it fail? I do not understand the theoretical justification for absolute background here.

mcara:
As soon as this PR was merged, readthedocs started failing. @stsci-hack Can you please fix RTD?

dulude:
> @dulude Could you please provide some (even a short one) descriptions to your PRs. This is second PR that has no description.
> 
> Also, @dulude and @stsci-hack , have you actually tested the code with relative imports and did it fail? I do not understand the theoretical justification for absolute background here.

@mcara , Sorry about the lack of descriptions. I'll be more verbose in the future. 
The whole reason I made these changes to the relative imports was because alignimages.py was crashing when it hit the first line with a relative import. I tested this several datasets and it always crashed exactly like this:

    (hstdp-2018.3) rigel:/Users/dulude $ python /Users/dulude/Documents/Code/HLAtransition/drizzlepac/drizzlepac/runastrodriz.py     jcs712010_asn.fits 
    attempted relative import with no known parent package
    ERROR: Cannot run astrodrizzle on jcs712010_asn.fits.
      File "/Users/dulude/Documents/Code/HLAtransition/drizzlepac/drizzlepac/runastrodriz.py", line 600, in main
        align_to_gaia=align_to_gaia)
      File "/Users/dulude/Documents/Code/HLAtransition/drizzlepac/drizzlepac/runastrodriz.py", line 121, in process
        import alignimages
      File "/Users/dulude/Documents/Code/HLAtransition/drizzlepac/drizzlepac/alignimages.py", line 24, in <module>
        from . import updatehdr
    ImportError: attempted relative import with no known parent package
    
    During handling of the above exception, another exception occurred:
    
      File "/Users/dulude/Documents/Code/HLAtransition/drizzlepac/drizzlepac/runastrodriz.py", line 610, in <module>
        main()
      File "/Users/dulude/Documents/Code/HLAtransition/drizzlepac/drizzlepac/runastrodriz.py", line 604, in main
        raise Exception(str(errorobj))
    Exception: attempted relative import with no known parent package

When we had our stand-up this morning, I mentioned the issue. @mdlpstsci said that she had experienced this same issue, and the solution was to replace relative imports with absolute imports, and @stsci-hack was ok with me pushing those changes through to master.

mcara:
Guys, the problem here is not in relative imports but rather in the design of your HLA pipeline module or script? From your example, it is clear that you use `runastrodriz.py` **as a script(!!!)** but it is designed as a **package module**.

You really need to decide what `runastrodriz.py` is: a script or a module? If it is a script, then it should be installed as a script (see, as an example, [`astropy.io.fits` scripts](https://github.com/astropy/astropy/tree/master/astropy/io/fits/scripts)) or, if you want it to be part of the `drizzlepac`, then you should get rid of https://github.com/spacetelescope/drizzlepac/blob/master/drizzlepac/runastrodriz.py#L118-L119 and instead have _relative_ imports from the `drizzlepac` itself.

Right now, the way you have designed this pipeline, `drizzlepac` has to import itself in order to work.

PR #223, opened 2019-02-08
Integrate hla pipeline - Updated for logging

mdlpstsci:
@stsci-hack @dulude I sent an email with some info.  Puzzled.


mdlpstsci:
OK.  I will update the CHANGES.rst.  I have done more research and have improved what goes to stdout/stderr vs the log file.  I am still exploring the improvements, but they can probably wait for another time.

PR #195, opened 2018-11-21
TST: JenkinsfileRT will open issue on failure, remove stsci.convolve from Jenkins

pllim:
JenkinsfileRT (nightly run) will open issue on failure with this patch. Also follow up of #192, where `stsci.convolve` dependency was removed but that was not reflected in the Jenkins files. cc @rendinam @mcara

PR #229, opened 2019-02-11
TST: Unpin pytest. Enable RT auto-notify on failure.

pllim:
Revival of #195 sans `stsci.convolve`.


pllim:
@rendinam , please double check that the `jobconfig` stuff is still correct. @mcara , this should not need a change log.

rendinam:
LGTM

PR #233, opened 2019-02-14
Restore API for perform_align function

stsci-hack:
The interface to perform_align() could not return any values when wrapped by the util.with_logging decorator.  However, the pipeline relies on the returned table to determine what to do next.  This set of changes restores the interface so that perform_align( ) once again returns the filteredTable.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -72,6 +72,8 @@ import numpy as np
 
 import math
 
+import math
+
 
 __taskname__ = "runastrodriz"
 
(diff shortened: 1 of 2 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -72,6 +72,8 @@
 
 import math
 
+import math
+
```
mdlpstsci: Why is this import here twice?

stsci-hack: Actually, it was just one of a few unused import statements which I removed.

PR #230, opened 2019-02-11
Switch to Python 3 style super(); formating adjustments

mcara:
Use Python 3 style `super()` instead of explicitly calling parent class. Use Python 3 class definition style. Fix _some_ formatting issues with the code (i.e., remove unnecessary empty lines).

PR #236, opened 2019-02-15
Fix bug from #230

stsci-hack:
Fixes a bug seen in the regression tests on boyle introduced by PR #230.


mcara:
@stsci-hack It looks like I was overzealous (in #230)...

PR #242, opened 2019-02-22
General cleanup and service integrations

jhunkeler:
N/D


nden:
@jhunkeler Did have to get a plan to use codecov with Jenkins locally or do you use a free trial?

jhunkeler:
Nope, still free. You have to inject the token manually outside of travis/appveyor/etc that's all.

Jenkinsfile
```diff
@@ -1,10 +1,18 @@
...
+    credentialsId: 'drizzlepac-codecov',
+    variable: 'codecov_token')]) {
 // Generate installation compatibility matrix
+/*
 matrix_python = ["3.6", "3.7"]
 matrix_astropy = [">=3.1.0"]
-matrix_numpy = [">=1.14", "==1.15.0"]
+matrix_numpy = ["<1.15", "<1.16"]
+*/
+matrix_python = ["3.7"]
```
rendinam: Was the 3.6 entry here removed on purpose?

jhunkeler: Negative, I'll change it back. I was only avoiding the long wait (and/or halt-and-catch-fire)

PR #247, opened 2019-02-26
Configure stsci-bot to ignore milestone for this repository

drdavella:
This should now be possible thanks to https://github.com/spacetelescope/stsci-bot/pull/11. 🤞


mcara:
Yep, it looks like we are getting closer to the set milestone.

PR #254, opened 2019-03-07
Fix setuptools

jhunkeler:
Latest `setuptools` produces the following error during build:
```        
from setuptools import setup, find_packages, Extension, _install_setup_requires
ImportError: cannot import name '_install_setup_requires'
```

This implements `pyproject.toml` (PEP517) as a replacement.


jhunkeler:
Administrative override. Cannot build this from source using latest tools otherwise.

PR #259, opened 2019-03-08
Ensure singletons do not use the match_relative_fit algorithm

mdlpstsci:
Singletons cannot use the match_relative_fit algorithm because the algorithm requires at least two images - one which servers as the reference and the remaining image(s) are aligned to the reference.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -198,9 +198,6 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
     # Define astrometric catalog list in priority order
     catalogList = ['GAIADR2', 'GAIADR1']
 
-    # Define fitting algorithm list in priority order
-    fit_algorithm_list = [match_relative_fit,match_default_fit,match_2dhist_fit]
-
     # 0: print git info
     if print_git_info:
         log.info("-------------------- STEP 0: Display Git revision info  ------------------------------------------------")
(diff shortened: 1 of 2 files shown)


drizzlepac/alignimages.py
```diff
@@ -331,7 +331,12 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
     best_fitStatusDict={}
     best_fitQual = 5
-    fit_algorithm_list= [match_2dhist_fit,match_default_fit]
+    # The match_relative_fit algorithm must have more than one image as the first image is
+    # the reference for the remaining images.
+    if len(imglist) > 1:
+        fit_algorithm_list=[match_relative_fit, match_2dhist_fit,match_default_fit]
+    else:
+        fit_algorithm_list=[match_2dhist_fit,match_default_fit]
     for catalogIndex in range(0, len(catalogList)): #loop over astrometric catalog
```
jhunkeler: You're are not recording the index value, or using it outside of access to the parent list, so you could simplify this:

```python
for catalog in catalogList:
    log.info("Astrometric Catalog: %s",str(catalog))
```

If you ever plan to record the index some day:

```python
for catalogIndex, catalog in enumerate(catalogList):
        log.info("Astrometric Catalog: {:02d}: {:s}".format(catalogIndex, str(catalog)))
```

mdlpstsci: Deferred to Git Issue #295.

stsci-hack:
This simple change looks as if it would work exactly as needed.

PR #262, opened 2019-03-09
Fix Jenkinsfile

jhunkeler:
While rebasing #242 my changes to the Jenkinsfile were reverted... Fixing it here.

* Enable matrix
* Disable parallel execution to save Jenkins slave resources


jhunkeler:
Hardware just won't cooperate at this time. Closing.

We'll revisit this at a later time.

PR #283, opened 2019-03-19
Mods to add/update regression tests for alignment code

mdlpstsci:
Added the regression tests for the alignment code to the drizzlepac package.  Modified the alignment tests to use the updated output (astropy table) from the alignment code (alignimages.py/perform_align()) to determine the processing state.  Currently, the output states are: Successful, Unsuccessful, Filtered, and Exception.  These tests were written to satisfy the criterion of 0 < total_RMS < 10 mas, so they are not true regression tests.  One routine, test_extract.py, has not been updated pending information needed from the originator of the test.

An update was required in the alignment code itself to accommodate these tests to (1) resolve issues where full filenames (e.g., *_flt.fits) are used for files which are part of associations, but they are to be used as singletons so they cannot be acquired from MAST, and (2) the tests run in the Pytest context which creates its processing directory on-the-fly.  Artifactory is used to store the input files.

Finally, only a few files (after backing out larger number of changes) in support of the original drizzlepac regression test code were modified to avoid name conflicts between ci_watson regression test helper code and home-grown drizzlepac regression test helpers.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -80,80 +80,98 @@ def check_and_get_data(input_list,**pars):
 
     Parameters
     ----------
-    imglist : list
+    input_list : list
         List of one or more calibrated fits images that will be used for catalog generation.
-
-    regtest: Boolean
-        Flag to signal the use of PyTest to run regresssion tests.  When using PyTest with individual images
-        which are part of an association (e.g., *_flt.fits), the files cannot be obtained from MAST as 
-        singletons. They are stored in Artifactory as part of the regression test suite.  
-
-    test_data_path : List of strings  
-        List of comma separated strings defining the path on Artifactory where the *_fl[t|c].fits files are stored.
-        The root of the path is defined by the environment variables TEST_BIGDATA.
-
 
     Returns
     =======
-    input_file_list : list
+    total_input_list: list
         list of full filenames
 
     """
-    totalInputList = []
+    empty_list = []       
+    retrieve_list = []    # Actual files retrieved via astroquery and resident on disk
+    candidate_list = []   # File names gathered from *_asn.fits file
+    ipppssoot_list = []   # ipppssoot names used to avoid duplicate downloads 
+    total_input_list = [] # Output full filename list of data on disk
 
-    # If the regression test flag is True, a regression test has been invoked where the input_list 
-    # actually contains the full filename(s) (e.g., *_flt.fits).  Note that not all regression tests use 
-    # this functionality as Artifactory is accessed for the data source with the use of this flag.  Some
-    # regression tests use astroquery to connect to MAST. This regression testing code is not a scalable
-    # solution, but it is necessary for the alignment portion of the drizzle software.
-    if pars['regtest'] == True:
-        # Artifactory root name and test folder containing the data
-        data_path = pars['test_data_path']
-        from ci_watson.artifactory_helpers import get_bigdata
-        for infile in input_list:
-            full_file_path = get_bigdata(data_path[0],data_path[1],data_path[2],infile)
-            totalInputList.append(infile)
-    else:
-        archive = pars['archive']
-        clobber = pars['clobber']
-        retrieve_list = [] # list of already retrieved ASNs, only request each once
-        for input_item in input_list:
-            try:
-                filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber)
-            except Exception:
-                filelist = []
-            if len(filelist) == 0:
-                # look for local copy of the file
-                input_rootname = input_item[:input_item.find('_')]
-                fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))
-                if len(fitsfilenames) > 0:
-                    imghdu = fits.open(fitsfilenames[0])
-                    imgprimaryheader = imghdu[0].header
-                    try:
-                        asnid = imgprimaryheader['ASN_ID'].strip().lower()
-                        if asnid == 'NONE':
-                            asnid = None
-                    except KeyError:
-                        asnid = None
-                    if asnid and asnid not in retrieve_list:
-                        try:
-                            filelist = aqutils.retrieve_observation(asnid,**pars)
-                            retrieve_list.append(asnid) # record asn's already retrieved
-                        except Exception:
-                            filelist = []
+    # Loop over the input_list to determine if the item in the input_list is a full association file 
+    # (*_asn.fits), a full individual image file (aka singleton, *_flt.fits), or a root name specification 
+    # (association or singleton, ipppssoot).
+    for input_item in input_list:
+        print('Input item: ', input_item)
+        indx = input_item.find('_')
+        
+        # Input with a suffix (_xxx.fits)
+        if indx != -1:
+            lc_input_item = input_item.lower()
+            suffix = lc_input_item[indx+1:indx+4]
+            print('file: ', lc_input_item)
+            # For an association, need to open the table and read the image names as this could
+            # be a custom association.  The assumption is this file is on local disk when specified
+            # in this manner (vs just the ipppssoot of the association).
+            # This "if" block just collects the wanted full file names.
+            if suffix == 'asn':
+                try:
+                    asntab = Table.read(input_item, format='fits')
+                except ValueError:
+                    log.error('File {} not found.'.format(input_item))
+                    return(empty_list)
+                for row in asntab:
+                    if row['MEMTYPE'].startswith('PROD'):
+                        continue
+                    memname = row['MEMNAME'].lower().strip()
+                    # Need to check if the MEMNAME is a full filename or an ipppssoot
+                    if memname.find('_') != -1:
+                        candidate_list.append(memname)
+                    else:
+                        candidate_list.append(memname + '_flc.fits')
+            elif any ([suffix == 'flc', suffix == 'flt']):
+                if lc_input_item not in candidate_list:
+                    candidate_list.append(lc_input_item)
+            else:
+                log.error('Inappropriate file suffix: {}.  Looking for "asn.fits", "flc.fits", or "flt.fits".'.format(suffix))
+                return(empty_list)
 
-            if len(filelist) > 0:
-                totalInputList += filelist
+        # Input is an ipppssoot (association or singleton), nine characters by definition.
+        # This "else" block actually downloads the data specified as ipppssoot.
+        else:
+            if len(input_item) == 9:
+                try:
+                    if input_item not in ipppssoot_list:
+                        # An ipppssoot of an individual file which is part of an association cannot be
+                        # retrieved from MAST
+                        retrieve_list = aqutils.retrieve_observation(input_item,**pars)
 
-        if len(filelist) > 0:
-            # remove duplicate list elements, sort resulting list of unique elements:
-            totalInputList = sorted(list(set(totalInputList)))
+                        # If the retrieved list is not empty, add filename(s) to the total_input_list.
+                        # Also, update the ipppssoot_list so we do not try to download the data again.  Need
+                        # to do this since retrieve_list can be empty because (1) data cannot be acquired (error)
+                        # or (2) data is already on disk (ok).
+                        if retrieve_list:
+                            total_input_list += retrieve_list
+                            ipppssoot_list.append(input_item)
+                        else:
+                            log.error('File {} cannot be retrieved from MAST.'.format(input_item))
+                            return(empty_list)
+                except Exception:
+                    exc_type, exc_value, exc_tb = sys.exc_info()
+                    traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
 
-    log.info("TOTAL INPUT LIST: {}".format(totalInputList))
-    # TODO: add trap to deal with non-existent (incorrect) rootnames
-    # TODO: Address issue about how the code will retrieve association information if there isn't a local file to get 'ASN_ID' header info
-    return(totalInputList)
+    # Only the retrieve_list files via astroquery have been put into the total_input_list thus far.
+    # Now check candidate_list to detect or acquire the requested files from MAST via
+    # astroquery.
+    if candidate_list:
+        for file in candidate_list:
+            # If the file is found on disk, add it to the total_input_list and continue
+            if glob.glob('{}'.format(file)):
+                total_input_list.append(file)
+                continue
+            else:
+                log.error('File {} cannot be found on the local disk.'.format(file))
+                return(empty_list)
 
+    log.info("TOTAL INPUT LIST: {}".format(total_input_list))
+    return(total_input_list)
 
 # ----------------------------------------------------------------------------------------------------------------------
 def perform_align(input_list, **kwargs):
@@ -189,15 +207,6 @@ def perform_align(input_list, **kwargs):
         generate_source_catalogs() generate the .reg region files for every chip of every input image and should
         generate_astrometric_catalog() generate file 'refcatalog.cat'?
  
-    regtest: Boolean
-        Flag to signal the use of PyTest to run regresssion tests.  When using PyTest with individual images
-        which are part of an association (e.g., *_flt.fits), the files cannot be obtained from MAST as 
-        singletons. They are stored in Artifactory as part of the regression test suite.  
-
-    test_data_path : str
-        List of comma separated strings defining the path on Artifactory where the *_fl[t|c].fits files are stored.
-        The root of the path is defined by the environment variables TEST_BIGDATA.
-
     Updates
     -------
     filteredTable: Astropy Table
@@ -210,7 +219,7 @@ def perform_align(input_list, **kwargs):
 
 @util.with_logging
 def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_wcs=False, result=None, runfile=None,
-                  print_fit_parameters=True, print_git_info=False, output=False, regtest=False, test_data_path=None):
+                  print_fit_parameters=True, print_git_info=False, output=False):
 
     log.info("*** HLAPIPELINE Processing Version {!s} ({!s}) started at: {!s} ***\n".format(__version__, __version_date__, util._ptime()[0]))
 
@@ -235,11 +244,12 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
             log.warning("WARNING: Unable to display Git repository revision information.")
 
     print(input_list)
+   
     # 1: Interpret input data and optional parameters
     log.info("-------------------- STEP 1: Get data ------------------------------------------------------------------")
     zeroDT = startingDT = datetime.datetime.now()
     log.info(str(startingDT))
-    imglist = check_and_get_data(input_list, archive=archive, clobber=clobber, regtest=regtest, test_data_path=test_data_path)
+    imglist = check_and_get_data(input_list, archive=archive, clobber=clobber)
     log.info("SUCCESS")
 
     currentDT = datetime.datetime.now()
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -103,10 +103,11 @@ def check_and_get_data(input_list,**pars):
 
     # If the regression test flag is True, a regression test has been invoked where the input_list 
     # actually contains the full filename(s) (e.g., *_flt.fits).  Note that not all regression tests use 
-    # this functionality as Artifactory is accessed for the data source with the use of this flag.
+    # this functionality as Artifactory is accessed for the data source with the use of this flag.  Some
+    # regression tests use astroquery to connect to MAST. This regression testing code is not a scalable
+    # solution, but it is necessary for the alignment portion of the drizzle software.
     if pars['regtest'] == True:
         # Artifactory root name and test folder containing the data
-        remote_root=os.environ['TEST_BIGDATA']
         data_path = pars['test_data_path']
         from ci_watson.artifactory_helpers import get_bigdata
         for infile in input_list:
@@ -119,7 +120,6 @@ def check_and_get_data(input_list,**pars):
         for input_item in input_list:
             try:
                 filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber)
-                #filelist = aqutils.retrieve_observation(input_item)
             except Exception:
                 filelist = []
             if len(filelist) == 0:
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -114,7 +114,7 @@ def check_and_get_data(input_list,**pars):
             if suffix == 'asn':
                 try:
                     asntab = Table.read(input_item, format='fits')
-                except ValueError:
+                except FileNotFoundError:
                     log.error('File {} not found.'.format(input_item))
                     return(empty_list)
                 for row in asntab:
@@ -126,7 +126,7 @@ def check_and_get_data(input_list,**pars):
                         candidate_list.append(memname)
                     else:
                         candidate_list.append(memname + '_flc.fits')
-            elif any ([suffix == 'flc', suffix == 'flt']):
+            elif suffix == 'flc' or suffix == 'flt':
                 if lc_input_item not in candidate_list:
                     candidate_list.append(lc_input_item)
             else:
@@ -135,40 +135,38 @@ def check_and_get_data(input_list,**pars):
 
         # Input is an ipppssoot (association or singleton), nine characters by definition.
         # This "else" block actually downloads the data specified as ipppssoot.
-        else:
-            if len(input_item) == 9:
-                try:
-                    if input_item not in ipppssoot_list:
-                        # An ipppssoot of an individual file which is part of an association cannot be
-                        # retrieved from MAST
-                        retrieve_list = aqutils.retrieve_observation(input_item,**pars)
+        elif len(input_item) == 9:
+            try:
+                if input_item not in ipppssoot_list:
+                    # An ipppssoot of an individual file which is part of an association cannot be
+                    # retrieved from MAST
+                    retrieve_list = aqutils.retrieve_observation(input_item,**pars)
 
-                        # If the retrieved list is not empty, add filename(s) to the total_input_list.
-                        # Also, update the ipppssoot_list so we do not try to download the data again.  Need
-                        # to do this since retrieve_list can be empty because (1) data cannot be acquired (error)
-                        # or (2) data is already on disk (ok).
-                        if retrieve_list:
-                            total_input_list += retrieve_list
-                            ipppssoot_list.append(input_item)
-                        else:
-                            log.error('File {} cannot be retrieved from MAST.'.format(input_item))
-                            return(empty_list)
-                except Exception:
-                    exc_type, exc_value, exc_tb = sys.exc_info()
-                    traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
+                    # If the retrieved list is not empty, add filename(s) to the total_input_list.
+                    # Also, update the ipppssoot_list so we do not try to download the data again.  Need
+                    # to do this since retrieve_list can be empty because (1) data cannot be acquired (error)
+                    # or (2) data is already on disk (ok).
+                    if retrieve_list:
+                        total_input_list += retrieve_list
+                        ipppssoot_list.append(input_item)
+                    else:
+                        log.error('File {} cannot be retrieved from MAST.'.format(input_item))
+                        return(empty_list)
+            except Exception:
+                exc_type, exc_value, exc_tb = sys.exc_info()
+                traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
 
     # Only the retrieve_list files via astroquery have been put into the total_input_list thus far.
     # Now check candidate_list to detect or acquire the requested files from MAST via
     # astroquery.
-    if candidate_list:
-        for file in candidate_list:
-            # If the file is found on disk, add it to the total_input_list and continue
-            if glob.glob('{}'.format(file)):
-                total_input_list.append(file)
-                continue
-            else:
-                log.error('File {} cannot be found on the local disk.'.format(file))
-                return(empty_list)
+    for file in candidate_list:
+        # If the file is found on disk, add it to the total_input_list and continue
+        if glob.glob(file):
+            total_input_list.append(file)
+            continue
+        else:
+            log.error('File {} cannot be found on the local disk.'.format(file))
+            return(empty_list)
 
     log.info("TOTAL INPUT LIST: {}".format(total_input_list))
     return(total_input_list)
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -83,6 +83,16 @@ def check_and_get_data(input_list,**pars):
     imglist : list
         List of one or more calibrated fits images that will be used for catalog generation.
 
+    regtest: Boolean
+        Flag to signal the use of PyTest to run regresssion tests.  When using PyTest with individual images
+        which are part of an association (e.g., *_flt.fits), the files cannot be obtained from MAST as 
+        singletons. They are stored in Artifactory as part of the regression test suite.  
+
+    test_data_path : List of strings  
+        List of comma separated strings defining the path on Artifactory where the *_fl[t|c].fits files are stored.
+        The root of the path is defined by the environment variables TEST_BIGDATA.
+
+
     Returns
     =======
     input_file_list : list
@@ -90,38 +100,55 @@ def check_and_get_data(input_list,**pars):
 
     """
     totalInputList = []
-    retrieve_list = [] # list of already retrieved ASNs, only request each once
-    for input_item in input_list:
-        try:
-            filelist = aqutils.retrieve_observation(input_item,**pars)
-        except Exception:
-            filelist = []
-        if len(filelist) == 0:
-            # look for local copy of the file
-            input_rootname = input_item[:input_item.find('_')]
-            fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))
-            if len(fitsfilenames) > 0:
-                imghdu = fits.open(fitsfilenames[0])
-                imgprimaryheader = imghdu[0].header
-                try:
-                    asnid = imgprimaryheader['ASN_ID'].strip().lower()
-                    if asnid == 'NONE':
-                        asnid = None
-                except KeyError:
-                    asnid = None
-                if asnid and asnid not in retrieve_list:
+
+    # If the regression test flag is True, a regression test has been invoked where the input_list 
+    # actually contains the full filename(s) (e.g., *_flt.fits).  Note that not all regression tests use 
+    # this functionality as Artifactory is accessed for the data source with the use of this flag.
+    if pars['regtest'] == True:
+        # Artifactory root name and test folder containing the data
+        remote_root=os.environ['TEST_BIGDATA']
+        data_path = pars['test_data_path']
+        from ci_watson.artifactory_helpers import get_bigdata
+        for infile in input_list:
+            full_file_path = get_bigdata(data_path[0],data_path[1],data_path[2],infile)
+            totalInputList.append(infile)
+    else:
+        archive = pars['archive']
+        clobber = pars['clobber']
+        retrieve_list = [] # list of already retrieved ASNs, only request each once
+        for input_item in input_list:
+            try:
+                filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber)
+                #filelist = aqutils.retrieve_observation(input_item)
+            except Exception:
+                filelist = []
+            if len(filelist) == 0:
+                # look for local copy of the file
+                input_rootname = input_item[:input_item.find('_')]
+                fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))
+                if len(fitsfilenames) > 0:
+                    imghdu = fits.open(fitsfilenames[0])
+                    imgprimaryheader = imghdu[0].header
                     try:
-                        filelist = aqutils.retrieve_observation(asnid,**pars)
-                        retrieve_list.append(asnid) # record asn's already retrieved
-                    except Exception:
-                        filelist = []
+                        asnid = imgprimaryheader['ASN_ID'].strip().lower()
+                        if asnid == 'NONE':
+                            asnid = None
+                    except KeyError:
+                        asnid = None
+                    if asnid and asnid not in retrieve_list:
+                        try:
+                            filelist = aqutils.retrieve_observation(asnid,**pars)
+                            retrieve_list.append(asnid) # record asn's already retrieved
+                        except Exception:
+                            filelist = []
+
+            if len(filelist) > 0:
+                totalInputList += filelist
 
         if len(filelist) > 0:
-            totalInputList += filelist
+            # remove duplicate list elements, sort resulting list of unique elements:
+            totalInputList = sorted(list(set(totalInputList)))
 
-    if len(filelist) > 0:
-        # remove duplicate list elements, sort resulting list of unique elements:
-        totalInputList = sorted(list(set(totalInputList)))
     log.info("TOTAL INPUT LIST: {}".format(totalInputList))
     # TODO: add trap to deal with non-existent (incorrect) rootnames
     # TODO: Address issue about how the code will retrieve association information if there isn't a local file to get 'ASN_ID' header info
@@ -161,6 +188,15 @@ def perform_align(input_list, **kwargs):
         Should utils.astrometric_utils.create_astrometric_catalog() generate file 'ref_cat.ecsv' and should
         generate_source_catalogs() generate the .reg region files for every chip of every input image and should
         generate_astrometric_catalog() generate file 'refcatalog.cat'?
+ 
+    regtest: Boolean
+        Flag to signal the use of PyTest to run regresssion tests.  When using PyTest with individual images
+        which are part of an association (e.g., *_flt.fits), the files cannot be obtained from MAST as 
+        singletons. They are stored in Artifactory as part of the regression test suite.  
+
+    test_data_path : str
+        List of comma separated strings defining the path on Artifactory where the *_fl[t|c].fits files are stored.
+        The root of the path is defined by the environment variables TEST_BIGDATA.
 
     Updates
     -------
@@ -174,7 +210,7 @@ def perform_align(input_list, **kwargs):
 
 @util.with_logging
 def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_wcs=False, result=None, runfile=None,
-                  print_fit_parameters=True, print_git_info=False, output=False):
+                  print_fit_parameters=True, print_git_info=False, output=False, regtest=False, test_data_path=None):
 
     log.info("*** HLAPIPELINE Processing Version {!s} ({!s}) started at: {!s} ***\n".format(__version__, __version_date__, util._ptime()[0]))
 
@@ -198,11 +234,12 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
         else:
             log.warning("WARNING: Unable to display Git repository revision information.")
 
+    print(input_list)
     # 1: Interpret input data and optional parameters
     log.info("-------------------- STEP 1: Get data ------------------------------------------------------------------")
     zeroDT = startingDT = datetime.datetime.now()
     log.info(str(startingDT))
-    imglist = check_and_get_data(input_list, archive=archive, clobber=clobber)
+    imglist = check_and_get_data(input_list, archive=archive, clobber=clobber, regtest=regtest, test_data_path=test_data_path)
     log.info("SUCCESS")
 
     currentDT = datetime.datetime.now()
--- a/tests/hla/base_classes.py
+++ b/tests/hla/base_classes.py
@@ -12,6 +12,7 @@ from drizzlepac.hlautils.astroquery_utils import retrieve_observation
 
 # Base classes for actual tests.
 # NOTE: Named in a way so pytest will not pick them up here.
+@pytest.mark.bigdata
 class BaseTest(object):
     prevdir = os.getcwd()
     use_ftp_crds = True
--- a/tests/hla/base_classes.py
+++ b/tests/hla/base_classes.py
@@ -1,0 +1,178 @@
+"""Regression test helpers."""
+import os
+
+import pytest
+from astropy.io import fits
+from astropy.utils.data import conf
+
+from ci_watson.artifactory_helpers import get_bigdata
+from ci_watson.artifactory_helpers import compare_outputs
+
+from drizzlepac.hlautils.astroquery_utils import retrieve_observation
+
+# Base classes for actual tests.
+# NOTE: Named in a way so pytest will not pick them up here.
+class BaseTest(object):
+    prevdir = os.getcwd()
+    use_ftp_crds = True
+    timeout = 30  # seconds
+    tree = 'dev'
+    docopy = False
+
+    # Numpy default for allclose comparison
+    rtol = 1e-6
+    atol = 0
+
+    # To be defined by instrument/test
+    input_repo = ''  # e.g., 'drizzlepac' or 'jwst-pipeline'
+    results_root = None  # e.g., 'drizzlepac-results'
+    input_loc = ''  # root directory for 'input' files
+    ref_loc = []    # root path for 'truth' files: ['test1','truth'] or ['test3']
+    ignore_keywords = []
+    ignore_table_keywords = []
+    ignore_fields = []
+    ignore_hdus = []
+
+    # To be defined by individual test
+    subdir = ''
+    curdir = ''
+
+    @pytest.fixture(autouse=True)
+    def setup_class(self, tmpdir, envopt):
+        """
+        Run test in own dir so we can keep results separate from
+        other tests.
+        """
+        # create working directory specified for the test
+        if not tmpdir.ensure(self.subdir, dir=True):
+            curdir = tmpdir.mkdir(self.subdir).strpath
+        else:
+            curdir = tmpdir.join(self.subdir).strpath
+        os.chdir(curdir)
+        self.curdir = curdir
+
+        # This controls astropy.io.fits timeout
+        conf.remote_timeout = self.timeout
+
+        # Update tree to point to correct environment
+        self.tree = envopt
+
+        # Configure environment for tests
+        self.set_environ()
+
+    def teardown_class(self):
+        """Reset path and variables."""
+        conf.reset('remote_timeout')
+        os.chdir(self.prevdir)
+        if self.use_ftp_crds and self.prevref is not None:
+            os.environ[self.refstr] = self.prevref
+
+    def set_environ(self):
+        """Set environment variables for test environment to be used
+
+           Each class/test will define what needs to be set, if anything.
+           For example, control use of Astrometry updates for HST with:
+               os.environ['ASTROMETRY_STEP_CONTROL'] = 'OFF'
+        """
+        pass
+
+    def get_input_path(self):
+        """
+        Return path within repository to remote source of input data
+        """
+        path = [self.input_repo, self.tree, self.input_loc]
+
+        return path
+
+    def get_data(self, *args, **kwargs):
+        """
+        Download `filename` into working directory using
+        `artifactory_helpers/get_bigdata()` or 
+        `astroquery_utils.retrieve_observation`.  Use of `astroquery_utils`
+	will allow getting data directly from MAST via astroquery.
+
+        Returns
+        --------
+        local_files : list
+            This will return a list of all the files downloaded with 
+            the full path to the local copy of the file.
+        """
+        if len(args[0]) == 9: # Only a rootname provided
+            local_files = retrieve_observation(args[0])
+        else:
+            # If user has specified action for no_copy, apply it with
+            # default behavior being whatever was defined in the base class.
+            docopy = kwargs.get('docopy', self.docopy)
+            local_files = get_bigdata(*self.get_input_path(),
+                                     *args,
+                                     docopy=docopy)
+            local_files = [local_files]
+
+        return local_files
+
+    def raw_from_asn(self, asn_file):
+        """
+        Return the list of  member exposures specified in the
+        association file.
+
+        .. WARNING::
+        This method needs to be defined by each subclass.
+        """
+        msg = "Sub-class needs to define this method."
+        raise NotImplementedError(msg)
+
+
+    def get_input_file(self, *args, refsep='$', **kwargs):
+        """
+        Download or copy input file (e.g., RAW) into the working directory.
+        Can be differentiated from 'get_data' by overloading with version for
+        the test that performs additional steps, such as identifying and
+        downloading calibration ref files from CRDS (HST does not do this
+        automatically).
+        """
+        # If user has specified action for no_copy, apply it with
+        # default behavior being whatever was defined in the base class.
+        docopy = kwargs.get('docopy', self.docopy)
+
+        self.get_data(*args, docopy=docopy)
+
+    def compare_outputs(self, outputs, raise_error=True, **kwargs):
+
+        # Parse any user-specified kwargs
+        ignore_keywords = kwargs.get('ignore_keywords', self.ignore_keywords)
+        ignore_hdus = kwargs.get('ignore_hdus', self.ignore_hdus)
+        ignore_fields = kwargs.get('ignore_fields', self.ignore_fields)
+        rtol = kwargs.get('rtol', self.rtol)
+        atol = kwargs.get('atol', self.atol)
+
+        compare_kws = dict(ignore_fields=ignore_fields, ignore_hdus=ignore_hdus,
+                        ignore_keywords=ignore_keywords,
+                        rtol=rtol, atol=atol)
+
+        input_path = [self.input_repo, self.tree, self.input_loc, *self.ref_loc]
+
+        return compare_outputs(outputs, raise_error=True,
+                               input_path=input_path,
+                               docopy = self.docopy,
+                               results_root = self.results_root,
+                               **compare_kws)
+
+def get_hdu(filename):
+    """Return the HDU for the file and extension specified in the filename.
+
+       This routine expects the filename to be of the format:
+           <filename>.fits[extn]
+
+        For example, "jw99999-a3001_t1_nircam_f140m-maskbar_i2d.fits[hdrtab]"
+    """
+    froot, fextn = filename.split('[')
+    fextn = fextn.replace(']','')
+    fits_file = fits.open(froot)
+    return fits_file[fextn]
+
+def build_hdulist(filename, extn_list):
+    """Create a new HDUList object based on extensions specified in extn_list"""
+    f = fits.open(filename)
+    fhdu = [f[extn] for extn in extn_list]
+
+    return fhdu
--- a/tests/hla/base_classes.py
+++ b/tests/hla/base_classes.py
@@ -172,8 +172,8 @@ def get_hdu(filename):
     return fits_file[fextn]
 
 def build_hdulist(filename, extn_list):
-    """Create a new HDUList object based on extensions specified in extn_list"""
-    f = fits.open(filename)
-    fhdu = [f[extn] for extn in extn_list]
+    """Create a new list object based on extensions specified in extn_list"""
+    with fits.open(filename) as f:
+        fhdu = [f[extn] for extn in extn_list]
 
     return fhdu
--- a/tests/hla/base_test.py
+++ b/tests/hla/base_test.py
@@ -1,0 +1,405 @@
+import os
+import pytest
+import math
+
+from astropy.io import fits
+from astropy.table import Table
+
+import numpy as np
+import stwcs
+from stwcs import updatewcs
+from stsci.tools import fileutil
+
+from ci_watson.artifactory_helpers import get_bigdata_root
+from ci_watson.hst_helpers import raw_from_asn, ref_from_image, download_crds
+try:
+    from ci_watson.artifactory_helpers import check_url
+except ImportError:
+    from ci_watson.artifactory_helpers import _is_url as check_url
+
+from .base_classes import BaseTest
+
+__all__ = ['BaseHLATest', 'BaseHLAParTest', 'centroid_compare', 'BaseUnit']
+
+@pytest.mark.usefixtures('_jail')
+class BaseHLATest(BaseTest):
+    ignore_hdus = []
+    input_repo = 'hst-hla-pipeline'
+    results_root = 'hst-hla-pipeline-results'
+    output_shift_file = None
+    fit_limit = 0.010 # 10 milli-arcseconds
+    
+    docopy = False  # Do not make additional copy by default
+    rtol = 1e-6
+
+    refstr = 'jref'
+    prevref = os.environ.get(refstr)
+
+    ignore_keywords = ['origin', 'filename', 'date', 'iraf-tlm', 'fitsdate',
+                       'upwtim', 'wcscdate', 'upwcsver', 'pywcsver',
+                       'history', 'prod_ver', 'rulefile']
+
+    reffile_lookup = ['IDCTAB', 'OFFTAB', 'NPOLFILE', 'D2IMFILE', 'DGEOFILE']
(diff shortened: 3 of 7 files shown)


stscicrawford:
It would be good to have @mcara and @jhunkeler review this

jhunkeler:
#285 implemented full use of `ci_watson` and removed redundant code. Please rebase.

tests/hla/base_classes.py
```diff
@@ -0,0 +1,178 @@
...
+from astropy.utils.data import conf
+
+from ci_watson.artifactory_helpers import get_bigdata
+from ci_watson.artifactory_helpers import compare_outputs
+
+from drizzlepac.hlautils.astroquery_utils import retrieve_observation
+
+# Base classes for actual tests.
+# NOTE: Named in a way so pytest will not pick them up here.
+class BaseTest(object):
```
jhunkeler: Need a `@pytest.mark.bigdata` here

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+
+    # If the regression test flag is True, a regression test has been invoked where the input_list 
+    # actually contains the full filename(s) (e.g., *_flt.fits).  Note that not all regression tests use 
+    # this functionality as Artifactory or a local disk is accessed for the data source with the use of 
+    # this flag.  Some regression tests use astroquery to connect to MAST. This regression testing code 
+    # is not a scalable solution, but it is necessary for the alignment portion of the drizzle software.
+    if pars['regtest'] == True:
+        # Test folder containing the data - the root path is defined by the environment variable TEST_BIGDATA
+        data_path = pars['test_data_path']
+        from ci_watson.artifactory_helpers import get_bigdata
```
jhunkeler: `ci_watson.get_bigdata`'s use at runtime is determined by `@pytest.mark.bigdata` so we should never need to nest this inside of a branch. Import this at the top of the module.

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+    else:
+        archive = pars['archive']
+        clobber = pars['clobber']
+        retrieve_list = [] # list of already retrieved ASNs, only request each once
+        for input_item in input_list:
+            try:
+                filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber)
+            except Exception:
+                filelist = []
+            if len(filelist) == 0:
```
jhunkeler:
```python
for input_item in input_list:
    filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber)
    if not filelist:
        # ...
```
Or 

```python
filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber) or []
```

Or 

```python
filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber) or list()
```

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+            full_file_path = get_bigdata(data_path[0],data_path[1],data_path[2],infile)
+            totalInputList.append(infile)
+    else:
+        archive = pars['archive']
+        clobber = pars['clobber']
+        retrieve_list = [] # list of already retrieved ASNs, only request each once
+        for input_item in input_list:
+            try:
+                filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber)
+            except Exception:
```
jhunkeler: `except Exception:` is the same as writing `except:`. If this `try/except` block is required due to how `aqutils.retrieve_observation` works, then the exception raised should be trapped explicitly (i.e. `except aqutils.OhNoes:`).

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+                filelist = []
+            if len(filelist) == 0:
+                # look for local copy of the file
+                input_rootname = input_item[:input_item.find('_')]
+                fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))
+                if len(fitsfilenames) > 0:
```
jhunkeler: Length check not required.

```python
if fitsfilenames:
    # ...
```

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+                    except KeyError:
+                        asnid = None
+                    if asnid and asnid not in retrieve_list:
+                        try:
+                            filelist = aqutils.retrieve_observation(asnid,**pars)
+                            retrieve_list.append(asnid) # record asn's already retrieved
+                        except Exception:
+                            filelist = []
+
+            if len(filelist) > 0:
```
jhunkeler: Length check not required:

```python
if filelist:
    # ...
```

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+                totalInputList += filelist
 
         if len(filelist) > 0:
```
jhunkeler:
```python
if filelist:
    # ...
```

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+                if len(fitsfilenames) > 0:
+                    imghdu = fits.open(fitsfilenames[0])
+                    imgprimaryheader = imghdu[0].header
                     try:
-                        filelist = aqutils.retrieve_observation(asnid,**pars)
-                        retrieve_list.append(asnid) # record asn's already retrieved
-                    except Exception:
-                        filelist = []
+                        asnid = imgprimaryheader['ASN_ID'].strip().lower()
+                        if asnid == 'NONE':
```
jhunkeler: Will never be `NONE` due to `.lower()`

tests/hla/test_align.py
```diff
@@ -0,0 +1,215 @@
...
+        try:
+            datasetTable = alignimages.perform_align(input_filenames,archive=False,clobber=True,debug=False,
+                update_hdr_wcs=False,print_fit_parameters=True,print_git_info=False,output=False,regtest=True,
+                test_data_path=data_path)
+
+            # Examine the output table to extract the RMS for the entire fit and the compromised information
+            if (datasetTable):
+                totalRMS = datasetTable['total_rms'][0]
+
+        except Exception:
```
jhunkeler: These `traceback` blocks prevent `pytest` from generating meaningful tracebacks after a failed test. I'm only commenting on this one but it applies to all of them.

Use:
```python
# Fail if exception of `SomeExceptionHere` is not raised
with not pytest.raises(SomeExceptionHere):
    # code here

# Pass if exception of type `SomeExceptionHere` is raised:
with pytest.raises(SomeExceptionHere):
    # code here
```

In all other cases let pytest parse the traceback and record it correctly.

tests/hla/test_align.py
```diff
@@ -0,0 +1,215 @@
...
+        self.curdir = os.getcwd()
+        totalRMS = 0.0
+        data_path = ['hst-hla-pipeline','dev','base_tests']
+        try:
+            datasetTable = alignimages.perform_align(input_filenames,archive=False,clobber=True,debug=False,
+                update_hdr_wcs=False,print_fit_parameters=True,print_git_info=False,output=False,regtest=True,
+                test_data_path=data_path)
+
+            # Examine the output table to extract the RMS for the entire fit and the compromised information
+            if (datasetTable):
```
jhunkeler: `if (value):` discouraged use `if value:` unless checking multiple condition groups.

tests/hla/test_extract.py
```diff
@@ -0,0 +1,179 @@
...
+
+            # Align images and correct WCS
+            tweakwcs.tweak_image_wcs(imglist, refcat, match=match)
+            for chip in imglist:
+                status = chip.meta['tweakwcs_info']['status']
+                if status.startswith('FAIL'):
+                    failures = True
+                    print("STATUS for {}: {}".format(chip.meta['filename'], status))
+                    break
+        except Exception:
```
jhunkeler: Blank exceptions are bad. Any failue real or imaged will simply write out `Failed to align xyz.ext`. At the very least do this:

```python
try:
    # something
except Exception as e:
    print("ALIGNMENT EXCEPTION: Failed to align {} due to {}".format(infile, e))
```

tests/hla/test_randomlist.py
```diff
@@ -0,0 +1,224 @@
...
+            except Exception:
+           
+                exc_type, exc_value, exc_tb = sys.exc_info()
+                traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
+                print("TEST_RANDOM. Exception Dataset: ", dataset, "\n")
+                numException += 1
+                continue
+
+        # Perform some clean up
+        if os.path.exists('ref_cat.ecsv'): os.remove('ref_cat.ecsv')
```
jhunkeler: `if ...:\n`

tests/hla/test_align.py
```diff
@@ -0,0 +1,215 @@
...
+
+        except Exception:
+            exc_type, exc_value, exc_tb = sys.exc_info()
+            traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
+
+        assert (0.0 < totalRMS <= RMS_LIMIT)
+
+    def test_astroquery(self):
+        """Verify that new astroquery interface will work"""
+        #self.curdir = os.getcwd()
```
jhunkeler: Dead code

tests/hla/base_classes.py
```diff
@@ -0,0 +1,178 @@
...
+        `astroquery_utils.retrieve_observation`.  Use of `astroquery_utils`
+	will allow getting data directly from MAST via astroquery.
+
+        Returns
+        --------
+        local_files : list
+            This will return a list of all the files downloaded with 
+            the full path to the local copy of the file.
+        """
+        if len(args[0]) == 9: # Only a rootname provided
```
jhunkeler: If that length ever changes...

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+            except Exception:
+                filelist = []
+            if len(filelist) == 0:
+                # look for local copy of the file
+                input_rootname = input_item[:input_item.find('_')]
+                fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))
+                if len(fitsfilenames) > 0:
+                    imghdu = fits.open(fitsfilenames[0])
+                    imgprimaryheader = imghdu[0].header
                     try:
```
jhunkeler: Might be easier to set `asnid = None` to start with, then check if it's still `None` at the end.

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
+            try:
+                filelist = aqutils.retrieve_observation(input_item,archive=archive,clobber=clobber)
+            except Exception:
+                filelist = []
+            if len(filelist) == 0:
+                # look for local copy of the file
+                input_rootname = input_item[:input_item.find('_')]
+                fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))
+                if len(fitsfilenames) > 0:
+                    imghdu = fits.open(fitsfilenames[0])
```
mcara: Files should be open with `with`-statement. This way they are closed automatically. Your code leave all file handles open until the function returns.

drizzlepac/alignimages.py
mcara: Not only that, but the previous `if` block establishes that `len(filelist)` is not 0 and therefore it _must_ be greater that 0 and therefore this `if` should be replaced with `else`.

drizzlepac/alignimages.py
mcara: I am afraid it is a wrong variable name. I have a feeling it should have been `if totalInputList:` instead of `if filelist:`

drizzlepac/alignimages.py
```diff
@@ -1169,7 +1206,16 @@ def interpret_fit_rms(tweakwcs_output, reference_catalog):
...
+                    'accommodate use of PyTest to run regression tests.  When using PyTest with individual images '
+                    'which are part of an association (e.g., *_flt.fits), the files cannot be obtained from MAST '
+                    'as singletons. They are stored in Artifactory or local disk as part of the regression test suite. The '
+                    'default is "False".')
+
+    PARSER.add_argument( '-t', '--raw_test_data_path', required=False,default=None,help='A comma separated list of strings '
+                    'defining the path on Artifactory or local disk where the *_fl[t|c].fits files are stored.  The root of '
+                    'the path is defined by the environment variables TEST_BIGDATA. ["hst-hla-pipeline","dev","base_tests"].')
 
     ARGS = PARSER.parse_args()
```
mcara: Why so many variables have either `hungarianNotation` or are `ALLCAPS`?

drizzlepac/alignimages.py
```diff
@@ -1169,7 +1206,16 @@ def interpret_fit_rms(tweakwcs_output, reference_catalog):
                     'generate_source_catalogs() generate the .reg region files for every chip of every input image and '
                     'should generate_astrometric_catalog() generate file "refcatalog.cat"? Unless explicitly set, the '
                     'default is "False".')
-    'Should the code generate '
+
+    PARSER.add_argument( '-r', '--regtest', required=False,choices=['True','False'],default='False',help='Flag to '
```
mcara: I **strongly** recommend the advice from https://stackoverflow.com/a/15008806 be followed (also see `'store_true'` and `'store_false'` in https://docs.python.org/3/library/argparse.html#action). This will make the code behave as most people expect, get rid of https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L1233-L1248 and of the function in https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L160-L172 which, by the way, is unnecessarily complicated: you could simply have:
```python
def convert_string_tf_to_boolean(invalue):
    return invalue == 'True'
```
(which should be done directly where it is needed - no need to define a function for such a simple statement).

mcara:
Also, looking at the `alignimages.py` code, the following code:
https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L92
https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L103-L111
can be replaced with:
```pyton
retrieve_list = ['none']
# ..........
if fitsfilenames:
    with fits.open(fitsfilenames[0]) as hdulist:
        asnid = hdulist[0].get('ASN_ID', 'none').lower().strip()
        if asnid not in retrieve_list:
            # ..........
```

mcara:
Another optimization:
https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L301-L304

can be replaced with 
```python
total_num_sources  = sum(map(len, table.values()))
```

drizzlepac/alignimages.py
mcara: Also see https://github.com/spacetelescope/drizzlepac/pull/283#issuecomment-476452356

mcara:
Nested `if`s such as https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L460-L461 _generally speaking_ can be combined into a single statement such as:
```python
if not tweakwcs_info_key.startswith("matched") and tweakwcs_info_key.lower() == 'rms':
```
However, in this specific case, if `tweakwcs_info_key.lower() == 'rms'` is true, then `tweakwcs_info_key.startswith("matched")` cannot be true and, therefore, the first `if` in the original code can be simply dropped (i.e., leave only this):
```python
if tweakwcs_info_key.lower() == 'rms':
```

mcara:
Code such as this:
https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L476-L483
should not check if a boolean value is `True` or `False` - a boolean can be used directly in an `if`.

Moreover, this code can be simplified as follows:
```python
filteredTable[index]['status'] = int(not fitStatusDict[explicitDictKey]['valid'])
filteredTable['compromised'] = int(fitStatusDict[explicitDictKey]['compromised'])
```

mcara:
`"{}".format(sciExtCtr)` in

https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L962-L964

can be replaced with `str(sciExtCtr)`. In fact, the above three lines can be replaced with the following one-liner (albeit a long one):
```python
sciExtDict = {str(k): fileutil.findExtname(hdulist, 'sci', extver=k) for k in range(1, num_sci_ext + 1)}
```

mcara:
This code:
https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L842-L848
 can be simplified to:
```python
pars['output'] = 'ref_cat.ecsv' if pars['output'] else None
out_catalog = amutils.create_astrometric_catalog(imglist, **pars)
```
All that copying is not necessary.

mcara:
In https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L916

`if chip_cat:` and  `if len(chip_cat) > 0:` are equivalent though the first syntax is Pythonical.

mcara:
Probably this:
https://github.com/spacetelescope/drizzlepac/blob/f6c483f5dfda7f3ed7c0725500da9f4381282176/drizzlepac/hlautils/astrometric_utils.py#L634-L635
would be clearer as:
```python
for chip in range(1, numSci + 1):
```

mcara:
https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L982-L985

Second `if` makes sense only if first is `False` and therefore should be replaced with `elif`. Also, what should happen if suffix/extension is neither `flt.fits` nor `flc.fits` such as `c0m.fits`?

Also, in `imageName.replace("flc", "flt_hlet")` is `flc` being replaced with `flt`? Is this a typo?

The following line could replace the above code:

```python
headerlet_filename = '_hlet'.join(path.splitext(imageName))
```

mcara:
Centroid algorithm - see
https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/tests/hla/base_test.py#L191-L211
probably can be made more efficient as:
```python
    def centroid(self, image, size, center):
        ylo = max(0, int(center[0]) - size // 2)
        yhi = min(ylo + size, image.shape[0])
        xlo = max(int(center[1]) - size // 2, 0)
        xhi = min(xlo + size, image.shape[1])

        cutout = image[ylo:yhi, xlo:xhi]

        flux = np.sum(cutout)
        if flux == 0.0:
            return None

        y, x = np.sum(cutout * np.indices(cutout.shape), axis=(1, 2)) / flux
        return (x, y, flux)
```

tests/hla/base_classes.py
mdlpstsci: Done

drizzlepac/alignimages.py
mdlpstsci: Reworked this code so download of files specified with full file names is done in the test routine and not in this code.   There is no longer a need to import anything from ci_watson (test support utilities) to contaminate this code.

drizzlepac/alignimages.py
mdlpstsci: Done

drizzlepac/alignimages.py
mdlpstsci: Done

drizzlepac/alignimages.py
mdlpstsci: Done and see following issue/comment.

drizzlepac/alignimages.py
mdlpstsci: I agree it should be "if totalInputList". Done.

drizzlepac/alignimages.py
mdlpstsci: Changed 'NONE' to 'none'.  I need to address bugs/errors in this PR.  Better coding practices and efficiency improvements will be handled in a different Issue which references this PR.

tests/hla/test_extract.py
mdlpstsci: Done

tests/hla/test_align.py
mdlpstsci: Done

tests/hla/test_randomlist.py
mdlpstsci: Done

tests/hla/test_align.py
mdlpstsci: Removed

tests/hla/base_classes.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    froot, fextn = filename.split('[')
+    fextn = fextn.replace(']','')
+    fits_file = fits.open(froot)
+    return fits_file[fextn]
+
+def build_hdulist(filename, extn_list):
+    """Create a new HDUList object based on extensions specified in extn_list"""
+    f = fits.open(filename)
+    fhdu = [f[extn] for extn in extn_list]
```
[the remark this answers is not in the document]
mdlpstsci: Done

tests/hla/base_classes.py
```diff
@@ -0,0 +1,179 @@
...
+        For example, "jw99999-a3001_t1_nircam_f140m-maskbar_i2d.fits[hdrtab]"
+    """
+    froot, fextn = filename.split('[')
+    fextn = fextn.replace(']','')
+    fits_file = fits.open(froot)
+    return fits_file[fextn]
+
+def build_hdulist(filename, extn_list):
+    """Create a new HDUList object based on extensions specified in extn_list"""
+    f = fits.open(filename)
```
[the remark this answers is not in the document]
mdlpstsci: Cool and Done

tests/hla/base_classes.py
mdlpstsci: Understood, but in this case I am willing to risk it.

tests/hla/test_extract.py
```diff
@@ -0,0 +1,179 @@
...
+                chip_id = chip.meta.get('chip_id',1)
+
+                # determine 30mas limit in pixels
+                xylimit = 30
+                # Perform comparisons
+                nmatches = tweak_info['nmatches']
+                xrms,yrms = tweak_info['rms']
+                xrms *= chip.wcs.pscale*1000
+                yrms *= chip.wcs.pscale*1000
+                if any([nmatches==0, xrms>xylimit, yrms>xylimit]):
```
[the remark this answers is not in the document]
mdlpstsci: test_extract.py is to be deprecated.

tests/hla/test_extract.py
```diff
@@ -0,0 +1,179 @@
...
+            for chip in imglist:
+                tweak_info = chip.meta.get('tweakwcs_info', None)
+                chip_id = chip.meta.get('chip_id',1)
+
+                # determine 30mas limit in pixels
+                xylimit = 30
+                # Perform comparisons
+                nmatches = tweak_info['nmatches']
+                xrms,yrms = tweak_info['rms']
+                xrms *= chip.wcs.pscale*1000
```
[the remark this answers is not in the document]
mdlpstsci: test_extract.py is to be deprecated.

tests/hla/test_extract.py
```diff
@@ -0,0 +1,179 @@
...
+
+        if not failures:
+            for chip in imglist:
+                tweak_info = chip.meta.get('tweakwcs_info', None)
+                chip_id = chip.meta.get('chip_id',1)
+
+                # determine 30mas limit in pixels
+                xylimit = 30
+                # Perform comparisons
+                nmatches = tweak_info['nmatches']
```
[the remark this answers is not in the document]
mdlpstsci: test_extract.py is to be deprecated.

tests/hla/test_extract.py
```diff
@@ -0,0 +1,179 @@
...
+
+        """
+        self.input_loc = 'catalog_tests'
+        self.curdir = os.getcwd()
+        truth_path = [self.input_repo, self.tree, self.input_loc, *self.ref_loc]
+
+        if not isinstance(input_filenames, list):
+            input_filenames = [input_filenames]
+
+        try:
```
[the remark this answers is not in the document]
mdlpstsci: test_extract.py is to be deprecated.

tests/hla/test_align.py
```diff
@@ -0,0 +1,215 @@
...
+            OR
+            export TEST_BIGDATA=/Users/YourNameHere/TestDataDirectory/
+     
+        This test file can be executed in the following manner:
+            $ pytest -s --bigdata test_align.py >& test_align_output.txt &
+            $ tail -f test_align_output.txt
+
+    """
+
+    ref_loc = ['truth']
```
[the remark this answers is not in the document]
mdlpstsci: Done

drizzlepac/alignimages.py
```diff
@@ -82,45 +82,72 @@ def check_and_get_data(input_list,**pars):
...
-                except KeyError:
-                    asnid = None
-                if asnid and asnid not in retrieve_list:
+
+    # If the regression test flag is True, a regression test has been invoked where the input_list 
+    # actually contains the full filename(s) (e.g., *_flt.fits).  Note that not all regression tests use 
+    # this functionality as Artifactory or a local disk is accessed for the data source with the use of 
+    # this flag.  Some regression tests use astroquery to connect to MAST. This regression testing code 
+    # is not a scalable solution, but it is necessary for the alignment portion of the drizzle software.
+    if pars['regtest'] == True:
```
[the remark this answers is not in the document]
mdlpstsci: Done

mcara:
While reviewing Warren's PR #294, I came across the following code in `alignimages.py`:

https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L1216-L1218

https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L1250-L1253

It would appear that `ARGS.raw_test_data_path` is a string of several comma-separated "words"/paths. If that is the case, the logic in the loop is incorrect and `test_data_path` would be a Python list of **characters**! I do not believe this is what was intended here. Most likely you are missing `split(',')` in `for item in ARGS.raw_test_data_path:`, i.e., the loop should be `for item in ARGS.raw_test_data_path.split(','):`.

However, even then, why write it? Just do the following:
```python
test_data_path = list(ARGS.raw_test_data_path.split(','))
```

mcara:
The entire block in
https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L1222-L1231

can be simplified to:

```python
input_list = []
for item in ARGS.raw_input_list:
    try:
        with open(item, 'r') as infile:
            input_list.expand(map(str.strip, infile.readlines()))
    except FileNotFoundError:
        input_list.append(item)
```

mdlpstsci:
>             # ..........
Good suggestion though 
asnid = hdulist[0].header['ASN_ID'].strip().lower()

In truth, this routine needs a re-write

drizzlepac/alignimages.py
mdlpstsci: In reading the Pythonic way to handle the exception clause, I see what you are suggesting.  In our case I really do not know all the possible exceptions which can be raised.  It seems in these cases it is acceptable to modify the except block to log the full stack trace.  I have done this in the PyTest routines, but there may be conflicts with the functionality of PyTest so I am familiar with all the goodness I get.  In this case, I would change the exception clause to be 

    except Exception:
        exc_type, exc_value, exc_tb = sys.exc_info()
        traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
        filelist = []

drizzlepac/alignimages.py
mdlpstsci: If I just changed the `if` to `else`, then the totalInputList variable would not be updated with the last files acquired via retrieve_observation().  More importantly, we are reworking this routine as it needs to be robust to handle all the input configurations, both pipeline and standalone user.

mcara: You are right. I should have analyzed this better. There are two issues here. You could have dropped `filelist` completely in favor of `totalInputList` (IMO, there is no need for `filelist` but it is really optional). The second issue _then_ is that `if len(filelist) > 0:` on line 144 is not needed at all: Python knows how to "add" an empty list to a list.

mcara: For example,
https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L119-L145
can be replaced with:
```python
totalInputList = []
for input_item in input_list:
    try:
        totalInputList.append(
            aqutils.retrieve_observation(input_item, archive=archive, clobber=clobber)
        )
    except Exception:
        input_rootname = input_item[:input_item.find('_')]
        fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))
        if fitsfilenames:
            with fits.open(fitsfilenames[0]) as hdulist:
                asnid = hdulist[0].get('ASN_ID', 'none').lower().strip()
                if asnid not in retrieve_list:
                    try:
                        totalInputList.append(aqutils.retrieve_observation(asnid, **pars))
                        retrieve_list.append(asnid) # record asn's already retrieved
                    except Exception:
                        pass
```

mcara: By the way, why does this list need to be sorted at all? `fitsfilenames = sorted(glob.glob("{}_fl?.fits".format(input_rootname)))`?

drizzlepac/alignimages.py
mdlpstsci: The code in this routine is being entirely reworked.

tests/hla/test_extract.py
mdlpstsci: This entire Python file is deprecated and will be deleted.

drizzlepac/alignimages.py
mdlpstsci: The code in this routine is being entirely reworked.

tests/hla/test_extract.py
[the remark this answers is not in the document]
mdlpstsci: This python code is deprecated and will be deleted.

tests/hla/test_extract.py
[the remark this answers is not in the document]
mdlpstsci: This python code is deprecated and will be deleted.

tests/hla/test_extract.py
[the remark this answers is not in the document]
mdlpstsci: This python code is deprecated and will be deleted.

tests/hla/test_extract.py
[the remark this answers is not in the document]
mdlpstsci: This python code is deprecated and will be deleted.

tests/hla/test_align.py
```diff
@@ -0,0 +1,215 @@
...
+            * ACS dataset 10265_01: 4x F606W full-frame ACS/WFC images
+            * ACS dataset 12580_02: 5x F475W & 6x F814W ACS/WFC images
+            * WFC3 dataset 11663_12: 4x F160W full-frame WFC3/IR images
+            * WFC3 dataset 12219_01: 8x F160W full-frame WFC3/IR images, 9x F336W full-frame WFC3/UVIS images
+            * WFC3 dataset 12379_02: 4X F606W, 4x F502N full-frame WFC3/UVIS images
+
+        """
+        self.input_loc = 'base_tests'
+        self.curdir = os.getcwd()
+        totalRMS = 0.0
```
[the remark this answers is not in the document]
mdlpstsci: This code has been cleaned up and simplified.  I will note the totalRMS is initialized to -1 in the alignment code depending upon while the alignment failed, this value can still be -1.  This is why the valid range is 0.0 < totalRMS <= RMS_LIMIT.

tests/hla/test_align.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def test_align_47tuc(self):
+        """ Verify whether 47Tuc exposures can be aligned to an astrometric standard.
+
+        Characeteristics of this test:
+          * Input exposures include both ACS and WFC3 images of the same general field-of-view
+            of 47Tuc suitable for creating a combined mosaic using both instruments.
+        """
+        self.input_loc = 'mosaic_47tuc'
+        totalRMS = 0.0
```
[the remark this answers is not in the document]
mdlpstsci: Same response as above for line 178.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        """
+        write centroid statistics to compare differences btw two images
+        """
+        stats = ("minimum", "median", "maximum")
+        images = (None, None, image1, image2)
+        im_type = ("", "", "test", "reference")
+
+        diff = []
+        distances = self.centroid_distances(image1, image2, amp, size)
+        indexes = (0, len(distances)//2, len(distances)-1)
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        fd = open(fname, 'w')
+        fd.write("*** %s ***\n" % title)
+
+        if len(distances) == 0:
+            diff = [0.0, 0.0, 0.0]
+            fd.write("No matches!!\n")
+
+        elif len(distances) == 1:
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+
+        center[0] /= center[2]
+        center[1] /= center[2]
+        return center
+
+    def centroid_close(self, list_of_centroids, size, point):
+        """
+        Find if any centroid is close to a point
+        """
+        for i in range(len(list_of_centroids)-1, -1, -1):
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+            if (abs(list_of_centroids[i][0] - point[0]) < size / 2 and
+                abs(list_of_centroids[i][1] - point[1]) < size / 2):
+                return 1
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+                center[1] += x * image[y,x]
+                center[2] += image[y,x]
+
+        if center[2] == 0.0: return None
+
+        center[0] /= center[2]
+        center[1] /= center[2]
+        return center
+
+    def centroid_close(self, list_of_centroids, size, point):
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        xhi = min(xlo + size, image.shape[1])
+
+        center = [0.0, 0.0, 0.0]
+        for y in range(ylo, yhi):
+            for x in range(xlo, xhi):
+                center[0] += y * image[y,x]
+                center[1] += x * image[y,x]
+                center[2] += image[y,x]
+
+        if center[2] == 0.0: return None
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        """
+        Compute the centroid of a rectangular area
+        """
+        ylo = int(center[0]) - size // 2
+        yhi = min(ylo + size, image.shape[0])
+        xlo = int(center[1]) - size // 2
+        xhi = min(xlo + size, image.shape[1])
+
+        center = [0.0, 0.0, 0.0]
+        for y in range(ylo, yhi):
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        ymax = coords[0].max()
+        xmin = coords[1].min()
+        xmax = coords[1].max()
+        return (ymin, ymax, xmin, xmax)
+
+    def centroid(self, image, size, center):
+        """
+        Compute the centroid of a rectangular area
+        """
+        ylo = int(center[0]) - size // 2
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+from ci_watson.artifactory_helpers import get_bigdata_root
+from ci_watson.hst_helpers import raw_from_asn, ref_from_image, download_crds
+try:
+    from ci_watson.artifactory_helpers import check_url
+except ImportError:
+    from ci_watson.artifactory_helpers import _is_url as check_url
+
+from .base_classes import BaseTest
+
+__all__ = ['BaseHLATest', 'BaseHLAParTest', 'centroid_compare', 'BaseUnit']
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+            center1 = self.centroid(image1, size, pt)
+            if center1 is None: continue
+
+            disty = center2[0] - center1[0]
+            distx = center2[1] - center1[1]
+            dist = math.sqrt(disty * disty + distx * distx)
+            dflux = abs(center2[2] - center1[2])
+            distances.append([dist, dflux, center1, center2])
+
+        distances.sort(key=centroid_compare)
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+                output_spec = (output_file, output_truth)
+        else:
+            output_spec = {'files':(output_file, output_truth),
+                           'pars':output_pars}
+        outputs = [output_spec]
+        self.compare_outputs(outputs)
+
+
+def centroid_compare(centroid):
+    return centroid[1]
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+    params = {'test_modes':[dict(input="",
+                                 test_dir=None,
+                                 step_class=None,
+                                 step_pars=dict(),
+                                 output_truth="",
+                                 output_hdus=[])
+                            ]
+             }
+
+    def test_modes(self, input, test_dir, step_class, step_pars,
```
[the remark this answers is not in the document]
mdlpstsci: Deferred to Git Issue #295.

drizzlepac/alignimages.py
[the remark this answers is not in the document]
mdlpstsci: This routine is entirely being reworked.

drizzlepac/alignimages.py
mdlpstsci: Done although this routine is entirely being reworked.

drizzlepac/alignimages.py
mdlpstsci: Just a comment to the hungarianNotation or ALLCAPS...
This is the way the software engineer did this in other projects.  This code is being pulled out into its own file and being reworked.

drizzlepac/alignimages.py
mdlpstsci: This code is being pulled out into its own file and being reworked.

drizzlepac/alignimages.py
mdlpstsci: This routine is entirely being reworked.

tests/hla/test_align.py
```diff
@@ -0,0 +1,215 @@
...
+    def test_align_ngc188(self):
+        """ Verify whether NGC188 exposures can be aligned to an astrometric standard.
+
+        Characeteristics of this test:
+          * Input exposures include both ACS and WFC3 images of the same general field-of-view
+            of NGC188 suitable for creating a combined mosaic using both instruments.
+        """
+        self.input_repo = 'hst-hla-pipeline'
+        self.input_loc = 'mosaic_ngc188'
+        totalRMS = 0.0
```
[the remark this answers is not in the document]
mdlpstsci: This code has been cleaned up and simplified.

tests/hla/test_align.py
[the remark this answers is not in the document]
mdlpstsci: The code has been cleaned up.

mdlpstsci:
> Nested `if`s such as
> 
> [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L460-L461)
> 
> Lines 460 to 461 in [88c66cb](/spacetelescope/drizzlepac/commit/88c66cb34d3b98cee75995b2790314371abdc7b5)
>  if not tweakwcs_info_key.startswith("matched"): 
>      if tweakwcs_info_key.lower() == 'rms': 
> _generally speaking_ can be combined into a single statement such as:
> 
> ```python
> if not tweakwcs_info_key.startswith("matched") and tweakwcs_info_key.lower() == 'rms':
> ```
> 
> However, in this specific case, if `tweakwcs_info_key.lower() == 'rms'` is true, then `tweakwcs_info_key.startswith("matched")` cannot be true and, therefore, the first `if` in the original code can be simply dropped (i.e., leave only this):
> 
> ```python
> if tweakwcs_info_key.lower() == 'rms':
> ```
Deferred to Git Issue #295.

mdlpstsci:
> Code such as this:
> 
> [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L476-L483)
> 
> Lines 476 to 483 in [88c66cb](/spacetelescope/drizzlepac/commit/88c66cb34d3b98cee75995b2790314371abdc7b5)
>  if fitStatusDict[explicitDictKey]['valid'] == True: 
>      filteredTable[index]['status'] = 0 
>  else: 
>      filteredTable[index]['status'] = 1 
>  if fitStatusDict[explicitDictKey]['compromised'] == False: 
>      filteredTable['compromised'] = 0 
>  else: 
>      filteredTable['compromised'] = 1 
> 
> should not check if a boolean value is `True` or `False` - a boolean can be used directly in an `if`.
> 
> Moreover, this code can be simplified as follows:
> 
> ```python
> filteredTable[index]['status'] = int(not fitStatusDict[explicitDictKey]['valid'])
> filteredTable['compromised'] = int(fitStatusDict[explicitDictKey]['compromised'])
> ```
> 
> (assuming that you really need integer values instead of booleans).

Deferred to Git Issue #295.

mdlpstsci:
> `"{}".format(sciExtCtr)` in
> 
> [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L962-L964)
> 
> Lines 962 to 964 in [88c66cb](/spacetelescope/drizzlepac/commit/88c66cb34d3b98cee75995b2790314371abdc7b5)
>  sciExtDict = {} 
>  for sciExtCtr in range(1, num_sci_ext + 1): 
>      sciExtDict["{}".format(sciExtCtr)] = fileutil.findExtname(hdulist,'sci',extver=sciExtCtr) 
> 
> can be replaced with `str(sciExtCtr)`. In fact, the above three lines can be replaced with the following one-liner (albeit a long one):
> 
> ```python
> sciExtDict = {str(k): fileutil.findExtname(hdulist, 'sci', extver=k) for k in range(1, num_sci_ext + 1)}
> ```

Deferred to Git Issue #295.

mdlpstsci:
Rather than duplicate all the comments above made by Mihai, any items which have not been addressed thus far, will be "Deferred to Git Issue #295."

drizzlepac/alignimages.py
dulude: Issue addressed in PR #298

drizzlepac/alignimages.py
dulude: Issue addressed in PR #298

mcara:
I do not understand the purpose of the [astrometry_evaluation_notes.txt](https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/tests/hla/astrometry_evaluation_notes.txt) file. How will users know they need to go to `GitHub` and search for text files buried somewhere within the `drizzlepac` in order to find appropriate documentation.

If this is a note for some very narrow audience, maybe this should go into a JIRA ticket or confluence page.

This also brings the issue of documentation. I have not seen any updates in the `doc` directory related to the HLA work. Again, is this for a very narrow audience?

@stsci-hack

mcara:
@mdlpstsci @stsci-hack I am not sure what is going on here. Specifically, on numerous comments the answer is "Done". What does this mean? I do not see any new commits that actually address the comments in the reviews.

On other occasion - _"This entire Python file is deprecated and will be deleted."_ - see https://github.com/spacetelescope/drizzlepac/pull/283#discussion_r269720094 the answer is that a file that is a new file added by this PR is deprecated and being deleted. I do not want to merge a file that will be deleted.

Please add more commits to this PR that address the most critical issues.

tests/hla/base_test.py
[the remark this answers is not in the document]
mcara: Why defer this?

mdlpstsci:
> I do not understand the purpose of the [astrometry_evaluation_notes.txt](https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/tests/hla/astrometry_evaluation_notes.txt) file. How will users know they need to go to `GitHub` and search for text files buried somewhere within the `drizzlepac` in order to find appropriate documentation.
> 
> If this is a note for some very narrow audience, maybe this should go into a JIRA ticket or confluence page.
> 
> This also brings the issue of documentation. I have not seen any updates in the `doc` directory related to the HLA work. Again, is this for a very narrow audience?
> 
> @stsci-hack

I copied all the files from the test directory of our previous repository.  It is simply a case that I did not delete or relocate the astrometry_evaluation_notes.txt file to an appropriate place.  On the same topic, the test_extract.py file is the one that is deprecated from our old repository.  Sorry that I confused you as I am so accustomed to seeing this file, I forgot that it has not even been committed to this new repository.  No worries that it will be committed.  Obviously, it does not need to be committed just to be deleted.  The documentation for this work is currently a lower priority JIRA ticket than the issues we are currently working for the HLA project.

mdlpstsci:
> @mdlpstsci @stsci-hack I am not sure what is going on here. Specifically, on numerous comments the answer is "Done". What does this mean? I do not see any new commits that actually address the comments in the reviews.
> 
> On other occasion - _"This entire Python file is deprecated and will be deleted."_ - see [#283 (comment)](https://github.com/spacetelescope/drizzlepac/pull/283#discussion_r269720094) the answer is that a file that is a new file added by this PR is deprecated and being deleted. I do not want to merge a file that will be deleted.
> 
> Please add more commits to this PR that address the most critical issues.
Sorry to have confused you.  The "Done" means that I have made the change in my working copy of the source.  When I am ready to commit the updated code for further review, I will "resolve" the conversation.  I am nearly ready for commit.  I hope you are understanding in that the work flow that I have been using seems to be different than your expectations.  Perhaps we should just have a face-to-face (charmingly old fashioned) to clarify expectations.

stsci-hack:
The file **astrometry_evaluation_notes.txt** was intended to serve as an informal way of creating notes for developers working on creating the original **hlapipeline** package.  This file can and should be excluded from ever being a merged part of the **drizzlepac** package in its current form.  Instead, those notes should be used (at a later date, through another git/jira issue and PR) to help develop documentation that describes the new code.   This file will (as I understand it, right @mdlpstsci ) be removed from this PR, so that once merged, it will never show up in master in its current form.

mdlpstsci:
I need to merge or rebase with master.

tests/hla/test_randomlist.py
```diff
@@ -168,8 +179,10 @@ def align_randomFields(self, randomTable):
                 continue
 
         # Perform some clean up
-        if os.path.exists('ref_cat.ecsv'): os.remove('ref_cat.ecsv')
-        if os.path.exists('refcatalog.cat'): os.remove('refcatalog.cat')
+        if os.path.exists('ref_cat.ecsv'):
```
mcara: I recommend using `os.path.isfile()` instead of the generic `os.path.exists`. It is more specific.

drizzlepac/alignimages.py
```diff
@@ -79,80 +79,98 @@ def check_and_get_data(input_list,**pars):
...
+                            log.error('File {} cannot be retrieved from MAST.'.format(input_item))
+                            return(empty_list)
+                except Exception:
+                    exc_type, exc_value, exc_tb = sys.exc_info()
+                    traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
+
+    # Only the retrieve_list files via astroquery have been put into the total_input_list thus far.
+    # Now check candidate_list to detect or acquire the requested files from MAST via
+    # astroquery.
+    if candidate_list:
```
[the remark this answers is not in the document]
mdlpstsci: OK, I changed this, but  IMHO this seems to be stylistic only.

drizzlepac/alignimages.py
```diff
@@ -79,80 +79,98 @@ def check_and_get_data(input_list,**pars):
...
+            suffix = lc_input_item[indx+1:indx+4]
+            print('file: ', lc_input_item)
+            # For an association, need to open the table and read the image names as this could
+            # be a custom association.  The assumption is this file is on local disk when specified
+            # in this manner (vs just the ipppssoot of the association).
+            # This "if" block just collects the wanted full file names.
+            if suffix == 'asn':
+                try:
+                    asntab = Table.read(input_item, format='fits')
+                except ValueError:
```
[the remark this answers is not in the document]
mdlpstsci: Of course, this was my mistake when trying various commands in a Python session.  Fixed.

drizzlepac/alignimages.py
```diff
@@ -79,80 +79,98 @@ def check_and_get_data(input_list,**pars):
...
+                for row in asntab:
+                    if row['MEMTYPE'].startswith('PROD'):
+                        continue
+                    memname = row['MEMNAME'].lower().strip()
+                    # Need to check if the MEMNAME is a full filename or an ipppssoot
+                    if memname.find('_') != -1:
+                        candidate_list.append(memname)
+                    else:
+                        candidate_list.append(memname + '_flc.fits')
+            elif any ([suffix == 'flc', suffix == 'flt']):
```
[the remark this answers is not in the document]
mdlpstsci: Agreed and fixed.

drizzlepac/alignimages.py
```diff
@@ -79,80 +79,98 @@ def check_and_get_data(input_list,**pars):
...
+                if lc_input_item not in candidate_list:
+                    candidate_list.append(lc_input_item)
+            else:
+                log.error('Inappropriate file suffix: {}.  Looking for "asn.fits", "flc.fits", or "flt.fits".'.format(suffix))
+                return(empty_list)
+
+        # Input is an ipppssoot (association or singleton), nine characters by definition.
+        # This "else" block actually downloads the data specified as ipppssoot.
+        else:
+            if len(input_item) == 9:
```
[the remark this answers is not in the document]
mdlpstsci: Yes and fixed.

tests/hla/test_align.py
```diff
@@ -45,75 +45,78 @@ class TestAlignMosaic(BaseHLATest):
...
         input_filenames = ['iaal01hxq_flc.fits', 'iaala3btq_flc.fits',
                             'iaal01hyq_flc.fits', 'iaala3bsq_flc.fits',
                             'j8boa1m8q_flc.fits', 'j8boa1m4q_flc.fits',
                             'j8boa1maq_flc.fits', 'j8boa1m6q_flc.fits']
 
-        data_path = ['hst-hla-pipeline','dev','mosaic_ngc188']
+        # Since these are full file names (*_flc.fits) which cannot be obtained via astroquery from
+        # MAST, get the data now using ci_watson.
+        for input_file in input_filenames:
+            abs_path = get_bigdata('hst-hla-pipeline','dev','mosaic_ngc188',input_file)
```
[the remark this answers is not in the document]
mdlpstsci: Indeed, I do not need abs_path, but the get_bigdata() returns a variable.  I have just removed the assignment to a local variable and learned about "_" as a return type as an added bonus.

tests/hla/test_align.py
```diff
@@ -173,17 +177,19 @@ def test_align_single_visits(self,input_filenames):
...
         """
-        self.input_loc = 'base_tests'
-        self.curdir = os.getcwd()
         totalRMS = 0.0
-        data_path = ['hst-hla-pipeline','dev','base_tests']
+
+        # Since these are full file names (*_flc.fits) which cannot be obtained via astroquery from
+        # MAST, get the data now using ci_watson.
+        for input_file in input_filenames:
+            abs_path = get_bigdata('hst-hla-pipeline','dev','base_tests',input_file)
```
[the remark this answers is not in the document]
mdlpstsci: Same as above - removed assignment.

tests/hla/test_randomlist.py
mdlpstsci: OK fixed.

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        hdu.header['CD1_1'] = image_wcs.wcs.cd[0][0]
+        hdu.header['CD1_2'] = image_wcs.wcs.cd[0][1]
+        hdu.header['CD2_1'] = image_wcs.wcs.cd[1][0]
+        hdu.header['CD2_2'] = image_wcs.wcs.cd[1][1]
+        hdu.header['CRVAL1'] = image_wcs.wcs.crval[0]
+        hdu.header['CRVAL2'] = image_wcs.wcs.crval[1]
+        hdu.header['CRPIX1'] = image_wcs.wcs.crpix[0]
+        hdu.header['CRPIX2'] = image_wcs.wcs.crpix[1]
+        hdu.header['CTYPE1'] = image_wcs.wcs.ctype[0]
+        hdu.header['CTYPE2'] = image_wcs.wcs.ctype[1]
```
mcara: @nden Couldn't most of the code here be simplified using `WCS.to_header()`?
```python
from astropy.wcs import WCSHDO_none  # at the top of the file
hdu.header.update(image_wcs.wcs.to_header(WCSHDO_none))
hdu.header['ORIENTAT'] = image_wcs.orientat
hdu.header['VAFACTOR'] = 1.0
```

mcara: Eh,... ignore my previous comment: `to_header()` gives `PC` coefficients instead of `CD` (they are the same numerically for HST but require extra handling, i.e., renaming).

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        return image
+
+    def read_wcs(self, filename):
+        """
+        Read the wcs of a fits file
+        """
+        hdu = fits.open(filename)
+
+        wcs = stwcs.wcsutil.HSTWCS(hdu, 1)
+        hdu.close()
```
mcara: Worse:
```python
with fits.open(filename) as hdu:
    wcs = stwcs.wcsutil.HSTWCS(hdu, 1)
```
Best (option): `HSTWCS` can open files itself:
```python
wcs = stwcs.wcsutil.HSTWCS(filename, 1)
```
Given this, is `read_wcs()` even necessary?

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+
+    def read_image(self, filename):
+        """
+        Read the image from a fits file
+        """
+        hdu = fits.open(filename)
+
+        image = hdu[1].data
+        hdu.close()
+        return image
```
mcara: Why not use `astropy.fits` "convenience" function here?
```python
def read_image(self, filename):
    return fits.getdata(filename, 1)
```

tests/hla/base_test.py
```diff
@@ -0,0 +1,405 @@
...
+        """
+        Print the wcs header cards
+        """
+        print("=== %s ===" % title)
+        print(wcs.to_header_string())
+
+
+    def read_image(self, filename):
+        """
+        Read the image from a fits file
```
mcara: "Read the image from the **first extension** of a fits file"

tests/hla/base_test.py
mdlpstsci: OK - ignored.

tests/hla/base_test.py
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
mdlpstsci: Deferred to Git Issue #295.

tests/hla/base_test.py
mdlpstsci: Deferred to Git Issue #295.

stsci-hack:
All functional issues have been addressed.  Additional issues were identified by the reviewers and those issues have been documented for resolution through a separate git issue (primarily, #295 ).  However, this code is needed in order to deliver required functionality for HSTDP 2019.2.

tests/hla/base_test.py
stsci-hack: Will be addressed in #309 .

tests/hla/base_test.py
stsci-hack: Addressed in #309

tests/hla/base_test.py
stsci-hack: Implemented in #309

stsci-hack:
> > `"{}".format(sciExtCtr)` in
> > [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L962-L964)
> > Lines 962 to 964 in [88c66cb](/spacetelescope/drizzlepac/commit/88c66cb34d3b98cee75995b2790314371abdc7b5)
> > sciExtDict = {}
> > for sciExtCtr in range(1, num_sci_ext + 1):
> > sciExtDict["{}".format(sciExtCtr)] = fileutil.findExtname(hdulist,'sci',extver=sciExtCtr)
> > can be replaced with `str(sciExtCtr)`. In fact, the above three lines can be replaced with the following one-liner (albeit a long one):
> > ```python
> > sciExtDict = {str(k): fileutil.findExtname(hdulist, 'sci', extver=k) for k in range(1, num_sci_ext + 1)}
> > ```
> 
> Deferred to Git Issue #295.

The suggested change is implemented in #309.

stsci-hack:
> > Code such as this:
> > [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L476-L483)
> > Lines 476 to 483 in [88c66cb](/spacetelescope/drizzlepac/commit/88c66cb34d3b98cee75995b2790314371abdc7b5)
> > if fitStatusDict[explicitDictKey]['valid'] == True:
> > filteredTable[index]['status'] = 0
> > else:
> > filteredTable[index]['status'] = 1
> > if fitStatusDict[explicitDictKey]['compromised'] == False:
> > filteredTable['compromised'] = 0
> > else:
> > filteredTable['compromised'] = 1
> > should not check if a boolean value is `True` or `False` - a boolean can be used directly in an `if`.
> > Moreover, this code can be simplified as follows:
> > ```python
> > filteredTable[index]['status'] = int(not fitStatusDict[explicitDictKey]['valid'])
> > filteredTable['compromised'] = int(fitStatusDict[explicitDictKey]['compromised'])
> > ```
> > 
> > 
> > (assuming that you really need integer values instead of booleans).
> 
> Deferred to Git Issue #295.

Implemented in PR #309.

stsci-hack:
> > Nested `if`s such as
> > [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/88c66cb34d3b98cee75995b2790314371abdc7b5/drizzlepac/alignimages.py#L460-L461)
> > Lines 460 to 461 in [88c66cb](/spacetelescope/drizzlepac/commit/88c66cb34d3b98cee75995b2790314371abdc7b5)
> > if not tweakwcs_info_key.startswith("matched"):
> > if tweakwcs_info_key.lower() == 'rms':
> > _generally speaking_ can be combined into a single statement such as:
> > ```python
> > if not tweakwcs_info_key.startswith("matched") and tweakwcs_info_key.lower() == 'rms':
> > ```
> > 
> > 
> > However, in this specific case, if `tweakwcs_info_key.lower() == 'rms'` is true, then `tweakwcs_info_key.startswith("matched")` cannot be true and, therefore, the first `if` in the original code can be simply dropped (i.e., leave only this):
> > ```python
> > if tweakwcs_info_key.lower() == 'rms':
> > ```
> 
> Deferred to Git Issue #295.

Implemented in PR #309

stsci-hack:
> The entire block in
> [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L1222-L1231)
> 
> Lines 1222 to 1231 in [629028e](/spacetelescope/drizzlepac/commit/629028ed18dd52649c253236b2efb8f6924e93da)
> 
>  # Build list of input images 
>  input_list = [] 
>  for item in ARGS.raw_input_list: 
>      if os.path.exists(item): 
>          with open(item, 'r') as infile: 
>              fileLines = infile.readlines() 
>          for fileLine in fileLines: 
>              input_list.append(fileLine.strip()) 
>      else: 
>          input_list.append(item) 
> can be simplified to by getting rid of the innermost loop such as:
> 
> ```python
> input_list = []
> for item in ARGS.raw_input_list:
>     if os.path.isfile(item):  # or, try:
>         with open(item, 'r') as infile:
>             input_list.expand(map(str.strip, infile.readlines()))
>     else:  # or, except FileNotFoundError:
>         input_list.append(item)
> ```

Implemented in PR #309.

stsci-hack:
> While reviewing Warren's PR #294, I came across the following code in `alignimages.py`:
> 
> [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L1216-L1218)
> 
> Lines 1216 to 1218 in [629028e](/spacetelescope/drizzlepac/commit/629028ed18dd52649c253236b2efb8f6924e93da)
> 
>  PARSER.add_argument( '-t', '--raw_test_data_path', required=False,default=None,help='A comma separated list of strings ' 
>                  'defining the path on Artifactory or local disk where the *_fl[t|c].fits files are stored.  The root of ' 
>                  'the path is defined by the environment variables TEST_BIGDATA. ["hst-hla-pipeline","dev","base_tests"].') 
> [drizzlepac/drizzlepac/alignimages.py](https://github.com/spacetelescope/drizzlepac/blob/629028ed18dd52649c253236b2efb8f6924e93da/drizzlepac/alignimages.py#L1250-L1253)
> 
> Lines 1250 to 1253 in [629028e](/spacetelescope/drizzlepac/commit/629028ed18dd52649c253236b2efb8f6924e93da)
> 
>  # Build list for test data path 
>  test_data_path = [] 
>  for item in ARGS.raw_test_data_path: 
>      test_data_path.append(item) 
> It would appear that `ARGS.raw_test_data_path` is a string of several comma-separated "words"/paths. If that is the case, the logic in the loop is incorrect and `test_data_path` would be a Python list of **characters**! I do not believe this is what was intended here. Most likely you are missing `split(',')` in `for item in ARGS.raw_test_data_path:`, i.e., the loop should be `for item in ARGS.raw_test_data_path.split(','):`.
> 
> However, even then, why write it? Just do the following:
> 
> ```python
> test_data_path = list(ARGS.raw_test_data_path.split(','))
> ```

This command-line parameter was removed entirely.

PR #298, opened 2019-03-28
Streamline command-line interface section of alignimages.py

dulude:
This PR addresses two items raised in #283 : Namely [this one](https://github.com/spacetelescope/drizzlepac/pull/283#discussion_r268901463), and [this one](https://github.com/spacetelescope/drizzlepac/pull/283#discussion_r268903028).
- rolled string to boolean conversion into add_argument() statements that expect "True"/"False" inputs. This eliminates the need for a second line to a call a string to boolean conversion subroutine for each argument expecting a "True"/"False" input.
- converted all upper-case and camelcase variable names to lowercase in the command-line interface section of alignimages.py
-  made some minor logic tweaks to the code that builds input_list in the command-line interface section of alignimages.py

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -129,24 +129,6 @@ def check_and_get_data(input_list,**pars):
 
 
 # ----------------------------------------------------------------------------------------------------------------------
-
-def str2bool(v):
-    """Converts string 'True' or 'False' value to Boolean True or Boolean False.
-
-    :param invalue: string
-        input true/false value
-
-    :return: Boolean
-        converted True/False value
-    """
-    if v.lower() in ('yes', 'true', 't', 'y', '1'):
-        return True
-    elif v.lower() in ('no', 'false', 'f', 'n', '0'):
-        return False
-    else:
-        raise argparse.ArgumentTypeError('Boolean value expected.')
-
-# ----------------------------------------------------------------------------------------------------------------------
 def perform_align(input_list, **kwargs):
     """Main calling function.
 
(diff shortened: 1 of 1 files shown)


mcara:
The idea in https://github.com/spacetelescope/drizzlepac/pull/283#discussion_r268903028 was that boolean arguments are not plain arguments but rather some sort of "switches" that either enable or disable a specific functionality/behavior/etc.

When one wants to do a recursive copy of a directory, one does the following:
```
cp -r origin dest
```
instead of
```
cp -r True origin dest
```
I have not really encountered second usage style before and I do not see any compelling justification to introduce such a behavior now.

Not every answer on StackOverflow is a good answer and the solution in the accepted answer https://stackoverflow.com/a/43357954 (while answering the OP question) is not a good solution in general. The [most up-voted answer](https://stackoverflow.com/a/15008806) which I cited, on the other hand, shows how to implement the "canonical" solution.

Another example of what I was talking about: https://stackoverflow.com/q/8259001 Take a look at the [second answer](https://stackoverflow.com/a/36710639) to see how this works. In that example `args.w` is just a boolean and there is no need to convert anything to boolean. It is even better for users: they get a familiar interface behavior and if the do not want to "Retain copies of the copies of the downloaded files in the astroquery created sub-directories?" they do not even need to write anything and if they do - they will need to simply add `-a` as an argument - not `-a True`.

dulude:
Resolved issues raised by @mcara in commit [fd8e02c](https://github.com/spacetelescope/drizzlepac/pull/298/commits/fd8e02cf318e58cad021d160a027476335691a49).

PR #299, opened 2019-03-28
Fix envvar bug, and added force parameter

stsci-hack:
Fix a bug where the code crashed when trying to find the environment variable ASTROMETRY_STEP_CONTROL when it had not been defined by the user yet.  Also, added 'force' parameter in order to allow the user to refresh local copy of the data which appears to be necessary after applying the subsequent headerlets.

### DIFF
--- a/drizzlepac/hlautils/testutils.py
+++ b/drizzlepac/hlautils/testutils.py
@@ -112,7 +112,7 @@ def compare_wcs_alignment(dataset, force=False):
         # Regardless of what happens, always reset the environment variable
         # if it was modified in the first place.
         # Restore user environment to original state
-        if control:
+        if control is None:  # Need to be explicit here since T/F are actually valid
             os.environ['ASTROMETRY_STEP_CONTROL'] = control
         else:
             del os.environ['ASTROMETRY_STEP_CONTROL']
--- a/drizzlepac/hlautils/testutils.py
+++ b/drizzlepac/hlautils/testutils.py
@@ -113,9 +113,9 @@ def compare_wcs_alignment(dataset, force=False):
         # if it was modified in the first place.
         # Restore user environment to original state
         if control is None:  # Need to be explicit here since T/F are actually valid
-            os.environ['ASTROMETRY_STEP_CONTROL'] = control
-        else:
             del os.environ['ASTROMETRY_STEP_CONTROL']
+        else:
+            os.environ['ASTROMETRY_STEP_CONTROL'] = control
 
     return alignment
 
--- a/drizzlepac/hlautils/testutils.py
+++ b/drizzlepac/hlautils/testutils.py
@@ -28,7 +28,7 @@ def compare_wcs_alignment(dataset, force=False):
 
         force : bool
             Specify whether or not to overwrite dataset files found locally
-            with fresh copies retrieved from MAST.  Default: False
+            with fresh copies retrieved from MAST.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


drizzlepac/hlautils/testutils.py
```diff
@@ -102,7 +110,8 @@ def compare_wcs_alignment(dataset):
         alignment[wcs] = extract_results(results)
 
     # Restore user environment to original state
-    os.environ['ASTROMETRY_STEP_CONTROL'] = control
+    if control:
+        os.environ['ASTROMETRY_STEP_CONTROL'] = control
```
mcara: In my opinion, there is a serious issue with this change/restoring of the environment: if a crash occurs, the environment will not be restored to the previous state. It think this should be either addressed somehow or documented/explained as to why this should be OK.

drizzlepac/hlautils/testutils.py
```diff
@@ -50,12 +54,16 @@ def compare_wcs_alignment(dataset):
     """
     # Setup
     #   Insure that database will be queried for new WCS solutions
-    control = os.environ['ASTROMETRY_STEP_CONTROL']
+    if 'ASTROMETRY_STEP_CONTROL' in os.environ:
+        control = os.environ['ASTROMETRY_STEP_CONTROL']
+    else:
+        control = None
```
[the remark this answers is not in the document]
stsci-hack: I knew there was a simple way to handle this, so thanks for reminding me.  Will get this fixed...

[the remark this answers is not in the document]
mcara: Original code is OK (i.e., nothing to "fix"). Just 3 fewer lines to look at.

drizzlepac/hlautils/testutils.py
stsci-hack: This was fixed by wrapping the code in a try...except...finally block which still preserves all traceback information and insures that no matter what happens, if the environment was changed it will be restored. :-)

drizzlepac/hlautils/testutils.py
```diff
@@ -50,12 +54,16 @@ def compare_wcs_alignment(dataset):
...
     os.environ['ASTROMETRY_STEP_CONTROL'] = 'ON'
 
     # Step 1:
     #   Determine alignment for pipeline-defined WCS
-    results = alignimages.perform_align([dataset], clobber=False, debug=True)
+    clobber = force
```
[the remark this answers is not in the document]
stsci-hack: Fixed.  This remained in the code as I originally had an input parameter with the exact opposite meaning and tried to interpret it separately.  Anyway, fixed...

drizzlepac/hlautils/testutils.py
```diff
@@ -26,6 +26,10 @@ def compare_wcs_alignment(dataset):
         dataset : str
             Rootname of either a single (un-associated) exposure or an ASN
 
+        force : bool
+            Specify whether or not to overwrite dataset files found locally
+            with fresh copies retrieved from MAST.  Default: False
```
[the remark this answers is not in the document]
stsci-hack: Old habits die hard where I have always tried to be as explicit in the docs as possible (a convention originally needed under the pre-sphinx doc standard).

drizzlepac/hlautils/testutils.py
```diff
@@ -53,65 +53,70 @@ def compare_wcs_alignment(dataset, force=False):
...
+                if wcs in wnames:
+                    hdrname = hnames[wnames.index(wcs)]
+                    log.info("Applying WCS {} to {}".format(hdrname, img))
+                    headerlet.restore_from_headerlet(img, hdrname=hdrname,
+                                                     archive=False, force=True)
+
+            results = alignimages.perform_align([dataset], clobber=False, debug=True)
+            alignment[wcs] = extract_results(results)
+
+    except Exception:
```
mcara: One more thing... I think `except` block is not necessary at all - you can safely remove it. Also, it really should be either `except:` or `except Exception as e:` (in order to do some additional processing on `e` before re-rising it).

drizzlepac/hlautils/testutils.py
mcara: except the except block is not necessary :-)

drizzlepac/hlautils/testutils.py
mcara: That is, you can have a construction like this:
```python
try:
     # do something dangerous
finally:
    # do something that always need to be done regardless of the dangerous stuff

Comment (stsci-hack) [reply drizzlepac/hlautils/testutils.py]: I was not quite sure what would happen in the case of exceptions, so I erred on the side of being very explicit (as per Zen of Python).  It turns out, though, that this logic still misses a key behavior which will be fixed in the next commit: removing the variable from os.environ if it wasn't there in the first place.

File: drizzlepac/hlautils/testutils.py [hunk 6]
```diff
@@ -49,60 +53,70 @@ def compare_wcs_alignment(dataset):
...
+        raise   # re-raise with all traceback information
+
+    finally:
+        # Regardless of what happens, always reset the environment variable
+        # if it was modified in the first place.
+        # Restore user environment to original state
+        if control:
```
mcara: Also, here specifically, it would be better/safer (IMO) to have a check such as `if control is None:` because an empty string could evaluate to `False`. That is, what if `environ['ASTROMETRY_STEP_CONTROL'] = ''`. Then currently this env will not be restored to its original state.

drizzlepac/hlautils/testutils.py
mcara: Yes! Great! `if control is None: del env['...'] else env['...'] = control` I think this will do.

stsci-hack: Removing 'except Exception' block worked just as intended.

drizzlepac/hlautils/testutils.py
stsci-hack: Done.

drizzlepac/hlautils/testutils.py
```diff
@@ -49,60 +53,69 @@ def compare_wcs_alignment(dataset):
...
+
+            results = alignimages.perform_align([dataset], clobber=False, debug=True)
+            alignment[wcs] = extract_results(results)
+
+    finally:
+        # Regardless of what happens, always reset the environment variable
+        # if it was modified in the first place.
+        # Restore user environment to original state
+        if control is None:  # Need to be explicit here since T/F are actually valid
+            os.environ['ASTROMETRY_STEP_CONTROL'] = control
```
mcara: I think the branches are flipped over.

stsci-hack: I will get this right eventually, with your help... fixed.

PR #300, opened 2019-03-29
Interpret pipeline poller inputs

stsci-hack:
This initial version of the pipeline poller output creates the 'obs_info_dict' dictionary as defined for input into the product naming code.  The doc string of the new primary interface function, interpret_input, includes a sample of the poller input and the corresponding dict definition that will be created by the code.  

The code was written at this point to simply perform the correct parsing of the input in a manner that could be debugged during development.  No attempt was made to make this particularly efficient and NO changes will be considered that only change the syntax of this code without fixing a bug that will prevent this code from working as intended in the pipeline.  Any syntax changes or efficiency improvements will need to be considered in separate issues and handled through separate PRs later. 

This code has already been tested on actual poller output (both original output, and 'scrambled' output) in an attempt to validate that the logic will work regardless of the order of input rows; a test that this code did pass.


drizzlepac/generate_final_product_filenames.py
```diff
@@ -1,9 +1,188 @@
...
+    visit_table.add_column(Column([instr]*len(visit_table)),name='instrument')
+    # parse Table into a tree-like dict
+    visit_tree = build_visit_tree(visit_table, exposure_limit=exposure_limit)
+    # Now create final dict
+    visit_dict = parse_visit_tree(visit_tree)
+    
+    return visit_dict
+
+# Translate the database query on a visit into actionable lists of filenames
+def build_visit_tree(visit_table, exposure_limit=2.0):
```
mcara: I am not sure about this whole function. I have a feeling it can be written in a different way. If anything I will comment on the main "thread".

mcara:
Also, I understand that the old code is the `drizzlepac` is not PEP8-compliant but at least new code could be made PEP8. This could be done either in a single PR or actually have each new PR update the files that it touches.

drizzlepac/generate_final_product_filenames.py
```diff
@@ -1,9 +1,188 @@
...
+    """Convert visit table into a tree listing all products to be created."""
+
+    # Each product will consist of the appropriate string as the key and
+    # a dict of 'info' and 'files' information
+    base_entry = {'info':"", 'files':[]}
```
mcara: Not used anywhere.

drizzlepac/generate_final_product_filenames.py
```diff
@@ -1,9 +1,188 @@
...
 
 """
+def interpret_input(results, exposure_limit=2.0):
+    """
+    Interpret the database query for a given visit to prepare the returned
+    values for use in generating the names of all the expected output products.
+    
+    Input will have format of:
+        ia1s70jtq_flt.fits,11150,A1S,70,70,149.232269,F110W,IR,/ifs/archive/ops/hst/public/ia1s/ia1s70jtq/ia1s70jtq_flt.fits
+        ia1s70iwq_flt.fits,11150,A1S,70,70,0.91161000000000003,F160W,IR,/ifs/archive/ops/hst/public/ia1s/ia1s70iwq/ia1s70iwq_flt.fits
```
mcara: I have to ask sphinx expert @sosey to give her feedback on how to best represent these dictionaries in the doctstring. Maybe use `.. code-block::`?

mcara:
@stsci-hack It was interesting to play with `build_visit_tree()` to see if I could come up with different approach(es). I think readability is important and I have a feeling that the shortest version is less readable. So, use at your own risk.

#### Option 1:
```python
def build_visit_tree(visit_table, exposure_limit=2.0): 
    vt = {} 
    for row in visit_table: 
        d = vt 
        for kw in ('instrument', 'detector', 'filters'): 
            if row[kw] not in d: 
                d[row[kw]] = {} 
            d = d[row[kw]]
        rwi = create_row_info(row)
        x = row['exptime'] 
        for e in d.keys() + [x]: 
            if e / exposure_limit <= x <= e * exposure_limit: 
                if e in d:
                    d[e].append(rwi)
                else:
                    d[x] = [rwi]
                break
    return vt 
```

#### Option 2:
```python
from collections import defaultdict
from itertools import dropwhile
from functools import reduce
def build_visit_tree2(visit_table, exposure_limit=2.0):
    vt = {}
    tmp = {}
    for row in visit_table:
        kwds = tuple(row[k] for k in ('instrument', 'detector', 'filters'))
        rwi = create_row_info(row)
        x = row['exptime']
        if kwds not in tmp:
            tmp[kwds] = defaultdict(list, {x: [rwi]})
        else:
            tmp[kwds][(list(dropwhile(
                lambda e: e / exposure_limit > x or
                x > e * exposure_limit, tmp[kwds])) + [x])[0]].append(rwi)
    for k, v in tmp.items():
        reduce(lambda s, kw: s.setdefault(kw, {}), k[:-1], vt)[k[-1]] = dict(v)
    return vt
```

drizzlepac/generate_final_product_filenames.py
pllim: Yes, code block directive would be your best bet.

Also, if this instruction is meant for user, not developer, it is better to put this in the user-facing RST doc, not buried inside API doc.

stsci-hack:
This work should not be merged until v3.0.2 has been publicly released, since it is only intended to be used in generating single-visit mosaics.

mdlpstsci:
I am starting to examine this code and will keep in mind the error that you, Mike, have discovered.

drizzlepac/generate_final_product_filenames.py
sosey: huh, I never got notified for this...but I agree with pllim

stsci-hack: The developer should always have some idea of what kind of input this code is expecting, especially when that input is NOT simple.  This allows the developer to understand what the code is trying to do, or how changes may interact with the input to avoid causing problems due to assumptions about the inputs.  Therefore, this would be a good candidate for documentation that both the developer and end user may need, so it should go in both places with a more complete explanation in the user-facing RST docs.  The use of code-block is a great suggestion in this case as well!

sosey: This is true, however I think the code, documentation, and future users and developers would benefit if all the classes, methods, and functions in this package were updated to include the common code documentation format. Right now it appears inconsistent and there are mostly multi-line docstrings, consider updating everything to included explicit sections for attributes, outputs, notes, examples, etc., that are reST compatible. These can be automatically compiled with sphinx as part of the general documentation that is useful for everyone.

stsci-hack:
This new code, along with all changes requested by reviewers, will be merged in through PR #322 instead.

PR #301, opened 2019-04-01
Add partial clean

stsci-hack:
A new parameter was added that allows the user to control what intermediate files are not deleted when 'clean=True'.  The parameter was implemented independently of the configobj parameters since this capability really is only intended to support scripting use of Astrodrizzle which may require working with custom sets of the intermediate files (like HLA processing).  

This is intended to support creation of single visit mosaic products (v3.1.0), and NOT for initial use with astrometric updates to HST data.

### DIFF
--- a/drizzlepac/astrodrizzle.py
+++ b/drizzlepac/astrodrizzle.py
@@ -57,7 +57,7 @@ log = logutil.create_logger(__name__, level=logutil.logging.NOTSET)
 
 
 def AstroDrizzle(input=None, mdriztab=False, editpars=False, configobj=None,
-                 wcsmap=None, keep_files=None, **input_dict):
+                 wcsmap=None, keep_files=[], **input_dict):
     """ AstroDrizzle command-line interface """
     # Support input of filenames from command-line without a parameter name
     # then copy this into input_dict for merging with TEAL ConfigObj
@@ -121,7 +121,7 @@ def AstroDrizzle(input=None, mdriztab=False, editpars=False, configobj=None,
 ##  Interfaces used by TEAL ##
 ##############################
 @util.with_logging
-def run(configobj, wcsmap=None, keep_files=None):
+def run(configobj, wcsmap=None, keep_files=[]):
     """
     Initial example by Nadia ran MD with configobj EPAR using:
     It can be run in one of two ways:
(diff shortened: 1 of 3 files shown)


drizzlepac/astrodrizzle.py
```diff
@@ -159,6 +159,23 @@ def run(configobj, wcsmap=None):
...
+    # validate input values for keep_file parameter
+    clean_files = ['blotImage','crmaskImage','finalMask',
+                    'staticMask','singleDrizMask','outSky',
+                    'outSContext','outSWeight','outSingle',
+                    'outMedian','dqmask','tmpmask',
+                    'skyMatchMask']
+    invalid_entries = []
+    for kfile in keep_files:
+        if kfile not in clean_files:
+            invalid_entries.append(kfile)
```
mcara: This is an ideal application for sets:
```python
invalid_entries  = list(set(keep_files).difference(clean_files))
```

drizzlepac/imageObject.py
```diff
@@ -128,15 +128,22 @@ def close(self):
...
                         'skyMatchMask']
-
+        if keep_files:
+            for kfile in keep_files:
+                if kfile in clean_files:
+                    clean_files.remove(kfile)
+                else:
+                    log.info("Did not find intermediate product type {} to keep.".format(kfile))
+                    log.info("Valid choices are: {}".format(clean_files))
+
```
mcara: Another good application for sets. Also, for every "missing" (from `clean_files`) intermediate product that should be kept, there will be a log info listing all `clean_files`. Maybe it would be better to do this only ones?

stsci-hack: This logic was moved from here into 'astrodrizzle.py' so that the user can be made aware of any problems with the specification of `keep_files` before the code crashes after running for some indeterminate (long?) period of time.  The change also results in only 1 message being logged for the user, not one for each missing/bad file type.

drizzlepac/astrodrizzle.py
stsci-hack: Implemented in https://github.com/spacetelescope/drizzlepac/pull/301/commits/95f65b33f237cc3313feaf05492717f6714443ef

stsci-hack:
Not merging this now since it is not needed as originally expected.  If needed at some point later, it can be added to the code at that point.

PR #303, opened 2019-04-04
Add headerlet docs

stsci-hack:
A first draft of some documentation on the new astrometric solutions and headerlets has been added to this package's doc directory for inclusion in the readthedocs output.   A new section was added at the top level labelled "Astrometry and Drizzle Products" which introduces and points to the new documentation.  The new docs are split into 2 components: end-user directed explanations and examples in the first section, and the usual API for all the new HLA-based alignment code based on the code's docstrings.   

This documentation also needs to be reviewed by INS in order to get their approval for the contents prior to merging.   

The intent is to have this as part of this initial public release, even if incomplete.  We can then update it further as we prepare the code to become operational.


stsci-hack:
Replaced by #344

Issue #295, opened 2019-03-27
Address code and efficiency clean-up as noted in PR#283/PR#259


mdlpstsci:
The purpose of this ticket is to document there is code and efficiency clean-up which needs to be done to the alignment portion of the drizzlepac package.  The comments and details can be found as part of the code review done on PR#283 and PR#259.  While some items were addressed, particularly bugs, the remaining items are blanket covered by this issue.  This ticket should be subdivided into more targeted issues to avoid code smuggling which has already happened in order to satisfy PR#283.

dulude:
PR #298  addresses two items raised in #283  : Namely [this one](https://github.com/spacetelescope/drizzlepac/pull/283#discussion_r268901463), and [this one](https://github.com/spacetelescope/drizzlepac/pull/283#discussion_r268903028).

PR #309, opened 2019-04-04
Address PR283 review comments

stsci-hack:
This PR will implement changes identified during the code review of PR #283 which were deferred and documented in #295 .  Multiple commits will be needed to resolve all the issues identified in #295.


tests/hla/base_test.py
```diff
@@ -343,22 +343,16 @@ def print_wcs(self, title, wcs):
...
         """
-        Read the image from a fits file
+        Read the image from first extension of a fits file
         """
-        hdu = fits.open(filename)
-
-        image = hdu[1].data
-        hdu.close()
-        return image
+        return fits.getdata(filename,ext=1)
```
mcara:
```suggestion
        return fits.getdata(filename, ext=1)
```

tests/hla/base_test.py
```diff
@@ -343,22 +343,16 @@ def print_wcs(self, title, wcs):
...
     def read_wcs(self, filename):
         """
         Read the wcs of a fits file
         """
-        hdu = fits.open(filename)
-
-        wcs = stwcs.wcsutil.HSTWCS(hdu, 1)
-        hdu.close()
+        from fits.open(filename) as hdu:
+            wcs = stwcs.wcsutil.HSTWCS(hdu, 1)
```
mcara: I would just replace this with:
```python
wcs = stwcs.wcsutil.HSTWCS(filename, 1)
```

mcara: and delete `from fits.open(filename) as hdu:` line above

mcara:
Also, I would suggest the following change to: https://github.com/spacetelescope/drizzlepac/blob/bdadf13688e47a18d9ae0882d8a19207343d278f/tests/hla/base_test.py#L149-L161

```
if isinstance(output_truth, tuple):
    output_spec = {'files':(output_file, output_truth[0]),
                   'pars': output_truth[1]}    
else:
    if output_hdus:
        output_spec = (output_file, output_truth, output_hdus)
    else:
        output_spec = (output_file, output_truth)
```

tests/hla/base_test.py
jhunkeler: ^ `from` should be be `with` (and/or these lines can removed as @mcara suggested).

stsci-hack:
All issues addressed in #339.

PR #310, opened 2019-04-04
Remove catalog_utils; move functionality to test_randomlist

mcara:
This PR moves the only function from `catalog_utils.py` - `randomSelectFromCSV()` - to `test_randomlist.py` where it is used.


tests/hla/test_randomlist.py
```diff
@@ -79,8 +132,9 @@ def test_align_randomFields(self):
 
         # Randomly select a subset of field names (each field represented by a row) from
         # the master CSV file and return as an Astropy table
-        randomCandidateTable = catutils.randomSelectFromCSV(input_file_path[0],
-            inputNumEntries, inputSeedValue)
+        randomCandidateTable = catutils._random_select_from_csv(
```
stsci-hack: This function is no longer coming from `catutils`...

mdlpstsci:
Oops. Sorry.  @stsci-hack should I go ahead and update 315?

stsci-hack:
The changes to alignimages noted here will need to be compared against the changes identified in PR #314 (and even #318).

mcara:
I do not know how to dig myself out of the git hole that I got myself into with all this rebasing, etc. So, I will be closing this PR without merging and re-do it in another PR.

PR #311, opened 2019-04-04
fix pep8

nden:
This contains only pep8 changes and general clean up.
`countExtn` is now imported from `stsci_tools.fileutil`.

### DIFF
--- a/drizzlepac/hlautils/astrometric_utils.py
+++ b/drizzlepac/hlautils/astrometric_utils.py
@@ -1,8 +1,6 @@
 """Utilities to support creation of astrometrically accurate reference catalogs
-        print("No detected sources!")
 
 The function, create_astrometric_catalog, allows the user to query an
-        print("No detected sources!")
 astrometric catalog online to generate a catalog of astrometric sources that
 should fall within the field-of-view of all the input images.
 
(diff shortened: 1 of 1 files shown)


drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -884,31 +874,33 @@ def read_hlet_wcs(filename, ext):
     """
     hstwcs = wcsutil.HSTWCS(filename, ext=ext)
     if hstwcs.naxis1 is None:
```
mcara: @nden Is there any circumstance under which `naxis` could be `None` in a WCS?

nden: Only if you create a default WCS without passing a data file.
Unless `filename` above can be `None`, `naxis1/2` should always be populated.

mcara: @nden Thanks! I was just curious (+wanted to e sure)

mcara: @nden Then maybe this `if` can be removed?

nden: This PR is simply PEP8 changes. I'll try to review the code today and will include it there.

PR #312, opened 2019-04-04
fix docstrings in astrometric_utils

nden:
This PR fixes the docstrings in astrometric_utils.
I built the sphinx documentation locally to see how they are but there are general issues with the sphinx configuration of drizzlepac that prevent it from displaying properly (I think, or I did something wrong).
In any casee it builds, just doesn't look good.

### DIFF
--- a/doc/source/hla.rst
+++ b/doc/source/hla.rst
@@ -1,0 +1,21 @@
+.. hla:
+
+HLA alignment API
+-----------------
+
+
+.. automodule:: drizzlepac.hlautils.astrometric_utils
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.create_astrometric_catalog
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.get_catalog
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.find_gsc_offset
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.extract_sources
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.classify_sources
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.generate_source_catalog
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.generate_sky_catalog
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.compute_photometry
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.filter_catalog
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.build_self_reference
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.within_footprint
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.create_image_footprint
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.find_hist2d_offset
+.. autofunction:: drizzlepac.hlautils.astrometric_utils.build_wcscat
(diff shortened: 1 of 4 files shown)


drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -189,7 +165,7 @@ def create_astrometric_catalog(inputs, **pars):
     for source in ref_dict:
         if 'GAIAsourceID' in source:
             g = source['GAIAsourceID']
-            if gaia_only and g.strip() is '':
+            if gaia_only and g.strip() == '':
```
mcara: Several issues here:
1. [minor,style]: `if gaia_only and not g.strip()'':` (empty strings, lists evaluate to `False`)
2. [inconsistency]: `g.strip()` implies that `g` is a string while `g = -1` below makes it an integer. Then in the line below, `ref_table.add_row((r, d, m, o, g))` what is the column type for `g`? Maybe, instead of assigning a negative integer, assign a string: `g = '-1'`?

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -189,7 +165,7 @@ def create_astrometric_catalog(inputs, **pars):
...
                 continue
         else:
             g = -1  # indicator for no source ID extracted
```
stsci-hack: As noted by @mcara , the value of `g` should be set here to `"-1"`, since the table row has that entry defined with dtype of `U25`.  The code will work with or without this change, though, since astropy `Table()` objects have no problems with the values either way.

doc/source/hla.rst
```diff
@@ -0,0 +1,21 @@
+.. hla:
+
+HLA alignment API
```
stsci-hack: The Mission Office has expressed an interest to NOT refer to these new products as HLA, since that implies a product generated by the separate HLA effort.  Instead, we have been asked to refer to this code and the products as 'Enhanced Pipeline Products' in the documentation, at least until some better term is decided upon by the Mission Offices.

doc/source/index.rst
```diff
@@ -117,6 +117,12 @@ The task 'runastrodriz' can be used to reproduce the same Drizzle processing tha
 
    Running Astrodriz <runastrodriz.rst>
 
+HLA API
```
stsci-hack: The Mission Office has expressed an interest to NOT refer to these new products as HLA, since that implies a product generated by the separate HLA effort.  Instead, we have been asked to refer to this code and the products as 'Enhanced Pipeline Products' in the documentation, at least until some better term is decided upon by the Mission Offices.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -70,57 +68,35 @@
...
 MODULE_PATH = os.path.dirname(inspect.getfile(inspect.currentframe()))
 VEGASPEC = os.path.join(os.path.dirname(MODULE_PATH),
-                        'data','alpha_lyr_stis_008.fits')
+                        'data', 'alpha_lyr_stis_008.fits')
 
 __all__ = ['create_astrometric_catalog', 'compute_radius', 'find_gsc_offset',
-            'extract_sources', 'find_hist2d_offset', 'generate_source_catalog',
-            'classify_sources', 'countExtn']
+           'extract_sources', 'find_hist2d_offset', 'generate_source_catalog',
+           'classify_sources', 'countExtn']
```
stsci-hack: The entry for 'countExtn' is invalid here, since that function was removed from this module.

PR #314, opened 2019-04-08
Astrometric code review

nden:
This PR is based on and assumes #312 is merged. I will rebase when #312 is merged.
All code changes are suggestions. I added comments to explain them.

### DIFF
--- a/drizzlepac/hlautils/astrometric_utils.py
+++ b/drizzlepac/hlautils/astrometric_utils.py
@@ -31,7 +31,7 @@ except Exception:
 from astropy import units as u
 from astropy.table import Table, vstack
 from astropy.coordinates import SkyCoord
-from astropy.io import fits as pf
+from astropy.io import fits as fits
 from astropy.io import ascii
 from astropy.convolution import Gaussian2DKernel
 from astropy.stats import gaussian_fwhm_to_sigma
@@ -91,7 +91,8 @@ Primary function for creating an astrometric reference catalog.
 """
 
 
-def create_astrometric_catalog(inputs, **pars):
+def create_astrometric_catalog(inputs, catalog="GAIADR2", output="ref_cat.ecsv",
+                               gaia_only=False, table_format="ascii.ecsv", existing_wcs=None):
     """Create an astrometric catalog that covers the inputs' field-of-view.
 
     Parameters
@@ -110,7 +111,6 @@ def create_astrometric_catalog(inputs, **pars):
 
     gaia_only : bool, optional
         Specify whether or not to only use sources from GAIA in output catalog
-        Default: False
 
     existing_wcs : ~stwcs.wcsutils.HSTWCS`
         existing WCS object specified by the user
@@ -126,20 +126,14 @@ def create_astrometric_catalog(inputs, **pars):
         Astropy Table object of the catalog
 
     """
-    # interpret input parameters
-    catalog = pars.get("catalog", 'GAIADR2')
-    output = pars.get("output", 'ref_cat.ecsv')
-    gaia_only = pars.get("gaia_only", False)
-    table_format = pars.get("table_format", 'ascii.ecsv')
-    existing_wcs = pars.get("existing_wcs", None)
-
     inputs, _ = parseinput.parseinput(inputs)
+
     # start by creating a composite field-of-view for all inputs
     # This default output WCS will have the same plate-scale and orientation
     # as the first chip in the list, which for WFPC2 data means the PC.
     # Fortunately, for alignment, this doesn't matter since no resampling of
     # data will be performed
-    if existing_wcs:
+    if existing_wcs is not None:
         outwcs = existing_wcs
     else:
         outwcs = build_reference_wcs(inputs)
@@ -285,8 +279,8 @@ def find_gsc_offset(image, input_catalog='GSC1', output_catalog='GAIA'):
     serviceType = "GSCConvert/GSCconvert.aspx"
     spec_str = "TRANSFORM={}-{}&IPPPSSOOT={}"
 
-    if 'rootname' in pf.getheader(image):
-        ippssoot = pf.getval(image, 'rootname').upper()
+    if 'rootname' in fits.getheader(image):
+        ippssoot = fits.getval(image, 'rootname').upper()
     else:
         ippssoot = fu.buildNewRootname(image).upper()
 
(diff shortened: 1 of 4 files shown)


drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -33,7 +31,7 @@
 from astropy import units as u
 from astropy.table import Table, vstack
 from astropy.coordinates import SkyCoord
-from astropy.io import fits as pf
+from astropy.io import fits as fits
```
nden: It's now standard to import for the old pyfits as `fits`.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -70,57 +68,36 @@
...
 """
 
 Primary function for creating an astrometric reference catalog.
 
 """
-def create_astrometric_catalog(inputs, **pars):
+
+
+def create_astrometric_catalog(inputs, catalog="GAIADR2", output="ref_cat.ecsv",
+                               gaia_only=False, table_format="ascii.ecsv", existing_wcs=None):
```
nden: In general it's more clear to set the default values of the keywords arguments in the function signature, especially if the number of arguments is not that big, as opposed to using `**kwargs` and then popping default values in the code. This allows a user to use simple (i)python tools to find out the default values. It also allows tools like pycodestyle and flake8 to catch mismatch of arguments, which was an issue in at least two functions where the docstrings defined parameters which were not used in the function.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -403,30 +354,44 @@ def extract_sources(img, **pars):
...
             bkg_rms = (5. * bkg.background_rms)
             bkg_rms_mean = bkg.background.mean() + 5. * bkg_rms.std()
             default_threshold = bkg.background + bkg_rms
-            if threshold is None or threshold < 0.0:
-                if threshold is not None and threshold < 0.0:
-                    threshold = -1*threshold*default_threshold
-                    log.info("{} based on {}".format(threshold.max(), default_threshold.max()))
-                    bkg_rms_mean = threshold.max()
-                else:
-                    threshold = default_threshold
```
nden: This was a bit obscure to me. A suggestioon is below but it needs to be double checked that it reproduces the desired functionality.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -871,187 +804,193 @@ def build_self_reference(filename, clean_wcs=False):
...
-        based on user provided parameter values.
 
-        .. note :: COPIED from drizzlepac.wcs_functions
+def build_hstwcs(crval, crpix, naxis1, naxis2, pscale, orientat):
+    """ Create an `stwcs.wcsutil.HSTWCS` object for a default instrument without
+    distortion based on user provided parameter values.
     """
     wcsout = wcsutil.HSTWCS()
-    wcsout.wcs.crval = np.array([crval1,crval2])
-    wcsout.wcs.crpix = np.array([crpix1,crpix2])
```
nden: `wcs.wcs.crval` is already an array.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -871,187 +804,193 @@ def build_self_reference(filename, clean_wcs=False):
...
+    #mask = ~np.bitwise_or(maskx, masky)
+    #x = x[mask]
+    #y = y[mask]
+
+    ## Now, confirm that these points fall within actual science area of WCS
+    #img_mask = create_image_footprint(img, wcs, border=1.0)
+    #inmask = np.where(img_mask[y.astype(np.int32), x.astype(np.int32)])[0]
+
+    sky = wcs.pixel_to_world(x, y, 1)
+    inmask = wcs.footprint_contains(sky)
```
nden: `wcs.footprint_contains`  is a new addition to astropy v 3.1. I am adding it as a suggestion here as it simplifies the code.
You may need to loop over all chips though in order to reproduce the current functionality.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -70,57 +68,36 @@
...
 MODULE_PATH = os.path.dirname(inspect.getfile(inspect.currentframe()))
 VEGASPEC = os.path.join(os.path.dirname(MODULE_PATH),
-                        'data','alpha_lyr_stis_008.fits')
+                        'data', 'alpha_lyr_stis_008.fits')
 
 __all__ = ['create_astrometric_catalog', 'compute_radius', 'find_gsc_offset',
-            'extract_sources', 'find_hist2d_offset', 'generate_source_catalog',
-            'classify_sources', 'countExtn']
+           'extract_sources', 'find_hist2d_offset', 'generate_source_catalog',
+           'classify_sources', 'countExtn']
```
stsci-hack: Since `countExtn` is being removed from here in favor of importing it from stsci.tools, the entry for this function in `__all__` can be removed.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -70,57 +68,36 @@
...
 
-"""
-Utility functions.
 
-Many were originally released in stsci.tools.fileutil.
-"""
 def buildRotMatrix(theta):
```
stsci-hack: If `countExtn` was removed in favor of using it directly from stsci.tools, the same could be done with this function as well.

drizzlepac/hlautils/astrometric_utils.py
stsci-hack: Your changes seem like they cover all the cases with much simpler logic.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -441,42 +406,42 @@ def extract_sources(img, **pars):
...
         log.info("Setting up DAOStarFinder with: \n    fwhm={}  threshold={}".format(fwhm, bkg_rms_mean))
         daofind = DAOStarFinder(fwhm=fwhm, threshold=bkg_rms_mean)
         # Identify nbrightest/largest sources
         if nlargest is not None:
-            if nlargest > len(segm.labels):
-                nlargest = len(segm.labels)
-            large_labels = np.flip(np.argsort(segm.areas)+1)[:nlargest]
+            #if nlargest > len(segm.labels):
+                #nlargest = len(segm.labels)
+            nlargest = min(nragest, len(segm.labels))
```
stsci-hack: Looks like a typo here: `nragest` should be `nlargest`.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -523,7 +488,7 @@ def extract_sources(img, **pars):
         log.info("Wrote source catalog: {}".format(output))
 
     if plot and plt is not None:
-        norm = None
+        norm = Nonelen(segm.labels)
```
stsci-hack: It appears that you meant to replace `None` with `len(segm.labels)`.

stsci-hack:
This PR will be replaced by #323, where problems identified by Azure/Travis here are corrected.

PR #315, opened 2019-04-08
Updates made for regression tests based upon code review comments

mdlpstsci:
This ticket contains changes made in particular for the regression tests based upon code review comments.  Due to the number and nature of other tickets generated during the review, this code will probably need to be updated before it can be merged.


drizzlepac/hlautils/testutils.py
```diff
@@ -77,6 +79,11 @@ def compare_wcs_alignment(dataset, force=False):
         log.info("Default WCSNAME: {}".format(default_wcsname))
         alignment = {default_wcsname:extract_results(results)}
 
+        # Download the calibration reference files to ensure availability
+        ref_files = ref_from_image(imglist[0], ['IDCTAB'])
```
stsci-hack: ACS and WFC3 observations require the use of the IDCTAB, DGEOFILE and NPOLFILE reference files.

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
+import sys
+import traceback
+import os
+import datetime
+import pytest
```
stsci-hack: Imported but not used...

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+import numpy as np
+from astropy.table import Table, vstack
+from astropy.io import ascii
```
stsci-hack: Imported but not used...

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+from .base_test import BaseHLATest
+from drizzlepac import alignimages
+import drizzlepac.hlautils.catalog_utils as catutils
+
+class TestAlignMosaic(BaseHLATest):
+    """ Process a large sample of ACS and WFC3 datasets to determine if they can be
+        aligned to an astrometric standard.
+    """
+
+    def test_align_randomFields(self):
```
stsci-hack: PEP8: method name should be all lower-case

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+        information regarding the images
+
+    Returns
+    =======
+    datasetNames: list
+        List of individual image or association base (IPPSSOOT) names
+    """
+
+    #dataFromTable = Table.read(filename, format='ascii')
+    datasetIDs = tableName['observationID']
```
stsci-hack: Since these variables are not used anywhere else, it would be simpler to simply use the column names directly in the code below, making the definitions for datasetIDs and asnIDs un-necessary.

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+    asnIDs     = tableName['asnID']
+
+    datasetNames = []
+
+    # Determine if the data is part of an association or is an individual image
+    for imgid,asnid in zip(datasetIDs,asnIDs):
```
stsci-hack:
```suggestion
    for imgid,asnid in zip(tableName['observationID'],tableName['asnID']):
```

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+            astrometric standard.  The success or fail status for each test is retained
+            as the overall success or fail statistic is the necessary output from
+            this test.
+        """
+
+        numSuccess = 0
+        numQualSuccess = 0
+        numUnsuccessful = 0
+        numException = 0
+        numProcessedDatasets = 0
```
stsci-hack: This is defined in Line 76, so this line should not be needed.

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+        for filename in os.listdir():
+            if filename.endswith('flt.fits') or filename.endswith('flc.fits'):
+                os.unlink(filename)
+
+        # Write out the table
+        allDatasetTable.write('resultsBigTest.ecsv', format='ascii.ecsv')
+
+        # Determine the percent success over all datasets processed
+        percentSuccess = numSuccess/numProcessedDatasets
+        print('TEST_RANDOM. Number of tests started: ', numStartTests)
```
stsci-hack: You could avoid the need for the variable `numStartTests` by replacing the original definition of this variable on Line 77 with this print statement.

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+        # If the dataset name represents an association ID, the multiplicity
+        # of images within the association need to be processed.  Otherwise,
+        # the dataset is a single image.
+        #
+        # If the "alignment" of a field/dataset fails for any reason, trap
+        # the exception and keep going.
+        allDatasetTable = Table()
+        datasetKey = -1
+        print("TEST_RANDOM. Dataset List: ", dataset_list)
+        for dataset in dataset_list:
```
stsci-hack: This is one instance where you could use `enumerate(dataset_list)` to define `datasetKey` automatically, rather than manually incrementing a separately defined variable.  No change in function, but perhaps less room for mistakes in incrementing this variable.  This suggestion, though, would not be appropriate if additional logic was being applied to incrementing this variable, but I could not find any in this specific method.

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+                continue
+
+        # Perform some clean up
+        if os.path.isfile('ref_cat.ecsv'): 
+            os.remove('ref_cat.ecsv')
+        if os.path.isfile('refcatalog.cat'):  
+            os.remove('refcatalog.cat')
+        for filename in os.listdir():
+            if filename.endswith('flt.fits') or filename.endswith('flc.fits'):
+                os.unlink(filename)
```
stsci-hack: It may be less confusing to maintain if `os.remove` was used here as well.

tests/hla/process_randomlist.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # Write out the table
+        allDatasetTable.write('resultsBigTest.ecsv', format='ascii.ecsv')
```
stsci-hack: Would it be more useful for comparing with other runs to have a date-time-stamped filename instead of a fixed generic filename?  Admittedly, a time-unique name would definitely not be needed just for running the test under pytest, so this is merely a suggested issue to consider when time permits.

tests/hla/test_randomlist.py
```diff
@@ -1,236 +1,55 @@
...
-import datetime
 import pytest
-import numpy as np
-from astropy.table import Table, vstack
-from astropy.io import ascii
-
-from .base_test import BaseHLATest
-from drizzlepac import alignimages
-import drizzlepac.hlautils.catalog_utils as catutils
+import drizzlepac
```
stsci-hack: Imported but not used...

tests/hla/process_randomlist.py
mdlpstsci: OK - fixed.

tests/hla/process_randomlist.py
mdlpstsci: Not sure what you mean.  These are used, except for pytest.

tests/hla/process_randomlist.py
mdlpstsci: These are used - will keep.

tests/hla/process_randomlist.py
mdlpstsci: OK - fixed.

tests/hla/process_randomlist.py
mdlpstsci: Agreed - fixed.

tests/hla/process_randomlist.py
mdlpstsci: Agreed - fixed.

tests/hla/process_randomlist.py
mdlpstsci: Agreed - fixed.

tests/hla/process_randomlist.py
stsci-hack: The packages 'ascii' and pytest' are not used anywhere in this code, so those imports can be removed.  The remaining import statements are used in the code and should obviously remain.

tests/hla/process_randomlist.py
mdlpstsci: Point taken though I am going to leave this alone as this variable (and the column in the output table) seems to be on its way to being deprecated.

tests/hla/process_randomlist.py
mdlpstsci: I should have done this from the start.  Updated using seconds since the epoch.

tests/hla/process_randomlist.py
mdlpstsci: Agreed - fixed.

tests/hla/test_randomlist.py
mdlpstsci: Removed - Fixed.

mdlpstsci:
All the comments have been addressed.  The test_randomlist.py will have one small fix needed if PR#310 is merged before this PR.

PR #317, opened 2019-04-08
Alignimages cleanup

dulude:
Refactored alignimages.py:
- Fixed PEP 8 violations
- Verified/corrected subroutine docstring formatting
- Verified subroutine definitions
- Corrected a few misspellings


stsci-hack:
The PEP8 changes identified here will need to be compared against the changes identified independently in PRs #314 and #310 for alignimages.

Issue #319, opened 2019-04-09
IVM computed by astrodrizzle (incorrectly) assumes flat field identical to 1


mcara:
Starting with version `1.12` of `numpy`, slicing `numpy` arrays with floating point numbers raises an exception. The code in the `drizzlepac` for retrieving flat field data uses floating point indices AND the entire code is encased in a `try-except` block that catches all errors but reports them as "flatfield file not found" type of error.

The error was first reported by user in helpdesk issue INC0134878. Due to the the catch-all  `try-except` the following would be written in the log:
```
[INFO    ] Automatically creating IVM files for chip 1
[WARNING ] Cannot find file /.../hst/cdbs/iref/uc72113pi_pfl.fits.
    Treating flatfield constant value of '1'.
```
The buggy code:
https://github.com/spacetelescope/drizzlepac/blob/baae7a3781775d570517166c69da442e734aa0d6/drizzlepac/imageObject.py#L534-L553

PR #320, opened 2019-04-09
Bug fix in the automatic computation of the IVM weights

mcara:
This PR fixes the bug described in #319.

CC: @jhunkeler

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,9 @@ of the list).
 3.0.2 (unreleased)
 ==================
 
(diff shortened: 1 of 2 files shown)


mcara:
@stsci-hack I think I made a mistake in the original commit because I switched from `fileutil.openImage()` to `fits.getdata()`. This would be OK if it is guaranteed that all flat field files are standard `FITS` files (no WAIVER FITS/GEIS). Can we make this assumption? Should I revert to the original code?

mcara:
I decided to switch back to `fileutil.openImage()` just to be on the safe side and continue supporting older files formats for the flat field files.

PR #325, opened 2019-04-11
Display algorithm name at the end of alignimages run and in headerlet file

dulude:
[HLA-227](https://jira.stsci.edu/browse/HLA-227)

- Added summary table that displays basic fit into for each astrometric catalog/fit method combination run.
- Added 'fit method' to filtered_table 
- Added fit method as new headerlet primary extension field 'FITMETH'
- Added fit method to headerlet wcsname.


dulude:
See PR #327 for a version of this PR with code that successfully merged with the current version of upstream/hla-staging.

PR #328, opened 2019-04-12
change processing order so that match_relative_fit is the preferred fit method

dulude:
- JIRA ticket [HLA-228](https://jira.stsci.edu/browse/HLA-228)

Switched order of processing loops so the code now loops over fit method, then astrometric catalog, rather then astrometric catalog, then fit method as it previously did. This will ensure that the code will always try match_relative_fit with both possible astrometric catalogs before anything else.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -131,7 +131,7 @@ def check_and_get_data(input_list, **pars):
 
         # Input is an ipppssoot (association or singleton), nine characters by definition.
         # This "else" block actually downloads the data specified as ipppssoot.
-        elif len(input_item) is 9:
+        elif len(input_item) == 9:
             try:
                 if input_item not in ipppssoot_list:
                     # An ipppssoot of an individual file which is part of an association cannot be retrieved
@@ -321,7 +321,7 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
         # data, 'doProcess=0' and 'status=9999' in the table (the status value by default), so there is no
         # need to update the filtered_table here.
 
-        if filtered_table['doProcess'].sum() is 0:
+        if filtered_table['doProcess'].sum() == 0:
             log.warning(
                 "No viable images in filtered table - no processing done.\n")
             current_dt = datetime.datetime.now()
@@ -530,7 +530,7 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
                         # total_rms
                         if fit_quality < 5:
                             # valid, non-compromised solution with total rms < 10 mas...go with this solution.
-                            if fit_quality is 1:
+                            if fit_quality == 1:
                                 best_fit_rms = fit_rms
                                 for item in imglist:
                                     item.best_meta = item.meta.copy()
@@ -545,7 +545,7 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
                                     item.best_meta = item.meta.copy()
                                 best_fit_status_dict = fit_status_dict.copy()
                                 best_fit_qual = fit_quality
-                            elif fit_quality is best_fit_qual:
+                            elif fit_quality == best_fit_qual:
                                 # new solution same level of fit_quality. Choose whichever one has the lowest
                                 # total rms as "best" and keep looping.
                                 if best_fit_rms >= 0.:
@@ -1142,7 +1142,7 @@ def update_image_wcs_info(tweakwcs_output):
     for item in tweakwcs_output:
         img_name = item.meta['filename']
         chipnum = item.meta['chip']
-        if chipnum is 1:
+        if chipnum == 1:
             chipctr = 1
             hdulist = fits.open(img_name, mode='update')
             num_sci_ext = amutils.countExtn(hdulist)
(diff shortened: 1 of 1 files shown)


drizzlepac/alignimages.py
```diff
@@ -450,39 +450,47 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
-        for catalogIndex in range(0, len(
-                catalog_list)):  # loop over astrometric catalog
-            log.info("{} STEP 5: Detect astrometric sources {}".format("-" * 20, "-" * 48))
-            log.info("Astrometric Catalog: %s", str(catalog_list[catalogIndex]))
-            reference_catalog = generate_astrometric_catalog(process_list,
-                                                             catalog=catalog_list[catalogIndex],
-                                                             output=output)
+        reference_catalog_dict={}
+        for algorithm_name in fit_algorithm_list:  # loop over fit algorithm type
+            for catalogIndex in range(0, len(catalog_list)):  # loop over astrometric catalog
```
stsci-hack:
```suggestion
            for catalogIndex, catalog_name in enumerate(catalog_list):  # loop over astrometric catalog
```
This would eliminate the need for line 457. Where the only entries in `catalog_list` are strings, there is no need to cast them to strings in line 457 either.

drizzlepac/alignimages.py
```diff
@@ -450,39 +450,47 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
+                log.info("{} STEP 5: Detect astrometric sources {}".format("-" * 20, "-" * 48))
+                catalog_name = str(catalog_list[catalogIndex])
```
stsci-hack: Don't need this if using the suggestion for line 455.

drizzlepac/alignimages.py
```diff
@@ -450,39 +450,47 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
+                log.info("Astrometric Catalog: {}".format(catalog_name))
+                # store reference catalogs in a dictionary so that generate_astrometric_catalog() doesn't
+                #  execute unnecessarily after it's been run once for a given astrometric catalog.
+                if catalog_name in reference_catalog_dict.keys():
```
stsci-hack: Use of `.keys()` here is not functionally necessary, although it is more explicit.

drizzlepac/alignimages.py
```diff
@@ -450,39 +450,47 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
-                fit_quality = 5
-                if catalogIndex < len(catalog_list) - 1:
-                    log.info("Try again with other catalog")
+                current_dt = datetime.datetime.now()
+                delta_dt = (current_dt - starting_dt).total_seconds()
+                log.info('Processing time of [STEP 5]: {} sec'.format(delta_dt))
+                starting_dt = current_dt
+
+                if len(reference_catalog) < MIN_CATALOG_THRESHOLD:
+                    log.warning("Not enough sources found in catalog {}".format(catalog_list[catalogIndex]))
```
stsci-hack:
```suggestion
                    log.warning("Not enough sources found in catalog {}".format(catalog_name))
```

drizzlepac/alignimages.py
dulude: Resolved issue with commit [bf5ab62](https://github.com/spacetelescope/drizzlepac/pull/328/commits/bf5ab623c7cd17cc5e11cbdf9572828829fa5235)

drizzlepac/alignimages.py
dulude: Resolved issue with commit [bf5ab62](https://github.com/spacetelescope/drizzlepac/pull/328/commits/bf5ab623c7cd17cc5e11cbdf9572828829fa5235)

drizzlepac/alignimages.py
dulude: issue resolved with commit [bf5ab62](https://github.com/spacetelescope/drizzlepac/pull/328/commits/bf5ab623c7cd17cc5e11cbdf9572828829fa5235)

drizzlepac/alignimages.py
dulude: Issue resolved by commit [cb068eb](https://github.com/spacetelescope/drizzlepac/pull/328/commits/cb068eb71b8047f72f4f3a8d8951faab996d40c7)

drizzlepac/alignimages.py
```diff
@@ -557,17 +564,17 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
                         filtered_table['status'][:] = 1
                         filtered_table['processMsg'][:] = "Fitting failure"
                         # It may be there are additional catalogs and algorithms to try, so keep going
                         fit_quality = 5  # Flag this fit with the 'bad' quality value
                         filtered_table['fit_qual'][:] = fit_quality
                         continue
-                    if fit_quality is 1:  # break out of inner fit algorithm loop
-                        break
-            if fit_quality is 1:  # break out of outer astrometric catalog loop
+                if fit_quality is 1:  # break out of inner astrometric catalog loop
```
stsci-hack:
```suggestion
                if fit_quality == 1:  # break out of inner astrometric catalog loop
```

drizzlepac/alignimages.py
```diff
@@ -557,17 +564,17 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
+                    break
+            if fit_quality is 1:  # break out of outer fit algorithm  loop
```
stsci-hack:
```suggestion
            if fit_quality == 1:  # break out of outer fit algorithm  loop
```

stsci-hack:
We need to investigate why Jenkins could not build this last commit before we can merge...

PR #329, opened 2019-04-15
Implemented parallel testing with user input file list

stsci-hack:
This shows how testing of multiple datasets can be performed in parallel with the user providing the master list on input.  It relies on pytest to report failures and skips rather than computing the 70% pass/fail criteria itself.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,6 +1,4 @@
 """Project default for pytest"""
-
-from astropy.tests.plugins.display import PYTEST_HEADER_MODULES
 from astropy.tests.helper import enable_deprecations_as_exceptions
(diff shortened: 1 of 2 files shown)


stsci-hack:
The changes demonstrated here will be replaced by a more correct and complete implementation by @mdlpstsci , at which time, this PR can be closed without merging.

stsci-hack:
@mdlpstsci : This version of the randomlist test now sorts the results into a separate directory for each dataset while allowing xdist to use multiple CPUs to run tests in parallel.

stsci-hack:
Replaced with #341.

PR #331, opened 2019-04-16
Update FWHM values to measure PSFs more accurately

stsci-hack:
These changes to the FWHM values for the instruments generate more accurate source positions, especially for WFC3/IR, and subsequently make the cross-matching and fitting more reliable.


stsci-hack:
To be replaced by #340 which correctly and verifiably has 'hla-staging' as parent instead.

PR #341, opened 2019-04-23
Parallelized testing

mdlpstsci:
Parallelized testing implemented for test_randomlist.py with command line options to specify the master list, starting row, and number of rows (aka datasets) to process.


tests/hla/test_randomlist.py
```diff
@@ -1,14 +1,38 @@
...
 import pytest
-from .process_randomlist import TestAlignMosaic
+
+from drizzlepac import alignimages
+
+def pytest_generate_tests(metafunc):
+    """Get the command line option."""
+    start_row = metafunc.config.option.start_row
+    num_rows = metafunc.config.option.num_rows
+    master_list = metafunc.config.option.master_list
```
stsci-hack:
```suggestion
    master_list = metafunc.config.option.master_list
    # Check to see if the specified file exists in the current working directory
    if not os.path.exists(master_list):
        # If not, see about copying the (default) file from the installation directory
        # Find where this module has been installed.
        install_dir = os.path.dirname(__file__)
        default_file = os.path.join(install_dir,master_list)
        if os.path.exists(default_file):
            # Copy file --- REQUIRES 'import shutil' to be added as well
                        shutil.copy2(default_file,os.getcwd())    
```

mdlpstsci: OK.  Will add to the source code.

PR #346, opened 2019-04-24
junit_xml is not a valid keyword

jhunkeler:
`results.xml` is invalid and cannot be parsed by `xunit2`.


mcara:
When and why did `results.xml` become invalid?

PR #348, opened 2019-04-24
Updated datasets for this test

stsci-hack:
Changes to correctly handle different instruments reference files and to update the test datasets to ones which more easily demonstrate the effects of the a priori solutions.  

These changes should address the problems seen in the regression tests with failures due to reference file problems.

### DIFF
--- a/tests/hla/test_apriori.py
+++ b/tests/hla/test_apriori.py
@@ -6,15 +6,14 @@ import numpy as np
 
 from drizzlepac.hlautils import testutils
 
-from ..resources import BaseACS
+from ..resources import BaseACS, BaseWFC3
 
 class TestAcsApriori(BaseACS):
     """ Tests which validate whether mosaics can be aligned to an astrometric standard,
         evaluate the quality of the fit, and generate a new WCS.
     """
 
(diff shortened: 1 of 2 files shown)


stsci-hack:
The code for this test has been reorganized to share the comparison function between 2 instrument-specific classes.

tests/hla/test_apriori.py
```diff
@@ -6,91 +6,102 @@
...
+
+        # Check that rotation and scale are within
+        delta_rot = np.abs(results['rotation'] - pipeline_results['rotation']) / pipeline_results['rotation']
+        delta_scale = np.abs(results['scale'] - pipeline_results['scale']) / pipeline_results['scale']
+        rot = (delta_rot < limit).all()
+        scale = (delta_scale < limit).all()
+
+        # Determine success/failure of this dataset's fit
+        wcs_success = status and fit_qual and delta and rot and scale
+        if wcs_success:
```
mcara:
```python
if all([status, fit_qual, delta, rot, scale]):
```
If accepted, make sure to delete previous line.

stsci-hack:
All review comments have been implemented in 4c91ec6.

stsci-hack:
If it is any relief, the updated version of this test passes when run interactively. ;-)

PR #349, opened 2019-04-25
Integrate HLA

jhunkeler:
For the sake of the project's integrity and the preservation of events, I think a merge-commit is in order. 

```
git checkout master
git checkout -tb integrate-hla
git merge -X theirs upstream/hla-staging
git push origin
# make the PR
```


drizzlepac/alignimages.py
```diff
@@ -700,174 +818,188 @@ def determine_fit_quality(imglist,filteredTable, print_fit_parameters=True):
...
+            float(item.meta['fit_info']['shift'][0])**2 +
+            float(item.meta['fit_info']['shift'][0])**2) * item.wcs.pscale  # radial offset in arssec
+        if float(num_xmatches) * 0.36 > 0.8 + (radial_offset / 10.0)**8:
+            radial_offset_check = True
 
-        largeRmsCheck = True
+        large_rms_check = True
         if fit_rms_val > 150. or max_rms_val > 150.:
-            largeRmsCheck = False
+            large_rms_check = False
```
mcara: @jhunkeler Please, can we do this: `large_rms_check = fit_rms_val > 150. or max_rms_val > 150.`

drizzlepac/alignimages.py
```diff
@@ -700,174 +818,188 @@ def determine_fit_quality(imglist,filteredTable, print_fit_parameters=True):
...
 
-        consistencyCheck = True
+        consistency_check = True
         rms_limit = max(item.meta['fit_info']['TOTAL_RMS'], 10.)
-        if not math.sqrt(np.std(np.asarray(xshifts)) ** 2 + np.std(np.asarray(yshifts)) ** 2) <= (
-                    rms_limit / 1000.0) / (item.wcs.pscale):
-            consistencyCheck = False
+        if not math.sqrt(np.std(np.asarray(xshifts)) ** 2 + np.std(
+                         np.asarray(yshifts)) ** 2) <= (rms_limit / 1000.0) / (item.wcs.pscale):
+            consistency_check = False
```
mcara: @jhunkeler 
I think `if not x <= y` can be safely replaced with `if x > y:`
```python
rms_limit = max(item.meta['fit_info']['TOTAL_RMS'], 10.) / 1000.0 / item.wcs.pscale
consistency_check = math.sqrt(np.std(xshifts)**2 + np.std(yshifts)**2) <= rms_limit
```

mcara: @jhunkeler Actually, let's simplify this further:
```python
rms_limit = max(item.meta['fit_info']['TOTAL_RMS'], 10.) / 1000.0 / item.wcs.pscale
consistency_check = math.sqrt(np.var(xshifts) + np.var(yshifts)) <= rms_limit
```

mcara:
@jhunkeler Many thanks!

drizzlepac/alignimages.py
```diff
@@ -202,361 +197,480 @@ def perform_align(input_list, **kwargs):
...
+
+        for imgname in extracted_sources.keys():
+            table = extracted_sources[imgname]["catalog_table"]
+
+            # Get the location of the current image in the filtered table
+            index = np.where(filtered_table['imageName'] == imgname)[0][0]
+
+            # First ensure sources were found
+
+            if table is None or table[1] == None:
```
mcara:
```suggestion
            if table is None or table[1] is None: 
```

doc/source/index.rst
```diff
@@ -117,8 +117,8 @@ The task 'runastrodriz' can be used to reproduce the same Drizzle processing tha
 
    Running Astrodriz <runastrodriz.rst>
 
-Enhanced Pipeline Products API
-------------------------------
+HLA API
```
stsci-hack: The Mission Offices have determined that this code should NOT be referred to as "HLA" code in the documentation.  This change can not be included based on their decision.

drizzlepac/alignimages.py
stsci-hack: Actual testing of the code demonstrated that this change does NOT work as expected.

doc/source/index.rst
mcara:
```suggestion
Enhanced Pipeline Products API
```

drizzlepac/alignimages.py
mcara: I can't imagine under what circumstances something could be `== None` but not _be_ `None`.

mcara: See https://github.com/astropy/astropy/issues/8640.

pllim: Why not use `len(table) == 0`?

stsci-hack:
```suggestion
            # This logic is actually correct, since 'table={1:None}' if no sources are found.
            if table[1] == None:
```

mcara: I would say leave the code alone "as is" if it works for you. I do not understand the way the data are organized and so I may have missed something. If `table={1:None}` for the case of no sources, then I can't understand why my initial suggestion did not work:
```python
>>> a = {1:None}                                                                                                           
>>> a[1] is None                                                                                                           
True
>>> a[1] == None                                                                                                           
True
```
Most likely, however, (I feel) you have empty `astropy.table.Table` and if that is the case, you should change the code to:
```python
if table is None or not table[1]:
```
This last suggestion will work clean when `table[1]` is either `None` or an empty table. Comparing a table to `None` to check its length is _very_ unusual. (For example if `a=[]` I would do `if not a:` and not `if a == None:`)

mcara: @lim actually, it should be `len(table[1]) == 0`. But it looks like sometimes `table[1]` could be `None` (according to https://github.com/spacetelescope/drizzlepac/pull/349#discussion_r278698963). In that case `not table[1]` should address both situations.

stsci-hack:
```suggestion
            if table is None or not table[1]: 
```

pllim: Didn't know `not table` would work, cool!

stsci-hack: This final(?) suggested change actually works as needed while meeting PEP8.

mcara: @pllim You are welcome!

mcara: ATTN: @jhunkeler

conftest.py
```diff
@@ -1,10 +1,12 @@
...
-import os
-import pytest
-
-from astropy.tests.plugins.display import PYTEST_HEADER_MODULES
 from astropy.tests.helper import enable_deprecations_as_exceptions
 
-
 # Uncomment the following line to treat all DeprecationWarnings as exceptions
 enable_deprecations_as_exceptions()
+
```
pllim: Two blanklines before func def.
```suggestion


```

drizzlepac/alignimages.py
```diff
@@ -202,361 +197,480 @@ def perform_align(input_list, **kwargs):
...
+                filtered_table[:]['processMsg'] = "No sources found"
+                current_dt = datetime.datetime.now()
+                delta_dt = (current_dt - starting_dt).total_seconds()
+                log.info('Processing time of [STEP 4]: {} sec'.format(delta_dt))
+                return
+
+            # The catalog of observable sources must have at least MIN_OBSERVABLE_THRESHOLD entries to be useful
+            total_num_sources = 0
+            for chipnum in table.keys():
+                total_num_sources += len(table[chipnum])
```
mcara: ```total_num_sources = sum(map(len, table.values()))``` - this replaces the loop above. I may have commented on this before in a previous PR.

pllim: `table.values()` does not work. Table object is not a dictionary. Is this table storing *iterables* in a single cell?!

pllim: I really don't understand what this code is doing. Even for a masked table, `len(table[colname])` returns the total number of rows, not just the unmasked one.

pllim: Ah, okay... I get it now. You named a dictionary a `table`. I am just gonna step away now.

setup.py
```diff
@@ -92,8 +93,8 @@
         'stsci.imagestats',
         'stsci.skypac',
         'stsci.stimage',
-        'stwcs',
-        'tweakwcs>=0.4.0',
+        'stwcs>=1.5.0',
+        'tweakwcs>=0.4.6',
```
mcara: why not 0.5.0?

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -487,17 +457,17 @@ def extract_sources(img, **pars):
...
-        tbl.write(output, format='ascii.commented_header')
-        log.info("Wrote source catalog: {}".format(output))
+        if not outroot.endswith('.cat'):
+            outroot += '.cat'
+        tbl.write(outroot, format='ascii.commented_header')
+        log.info("Wrote source catalog: {}".format(outroot))
 
     if plot and plt is not None:
-        norm = None
+        norm = len(segm.labels)
```
mcara:
```
if vmax is None:
    norm = ImageNormalize(stretch=SqrtStretch())
else:
    norm = len(segm.labels)

File: drizzlepac/hlautils/astrometric_utils.py [hunk 9]
```diff
@@ -594,21 +554,24 @@ def generate_source_catalog(image, **kwargs):
...
+    # remove parameters that are not needed by subsequent functions
+    del detector_pars['fwhmpsf']
+
     # Build source catalog for entire image
     source_cats = {}
     numSci = countExtn(image, extname='SCI')
+    outroot = None
 
     for chip in range(numSci):
         chip += 1
```
mcara: may just adjust the range to `range(1, numSci + 1)`.

setup.py
jhunkeler:
```
Searching for stwcs>=1.5.0
Reading https://pypi.org/simple/stwcs/
No local packages or working download links found for stwcs>=1.5.0
error: Could not find suitable distribution for Requirement.parse('stwcs>=1.5.0')
script returned exit code 1
```

pllim: That is because stwcs stable release is only at 1.4.2. Anything after are only either alpha or rc and never pushed to PyPI.

stsci-hack: Regarding the tweakwcs version pin: changes made in 0.4.6 addressed problems that affected the operation of the alignment code, whereas the bug fixes that followed may make the code a little more accurate, it did not affect whether the code would run or not.

As for STWCS: there are release candidate tags on master which satisfy this requirement according to github.  Unfortunately, our code absolutely requires these newer versions.  So, they have been turned into pip installable packages yet just means that our code will have to wait to be delivered until stwcs (and photutils, for that matter) have been 'released' and turned into pip-installable products.

pllim: For `pip install` to work properly, `stwcs` release will have to be stable, RC won't cut it unless you want to do `pip install --pre`.

mcara: > just means that our code will have to wait to be delivered until stwcs (and photutils, for that matter) have been 'released'

@jhunkeler This complicates release of bug fixes for `drizzlepac` that users reported and are waiting for a release. Would it be possible to make a `drizzlepac` release without HLA pipeline?

pllim: > release without HLA pipeline?

Without HLA = no. With broken HLA = yes.

mcara: huh?

pllim: If I understand this diff correctly, HLA is already in drizzlepac `master`. You can make a release without merging this, yes. But it does not mean releasing drizzlepac *without* HLA, technically.

mcara: I am confident in @jhunkeler supernatural abilities.

jhunkeler: To release without HLA we'd need to `git revert` all of the HLA commits on `master`. 

So is it possible? Yes.
Will it look ugly? Absolutely.
Could there be complications? Probably.

pllim: Theoretically... You could branch off the repo at an older hash before first HLA merge, then re-apply any drizzlepac-only bug fixes as backports, and tag a release from *that* branch... But yeah, Joe is correct in this Three Answers ™️ .

pllim: Another option is to make `stwcs` an *optional* dependency. That is, everything would still install, but certain parts requiring `stwcs` would throw an error if it is not installed or too old. Joe hates this because it complicates the delivery process.

mcara: I interpret @jhunkeler Three Answers&trade; somewhat differently:

- _"Will it look ugly? Absolutely."_ Relative to some idealized situation? Maybe... Compared to the current state? I doubt.
- _"Could there be complications? Probably."_ Hopefully not or negligible complications. 🤞

stsci-hack: After discussing this with Nadia, the real solution to this issue would be to establish a requirements file where this version dependency is specified for stwcs, stsci.tools, and all other critical packages, while removing the ('offending', premature) changes to the setup.py file.

stsci-hack: The only thing holding up the release of everything is the configuration of the package for installation and testing.  I would hope it would actually take more effort to separate out the changes for the non-HLA-related changes than it would take to fix the configuration issues.  In addition, There are additional issues raised by the testing that has been done which need to be addressed before the non-HLA-related changes can be release anyway.  As a result, I would see it as counter-productive to separate the sets of changes, and instead continue working toward releasing all the code now.

jhunkeler:
Just a reminder: The :heavy_check_mark: next to the Jenkins run means nothing. `test_apriori` is failing.

To save time:

```sh
$ cd drizzlepac
$ git remote add jhunk https://github.com/jhunkeler/drizzlepac
$ git fetch jhunk
$ git checkout jhunk/integrate-hla
$ python setup.py develop
$ export TEST_BIGDATA=https://bytesalad.stsci.edu
$ pytest --bigdata --basetemp=test_results -v tests/hla/test_apriori.py
```

stsci-hack:
The commands you provided to run 'test_apriori.py' work perfectly for me, with all tests passing.  It would be helpful to see the log of a failing test with this version of the code.

jhunkeler:
> The commands you provided to run 'test_apriori.py' work perfectly for me, with all tests passing. It would be helpful to see the log of a failing test with this version of the code.

https://ssbjenkins.stsci.edu/blue/rest/organizations/jenkins/pipelines/STScI/pipelines/drizzlepac/branches/PR-349/runs/9/nodes/22/steps/139/log/?start=0

stsci-hack:
The apriori tests are failing due to an incorrectly configured python environment.  Our code requires versions of critical packages, like stwcs and photutils, which are only available from master yet, and therefore have not been released yet.  The errors I saw all relate to our latest code now being incompatible with the old version of photutils, to be specific.  This should be resolved with the inclusion of the requirements-dev.txt file, included in 38af58b, for use with pip when setting up the test environment.

jhunkeler:
Oops, I didn't notice it earlier but `Jenkinsfile` wasn't modified only `JenkinsfileRT` was. I changed things in the `Jenkinsfile` a minute ago, so we'll see what happens.

jhunkeler:
```  Running setup.py develop for drizzlepac

    ERROR: Error [Errno 22] Invalid argument: '/opt/conda/envs/tmp_env0/bin/python' while executing command /opt/conda/envs/tmp_env0/bin/python -c 'import setuptools, tokenize;__file__='"'"'/home/jenkins/workspace/STScI_drizzlepac_PR-349/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps

ERROR: Could not install packages due to an EnvironmentError: [Errno 22] Invalid argument: '/opt/conda/envs/tmp_env0/bin/python'


script returned exit code 1```

@rendinam  The Jenkins bug we've been battling for two weeks just reared its ugly head. Fantastic.

PR #360, opened 2019-05-04
Refactor HLA pipeline code and some bug fixes

mcara:
This PR includes a major code review of `alignimages.py` script and the rest of the HLA code. Along the way I fixed a couple of minor bugs and enhanced the code in some places to be more resilient.

I will provide a little more details in a later comment.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -36,8 +36,8 @@ from drizzlepac.hlautils import get_git_rev_info
 
 __taskname__ = 'alignimages'
 
-__version__ = 0.1
-__version_date__ = '15-Feb-2019'
+__version__ = 0.2
+__version_date__ = '04-May-2019'
 
 
 _MIN_CATALOG_THRESHOLD = 3
@@ -77,38 +77,45 @@ log = logutil.create_logger('alignimages', level=logutil.logging.INFO,
                             stream=sys.stdout)
 
 
-def check_and_get_data(input_list, **pars):
+def check_and_get_data(files, **pars):
     """ Verify that all specified files are present. If not, retrieve them
     from MAST.
 
     Parameters
     ----------
-    input_list : list
+    files: list
         List of one or more calibrated fits images that will be used for
         catalog generation.
 
     Returns
-    =======
-    total_input_list: list
+    -------
+    globbed_files: list
         list of full filenames
 
     """
+<<<<<<< HEAD
     retrieve_list = []  # Files retrieved via astroquery and resident on disk
     candidate_list = []  # File names gathered from *_asn.fits file
     ipppssoot_list = []  # ipppssoot names used to avoid duplicate downloads
     total_input_list = []  # Output full filename list of data on disk
     member_suffix = '_flc.fits'
+=======
+    retrieved_obs = []  # Files retrieved via astroquery and resident on disk
+    candidate_files = []  # File names gathered from *_asn.fits file
+    ipppssoot_files = []  # ipppssoot names used to avoid duplicate downloads
+    globbed_files = []  # Output full filename list of data on disk
+>>>>>>> Remove container type from var name
 
-    # Loop over the input_list to determine if the item in the input_list is a
+    # Loop over the `files` to determine if the item in the `files` is a
     # full association file (*_asn.fits), a full individual image file
     # (aka singleton, *_flt.fits), or a root name specification
     # (association or singleton, ipppssoot).
-    for input_item in input_list:
-        log.info('Input item: {}'.format(input_item))
-        indx = input_item.find('_')
+    for fname in files:
+        log.info('Input item: {}'.format(fname))
+        indx = fname.find('_')
 
         if indx != -1:  # Input with a suffix (_xxx.fits)
-            lc_input_item = input_item.lower()
+            lc_input_item = fname.lower()
             suffix = lc_input_item[indx + 1:indx + 4]
             log.info('file: {}'.format(lc_input_item))
             # For an association, need to open the table and read the image
@@ -118,9 +125,9 @@ def check_and_get_data(input_list, **pars):
             # just collects the wanted full file names.
             if suffix == 'asn':
                 try:
-                    asntab = Table.read(input_item, format='fits')
+                    asntab = Table.read(fname, format='fits')
                 except FileNotFoundError:
-                    log.error('File {} not found.'.format(input_item))
+                    log.error('File {} not found.'.format(fname))
                     return []
 
                 for row in asntab:
@@ -131,11 +138,11 @@ def check_and_get_data(input_list, **pars):
                     # or an ipppssoot
                     if memname.find('_') == -1:
                         memname += + '_flc.fits'
-                    candidate_list.append(memname)
+                    candidate_files.append(memname)
 
             elif suffix in ['flc', 'flt']:
-                if lc_input_item not in candidate_list:
-                    candidate_list.append(lc_input_item)
+                if lc_input_item not in candidate_files:
+                    candidate_files.append(lc_input_item)
 
             else:
                 log.error(
@@ -147,26 +154,25 @@ def check_and_get_data(input_list, **pars):
         # Input is an ipppssoot (association or singleton), nine characters by
         # definition. This "else" block actually downloads the data specified
         # as ipppssoot.
-        elif len(input_item) == 9:
+        elif len(fname) == 9:
             try:
-                if input_item not in ipppssoot_list:
+                if fname not in ipppssoot_files:
                     # An ipppssoot of an individual file which is part of an
                     # association cannot be retrieved from MAST
-                    retrieve_list = aqutils.retrieve_observation(input_item,
-                                                                 **pars)
+                    retrieved_obs = aqutils.retrieve_observation(fname, **pars)
 
                     # If the retrieved list is not empty, add filename(s) to
-                    # the total_input_list. Also, update the ipppssoot_list
+                    # the globbed_files. Also, update the ipppssoot_files
                     # so we do not try to download the data again. Need
-                    # to do this since retrieve_list can be empty because
+                    # to do this since retrieved_obs can be empty because
                     # (1) data cannot be acquired (error) or (2) data
                     # is already on disk (ok).
-                    if retrieve_list:
-                        total_input_list += retrieve_list
-                        ipppssoot_list.append(input_item)
+                    if retrieved_obs:
+                        globbed_files += retrieved_obs
+                        ipppssoot_files.append(fname)
                     else:
                         log.error('File {} cannot be retrieved from MAST.'
-                                  .format(input_item))
+                                  .format(fname))
                         return []
 
             except Exception:
@@ -174,21 +180,21 @@ def check_and_get_data(input_list, **pars):
                 traceback.print_exception(exc_type, exc_value, exc_tb,
                                           file=sys.stdout)
 
-    # Only the retrieve_list files via astroquery have been put into the
-    # total_input_list thus far. Now check candidate_list to detect or
+    # Only the retrieved_obs files via astroquery have been put into the
+    # globbed_files thus far. Now check candidate_files to detect or
     # acquire the requested files from MAST via astroquery.
-    for file in candidate_list:
-        # If the file is found on disk, add it to the total_input_list
+    for file_name in candidate_files:
+        # If the file is found on disk, add it to the globbed_files
         # and continue
-        if glob.glob(file):
-            total_input_list.append(file)
+        if glob.glob(file_name):
+            globbed_files.append(file_name)
         else:
             log.error('File {} cannot be found on the local disk.'
-                      .format(file))
+                      .format(file_name))
             return []
 
-    log.info('TOTAL INPUT LIST: {}'.format(total_input_list))
-    return total_input_list
+    log.info('TOTAL INPUT LIST: {}'.format(globbed_files))
+    return globbed_files
 
 
 def _log_step(description, pad_char='-'):
@@ -197,7 +203,7 @@ def _log_step(description, pad_char='-'):
 
 
 @util.with_logging
-def perform_align(input_list, archive=False, clobber=False, debug=False,
+def perform_align(images, archive=False, clobber=False, debug=False,
                   update_hdr_wcs=False, result=None, runfile=None,
                   print_fit_parameters=True, print_git_info=False,
                   output=False, num_sources=250):
@@ -206,7 +212,7 @@ def perform_align(input_list, archive=False, clobber=False, debug=False,
 
     Parameters
     ----------
-    input_list : list, optional
+    images : list
         List of one or more IPPSSOOTs (rootnames) to align.
 
     archive : bool, optional
@@ -287,7 +293,7 @@ def perform_align(input_list, archive=False, clobber=False, debug=False,
             log.warning("WARNING: Unable to display Git repository revision "
                         "information.")
 
-    log.info(input_list)
+    log.info(images)
 
     def _log_run_time(start_time, step):
         current_time = datetime.datetime.now()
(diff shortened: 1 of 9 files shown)


mcara:
#### General Code Clean-Up Changes:

- PEP8 compliance changes (line length, spacing, variable names, etc.);
- Removed unnecessary temporary variables;
- Removed unnecessary `continue`, etc. statements;
- Removed unnecessary comments; corrected some comments;
- Corrected some documentation. Replaced generic `object` with specific object types;
- Occasionally to make code fit into 80 char boundary, I had to shorten variable names (i.e., `filtered_table`->`fitab`). This, generally, should not affect readability (IMO);
- Whenever possible, used context managers to open/close files;
- When appropriate, switched from standard loops to list/dict comprehension or functional programming;
- Got rid of numerous tests/comparisons of booleans to `True` or `False`, i.e., https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L614
- Replaced in numerous places unnecessarily complicated boolean computations such as https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L881-L883 with direct boolean constructions such as `nmatches_check  = num_xmatches > 4`
- Simplified mathematical computations, i.e., avoiding `sqrt(x)**2` as in https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L902-L903

#### Major changes:

- Significantly modified `perform_align()` in order to reduce nesting level of the code throughout that function:
   - merged two loops into a single one https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L443-L444
   - placed innermost `try-except` block https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L492 more tightly around the code that might fail instead of covering code that should not fail. Unfortunately outer `try-except` block is still covering most of the code.
   - reduced nesting of comparison operations, i.e., https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L520-L521 or https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L535-L536 or of combined loop-comparison constructions such as https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L591-L596
- Removed the `run_align()`. The only place this function was used was inside the `perform_align()` function https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L213-L215 which was function as a wrapper just to handle function return through parameters(!) There is no need here for `run_align()` to pass return results through input arguments.
- Significantly modified `interpret_fit_rms()` *code* (the functionality should stay the same) to handle the case when images (or `TPWCS`-derived objects) do not have a `group_id` assigned (i.e., when it is `None` meaning images are not grouped). This may not hava an effectg on the current code but makes it more robust and future proof.

#### Added the following "local" functions:

- `_log_step()` to log a string surrounded by `'-'`;
- `_log_run_time` to write run time to the log;
- `_compute_fit_sky_rms` - needed due to a more significant change in `interpret_fit_rms()`.

#### Bug Fixes:

- In https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L1069-L1070 most likely `sys` and `log` were swapped, i.e., `sys.error()->sys.exit()` and `log.exit()->log.error()`.
- In https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L886-L888 adding `float(item.meta['fit_info']['shift'][0])**2` twice is probably a mistake and second term should have been `float(item.meta['fit_info']['shift'][1])**2. Overall this computation was replaced with `numpy.linalg`.

mcara:
In addition:

- Start of the `range` can be omitted when it is zero as in https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L509
- Table values can be modified for the entire column - no need to iterate over each row as in https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L966-L967
- Renamed container variables such as `fit_info_dict` to simply `fit_info`; `input_list` to `images`, etc. Container type most often is easy to find out from documentation. It helps declutter code and keep it width within PEP8 requirements.

drizzlepac/alignimages.py
```diff
@@ -25,646 +25,676 @@
...
+                    catalog, algorithm.__name__
+                )
+                fit_info[kwd] = fit_status[
+                    next(iter(fit_status))
+                ]
+                fit_info[kwd]['fit_qual'] = fit_quality
+
+                # Figure out which fit solution to go with based on
+                # fit_quality value and maybe also total_rms
+                if fit_quality >= 5:
```
pllim: This can be re-written such that `fit_quality ...: pass` is unnecessary. Also, the way this is set up right now, `best_fit_qual` is ignored when `fit_quality >= 5`.

drizzlepac/alignimages.py
```diff
@@ -673,35 +703,40 @@ def match_relative_fit(imglist, reference_catalog):
...
 
-    # Set all the group_id values to be the same so the various images/chips will be aligned to the astrometric
-    # reference catalog as an ensemble.
-    # astrometric reference catalog as an ensemble. BEWARE: If additional iterations of solutions are to be
-    # done, the group_id values need to be restored.
+    # Set all the group_id values to be the same so the various images/chips
+    # will be aligned to the astrometric reference catalog as an ensemble.
+    # astrometric reference catalog as an ensemble.
+    #
+    # BEWARE: If additional iterations of solutions are to be done,
```
pllim: If user really have to beware here, the message should go into `log.warning`, no?

drizzlepac/alignimages.py
```diff
@@ -1036,96 +1084,129 @@ def generate_source_catalogs(imglist, **pars):
...
+        source detection.
 
     Returns
     -------
-    sourcecatalogdict : dictionary
-        a dictionary (keyed by image name) of two-element dictionaries which contain the following:
-            * a dictionary of the detector-specific processing parameters
-            * an astropy table of position and photometry information of all detected sources
+    src_cat : dictionary
+        a dictionary (keyed by image name) of two-element dictionaries which
```
pllim:
```suggestion
        A dictionary (keyed by image name) of two-element dictionaries which
```

drizzlepac/alignimages.py
```diff
@@ -1036,96 +1084,129 @@ def generate_source_catalogs(imglist, **pars):
...
+        contain the following:
```
pllim:
```suggestion
        contains the following:
```

drizzlepac/alignimages.py
```diff
@@ -1036,96 +1084,129 @@ def generate_source_catalogs(imglist, **pars):
...
+                    # specific parameters
+                    detector_pars.update(pars)
+                    src_cat[imgname]["params"] = detector_pars
+
+                except KeyError:
+                    log.error(
+                        "ERROR! Unrecognized detector '{}'. Exiting..."
+                        .format(detector)
+                    )
+                    sys.exit(
```
pllim: I am not really sure about the need to kill the whole Python **session** here (or anywhere else in a function).

drizzlepac/alignimages.py
```diff
@@ -25,646 +25,676 @@
...
-                return (empty_list)
+                return []
 
-        # Input is an ipppssoot (association or singleton), nine characters by definition.
-        # This "else" block actually downloads the data specified as ipppssoot.
-        elif len(input_item) == 9:
+        # Input is an ipppssoot (association or singleton), nine characters by
+        # definition. This "else" block actually downloads the data specified
+        # as ipppssoot.
+        elif len(fname) == 9:
```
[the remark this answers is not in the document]
mcara: @pllim 
> Documentation formatting: There is a difference in the rendered HTML between var: type and var : type. You usually want the latter.

Yes, you are right. I forgot...

drizzlepac/alignimages.py
```diff
@@ -25,646 +25,676 @@
...
-                        filtered_table['fit_qual'][:] = fit_quality
-                        current_dt = datetime.datetime.now()
-                        delta_dt = (current_dt - starting_dt).total_seconds()
-                        log.info('Processing time of [STEP 5]: {} sec'.format(delta_dt))
-                        return (filtered_table)
+                    fitab['fit_qual'][:] = fit_quality
+                    continue
+
+                # determine the quality of the fit
+                fit_rms, fit_num, fit_quality, fitab, fit_status = \
```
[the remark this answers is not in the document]
mcara: It is kind of difficult to achieve the ideal situation here with such deep nesting and long variable names and trying to  stay within 80 col limit. What would be preferable: exceed 80 char or `\`? (I actually would like to know) Another alternative would be to rename results to something like `x, y, z, s, t = ...`. I can make it fit.

drizzlepac/alignimages.py
mcara: @pllim My understanding is that `best_fit_qual` is worst when 5 and best at 1. I was trying to avoid the situation on line 520-521 in the old file
```
if fit_quality < 5:
    if fit_quality == 1:
        ....
```
in order to decrease nesting level. If there is a better approach - let me know (or make a PR later).

drizzlepac/alignimages.py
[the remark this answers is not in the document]
pllim: Usually... If you are 3 levels deep or deeper, it is time to make it a separate function, but I digress. If you feel like you really need the `\`, that is fine too. It was just a passing comment.

pllim:
@jhunkeler or @rendinam , any ideas why `conda` isn't found in Jenkins?

```
+ which conda
which: no conda in (/opt/toolchain/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
```

https://ssbjenkins.stsci.edu/blue/organizations/jenkins/STScI%2Fdrizzlepac/detail/PR-360/11/pipeline/23

jhunkeler:
@pllim Its using a docker image without conda already installed... for a 100% not-conda experience. The images are based on the DSMO python builds.

`xx.nodetype` = (`python3.6` | `python3.7` | ...)

pllim:
@jhunkeler , then I don't know why it is failing. It keeps doing `which conda` for no reason? Also this:

```
No conda env dump stash available for mtx-0_py36_npGE116L117_apGE31L32
No conda env dump stash available for sdist
Unstashing test report for: mtx-0_py36_npGE116L117_apGE31L32
+ grep 'testsuite errors' results.mtx-0_py36_npGE116L117_apGE31L32.xml
Unstashing test report for: sdist
No results imported.
Post-build stage completed.
```

So why is there a red "x"?

mdlpstsci:
> #### General Code Clean-Up Changes:

>     * Removed the `run_align()`. The only place this function was used was inside the `perform_align()` function https://github.com/spacetelescope/drizzlepac/blob/65f4f738b0944322c8b139f4c75623b097a89005/drizzlepac/alignimages.py#L213-L215
>        which was function as a wrapper just to handle function return through parameters(!) There is no need here for `run_align()` to pass return results through input arguments.

Actually, there was a very specific reason for this to exist, but perhaps you will propose a better solution as your current change breaks the code.  As a Python novice I readily admit to not understanding how decorators work.  We found the @util.with_logging decorator on a function caused any returned value to be "swallowed" into oblivion.  In our case, the filtered table needs to be returned for diagnosis and analysis whether the `perform_align(()` is being invoked via `runastrodriz.py`, by a `pytest`, or when manually testing on the command line.  After some discussion, it was decided to create `run_align()` so the table could be passed as a parameter from the `perform_align()` function to`run_align()`.  The `perform_align()` then returns the table as output.  This was done (1) because it seemed awkward to have the invoking routine or really the command line invocation have to create an empty table first to pass to `perform_align()`, and (2) we did not want to alter the underlying function of `@util.with_logging` as this is used by many other pieces of code and a modification could have had an unexpected and unpleasant side effect.

drizzlepac/hlautils/analyze.py
```diff
@@ -96,179 +111,176 @@ def analyze_data(input_file_list, **kwargs):
...
         elif mtflag == 'T':
-            no_proc_key = MTKEY
+            # Moving target
+            no_proc_key = _MTKEY
             no_proc_value = mtflag
 
-        # Bostrophidon without or with dwell (WFC3 only)
         elif any([scan_typ == 'C', scan_typ == 'D']):
-            no_proc_key = SCNKEY
+            # Bostrophidon without or with dwell (WFC3 only)
```
mdlpstsci: `scan_typ` has not been initialized.

drizzlepac/alignimages.py
```diff
@@ -1134,188 +1215,208 @@ def update_image_wcs_info(tweakwcs_output):
...
-            fit_rms = group_dict[group_id]['FIT_RMS']
-            ra_rms = group_dict[group_id]['RMS_RA']
-            dec_rms = group_dict[group_id]['RMS_DEC']
         else:
-            fit_rms = None
-            ra_rms = None
-            dec_rms = None
+            # fit info is not complete when fit status is other than 'SUCCESS'
+            fit_rms, ra_rms, dec_rms = None
+
```
mdlpstsci: Having tripped on this myself... `Nonetype` is not iterable, so you cannot do this (unless you want the exception).

drizzlepac/alignimages.py
```diff
@@ -1134,188 +1215,208 @@ def update_image_wcs_info(tweakwcs_output):
...
 
-        item.meta['fit_info']['FIT_RMS'] = fit_rms
-        item.meta['fit_info']['TOTAL_RMS'] = total_rms
-        item.meta['fit_info']['NUM_FITS'] = len(group_ids)
-        item.meta['fit_info']['RMS_RA'] = ra_rms
-        item.meta['fit_info']['RMS_DEC'] = dec_rms
-        item.meta['fit_info']['catalog'] = reference_catalog.meta['catalog']
+        else:
+            # fit info is not complete when fit status is other than 'SUCCESS'
+            fit_rms, ra_rms, dec_rms = None
```
mdlpstsci: Same comment as for Line 1386.

drizzlepac/alignimages.py
mcara: Hm, interesting & continuing learning

mcara: Oh, I did not mean to put `,` - it should have been `=`. I cannot explain what was going on in my head. Will fix.

drizzlepac/alignimages.py
mcara: @mdlpstsci I fixed most except logging context manager issue. I need to investigate that one.

mcara:
@mdlpstsci Solved the `with_logging` issue in https://github.com/spacetelescope/drizzlepac/pull/364

setup.py
```diff
@@ -96,7 +96,7 @@
         'tweakwcs>=0.5.0',
         'stregion',
         'requests',
-        # HLA-pipeline specific:
+        # EPP-pipeline specific:
```
pllim: What is EPP?

mcara: @pllim https://github.com/spacetelescope/drizzlepac/pull/312#discussion_r273654042

mdlpstsci: Just an FYI as an alternative to EPP (maybe).  The new products being produced are being called the Hubble Advanced Products (HAP) in the system databases.

mcara: Whatever you or MO decide. Just let me know.

PR #366, opened 2019-05-08
Trap and avoid source identification problems

stsci-hack:
Some datasets result in the relative alignment between exposures being disrupted due to fits based on inaccurate source positions or other reasons.  These changes recognize when a fit has inconsistent RMS values for RA and DEC and flags that fit as invalid.  Other updates include:

- PEP8 updates for runastrodriz.
- Explicitly forcing 'clobber=False' in runastrodriz to avoid any potential for problems with astroquery
- updating the install requirements to include scikit-image to avoid numpy errors when running the alignment code with deblending turned on.
- Ignoring all sources in a segment if no identified source corresponds to the peak of the source.

These correct issues found during testing of HSTDP 2019.3a.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -39,7 +39,7 @@ MIN_OBSERVABLE_THRESHOLD = 10
 MIN_CROSS_MATCHES = 3
 MIN_FIT_MATCHES = 6
 MAX_FIT_RMS = 10  # RMS now in mas, 1.0
-MAX_FIT_LIMIT = 1000  # Maximum RMS that a result is useful
+MAX_FIT_LIMIT = 150  # Maximum RMS that a result is useful
 MAX_SOURCES_PER_CHIP = 250  # Maximum number of sources per chip to include in source catalog
 
 # Module-level dictionary contains instrument/detector-specific parameters used later on in the script.
@@ -382,7 +382,7 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
 
             # First ensure sources were found
 
-            if table is None or not table[1]: 
+            if table is None or not table[1]:
                 log.warning("No sources found in image {}".format(imgname))
                 filtered_table[:]['status'] = 1
                 filtered_table[:]['processMsg'] = "No sources found"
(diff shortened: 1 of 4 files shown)


drizzlepac/alignimages.py
```diff
@@ -900,7 +903,8 @@ def determine_fit_quality(imglist, filtered_table, catalogs_remaining, print_fit
         consistency_check = True
         rms_limit = max(item.meta['fit_info']['TOTAL_RMS'], 10.)
         if not math.sqrt(np.std(np.asarray(xshifts)) ** 2 + np.std(
-                         np.asarray(yshifts)) ** 2) <= (rms_limit / 1000.0) / (item.wcs.pscale):
+                         np.asarray(yshifts)) ** 2) <= (rms_limit / 1000.0) / (item.wcs.pscale) or \
+                         rms_ratio > 0.1:
             consistency_check = False
```
mdlpstsci: It would be good to declare what appear to be "magic numbers" (10, 1000, 0.1) to make the code understandable.

drizzlepac/runastrodriz.py
```diff
@@ -334,16 +334,17 @@ def process(inFile,force=False,newpath=None, inmemory=False, num_cores=None,
...
+            ftmp = open(_tmptrl, 'w')
             ftmp.writelines(_trlmsg)
             ftmp.close()
-            _appendTrlFile(_trlfile,_tmptrl)
+            _appendTrlFile(_trlfile, _tmptrl)
             _trlmsg = ""
 
             # Create an empty astropy table so it can be used as input/output for the perform_align function
-            #align_table = Table()
+            # align_table = Table()
```
mdlpstsci: We can get rid of Line 344 now as we do not want to mislead someone reading the code.  It was part of the experimenting with return values when the logging decorator caused the return value to be absorbed.

drizzlepac/alignimages.py
stsci-hack: Fix included in 2d3949e .

drizzlepac/runastrodriz.py
stsci-hack: Fix included in 2d3949e .

PR #371, opened 2019-05-09
Allow HLA astrometric code to work with photutils 0.6

larrybradley:
This PR makes a few changes to allow the HLA astrometric code to work with `photutils 0.6` (latest stable release).  Once `drizzlepac` changes its minimal dependency to `photutils >= 0.7` some of this code can be removed.

CC: @stsci-hack

### DIFF
--- a/drizzlepac/hlautils/astrometric_utils.py
+++ b/drizzlepac/hlautils/astrometric_utils.py
@@ -428,7 +428,7 @@ def extract_sources(img, dqmask=None, fwhm=3.0, threshold=None, source_box=7,
             # Define raw data from this slice
             detection_img = img[seg_slice]
             # zero out any pixels which do not have this segments label
-            detection_img[np.where(segm.data[seg_slice] == 0)] = 0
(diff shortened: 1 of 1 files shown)


drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -421,7 +443,7 @@ def extract_sources(img, dqmask=None, fwhm=3.0, threshold=None, source_box=7,
             # Define raw data from this slice
             detection_img = img[seg_slice]
             # zero out any pixels which do not have this segments label
-            detection_img[np.where(segm.data[seg_slice] == 0)] = 0
+            detection_img[segm.data[seg_slice] == 0] = 0
```
larrybradley: @stsci-hack I removed the `np.where()` call because it isn't needed here.

larrybradley:
@stsci-hack FYI, I just made one more quick fix to the code.

stsci-hack:
These changes were tested on a standard dataset with both photutils v0.6 and photutils v0.7dev and obtained the same fit as before.

PR #373, opened 2019-05-15
CRITICAL FIX for WCS update with best solution

stsci-hack:
This set of changes fixes a problem identified in the original code where the 'best fit' WCS is NOT being used to update the science data, where instead the last solution was being used.  This fix also removes the check for the ratio of the rms_RA and rms_DEC, since it is no longer needed with this fixing the underlying problem.  

The fix was to eliminate the use of 'best_meta' and replace it with a full copy of the 'best' imglist.

This needs to be included immediately in the pending, being tested, release.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -41,7 +41,7 @@ MIN_FIT_MATCHES = 6
 MAX_FIT_RMS = 10  # RMS now in mas, 1.0
 MAX_FIT_LIMIT = 150  # Maximum RMS that a result is useful
 MAX_SOURCES_PER_CHIP = 250  # Maximum number of sources per chip to include in source catalog
-MAX_RMS_RATIO = 1.0 # Maximum ratio between RMS in RA and DEC which still represents a valid fit
+# MAX_RMS_RATIO = 1.0  # Maximum ratio between RMS in RA and DEC which still represents a valid fit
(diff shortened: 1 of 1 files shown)


drizzlepac/alignimages.py
```diff
@@ -527,28 +524,33 @@ def run_align(input_list, archive=False, clobber=False, debug=False, update_hdr_
...
+
+                                        best_imglist = copy.deepcopy(imglist)
+
                                         best_fit_status_dict = fit_status_dict.copy()
+                                        best_fit_qual = fit_quality
                             else:  # new solution has worse fit_quality. discard and continue looping.
                                 continue
                             # preserve best fit solution so that it can be inserted into a reinitialized imglist next time through.
-                            temp_imglist = copy.deepcopy(imglist)
+                            best_imglist = copy.deepcopy(imglist)
```
mdlpstsci: I am not sure why line #553 is needed here.  The only case not covered by the if/elif/else is when fit_quality >= 5.  When fit_quality >= 5, the best_imglist is an empty list, and if this variable is then used later and is not checked for being empty (before it is filled at all), then this could cause a problem.

stsci-hack: It was originally intended to initialize the 'best' solution as needed for the now replaced/deleted `temp_imglist` list.  Additional testing demonstrated that it is not actually needed and has been removed in commit 842988d.

PR #381, opened 2019-05-24
Stop overwriting WCSNAME in drz products

stsci-hack:
This simple change insures that the WCSNAME header keyword gets populated based on the input WCSNAME values.  It also works in concert with [FITSBLENDER PR 25](https://github.com/spacetelescope/fitsblender/pull/25) to add WCSNAME to the HDRTAB extension of the drz products and has the WCSNAME keyword in the SCI header.


mcara:
@stsci-hack You may need to run regression tests and OKify them, please.

stsci-hack:
I will see what happens with the tests and update the 'truth' files accordingly.  Thanks!

PR #382, opened 2019-05-24
Increased search radius for alignment algorithms

stsci-hack:
A couple of problems arose with trying to align `J8C601040` and any observations like it; namely, the offset was larger than what our search algorithms were set up to find.  However, other problems stem from blindly trying too large of a search radius.  So, this version of the alignment code recognizes both a `blind_radius` as well as a tighter search `radius`.  The 'blind_radius' value will be used with only the default pipeline WCS has been defined, whereas the tighter `radius` will be used when the input WCS has been corrected by either an a priori or a posteriori solution.  The `blind_radius` has been increased to 250, while keeping the tighter 'radius' at 75 pixels as we used before.  

In addition, the parameters used for controlling the search are now dicts at the top of the module to allow for easier management, instead of being 'magic numbers' hard-coded in each function.

### DIFF
--- a/drizzlepac/alignimages.py
+++ b/drizzlepac/alignimages.py
@@ -61,8 +61,10 @@ detector_specific_params = {"acs": {"hrc": {"fwhmpsf": 0.152,  # 0.073
                                               "classify": True,
                                               "threshold": None}}}
 
-SEARCH_RELATIVE = {'radius': 250, 'tolerance': 2.0, 'separation':0.1, 'use2dhist':True}
-SEARCH_2DHIST = {'radius': 250, 'tolerance': 2.0, 'separation':0.1, 'use2dhist':True}
+SEARCH_RELATIVE = {'radius':75, 'blind_radius': 250, 'tolerance': 2.0, 
+                   'separation':0.1, 'use2dhist':True}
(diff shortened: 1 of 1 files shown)


drizzlepac/alignimages.py
```diff
@@ -896,7 +930,7 @@ def determine_fit_quality(imglist, filtered_table, catalogs_remaining, print_fit
 
         # Execute checks
         nmatches_check = False
-        if num_xmatches > 4:
+        if num_xmatches > 5:
```
mdlpstsci: Please change the hard-coded "5" to a module level constant with a descriptive name and ensure the MIN_CROSS_MATCHES reconciles with this new constant.  I would also ask here if MIN_CROSS_MATCHES would work for both checks (line 924) and line 933 (if num_xmatches >= MIN_CROSS_MATCHES)?

stsci-hack:
The latest version of pipeline processing reverts to apriori solutions for these datasets, so this work is not needed for pipeline use.  It serves as an interesting investigation into one way to try to make the alignment process more robust, but an updated version of histogram2d from the HLA would be easier and proven.  Therefore, this PR will not be merged and will be closed.

PR #394, opened 2019-06-06
[WIP] Generation of SExtractor-like source generation catalogs

mdlpstsci:
This code generates the source generation catalogs for the `total` (white light) detection products, as well as the `filter`-specific catalogs for each `total` product.


drizzlepac/hlautils/se_source_generation.py
```diff
@@ -0,0 +1,455 @@
...
+        A background map based upon the `~photutils.background.SExtractorBackground`
+        estimator
+
+    bkg_rms_mean : float
+        Mean bkg.background FIX
+
+    """
+
+    # Open the "white light" image and get the image data
+    imghdu = fits.open(source_filename)
```
stsci-hack: It may be helpful to support both a filename or an already opened FITS HDUList object as input here.  That would allow this code to be called with an in-memory copy of the images instead of forcing the calling code to write it to disk before calling this code.  This is NOT something that needs to be done now, but rather is a refinement for future development.

PR #400, opened 2019-06-12
Updates to the SExtractor-like catalog generation code.

mdlpstsci:
This is an update to the source code which generates the SExtractor-like catalogs for HAP.  This code runs through to completion via test mode (run_photutils()).  Please note there are known issues which need to be addressed which will be done on a different branch in a different PR. Some issues: compute circular aperture photometry for CI, fix table metadata format, resolve why some column descriptions do not appear in the output tables, add number of sources to table metadata, and add a source quality column to the table.


drizzlepac/hlautils/se_source_generation.py
```diff
@@ -77,17 +77,20 @@ def create_sextractor_like_sourcelists(source_filename, catalog_filename, param_
 
     """
 
-    # Open the "white light" image and get the image data
+    # Open the "white light" image and get the SCI image data
     imghdu = fits.open(source_filename)
     imgarr = imghdu[1].data
```
stsci-hack: It would be more general to use '[('sci',1)]' instead of '[1]'.  In addition, if 'build' gets turned off, the while light image would be a simple FITS image, so adding logic to look in the PRIMARY header if 'sci,1' can not be found would guard against that situation.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
         seg_table["RA_icrs"].info.format = ".10f"
         seg_table["Dec_icrs"].info.format = ".10f"
-        seg_table["RA_icrs"].unit = u.deg
-        seg_table["Dec_icrs"].unit =  u.deg
         print("seg_table (filter): {}".format(seg_table))
 
         seg_table.write(catalog_filename, format="ascii.ecsv")
         log.info("Wrote filter source catalog: {}".format(catalog_filename))
 
+def get_header_data(imghdu):
```
stsci-hack: This should either be made private (add '_' to the beginning of the function name) or a docstring should be added to explain what this does and to describe the inputs/outputs.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
+
+    keyword_dict = {}
+
```
stsci-hack: General comment for this entire function: using the exact metadata terms for the dictionary keys would make applying this dict to the metadata almost trivial for those keywords that have a 1-to-1 relation with the FITS keywords.  For example, 'Proposal ID' instead of 'proposal_id'.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
+    keyword_dict["wcs_name"] = imghdu[1].header["WCSNAME"]
+    keyword_dict["orientation"] = imghdu[1].header["ORIENTAT"]
+    keyword_dict["aperture_ra"] = imghdu[1].header["RA_APER"]
+    keyword_dict["aperture_dec"] = imghdu[1].header["DEC_APER"]
+    keyword_dict["aperture_pa"] = imghdu[1].header[" "]
+
+    return(keyword_dict)
+
+def annotate_table(input_table, keyword_dict, product="tdp"):
+
```
stsci-hack: If the metadata entry is used as the key for the dict, you should be able to apply those keywords in just a couple of lines with a simple loop over those keys, like "for key,value in dict.items(): if key not in key_list: input_table.meta[key] = value"
Some keys need additional processing, so "key_list" would be a list of keys that need extra work to get it into the metadata.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
+    if product.lower() == "tdp":
+        input_table.meta["Exposure Start"] = 0.000000 
+        input_table.meta["Exposure Time"] = 0.000
+    else:
+        input_table.meta["Exposure Start"] = keyword_dict["expo_start"]
+        input_table.meta["Exposure Time"] = keyword_dict["expo_time"]
+    input_table.meta["CCD Gain"] = keyword_dict["ccd_gain"]
+
+    if product.lower() == "tdp":
+        input_table.meta["Filter"] = keyword_dict["filter_tdp"]
```
stsci-hack: Is there a functional reason why this could not be included in the if...else statement above?

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
+    input_table.meta["Instrument"] = keyword_dict["instrument"]
+    input_table.meta["Detector"] = keyword_dict["detector"]
+    input_table.meta["Target RA"] = keyword_dict["target_ra"]
+    input_table.meta["Target DEC"] = keyword_dict["target_dec"]
+    input_table.meta["Orientation"] = keyword_dict["orientation"]
+    input_table.meta["Aperture RA"] = keyword_dict["aperture_ra"]
+    input_table.meta["Aperture DEC"] = keyword_dict["aperture_dec"]
+    input_table.meta["Aperture PA"] = keyword_dict["aperture_pa"]
+    if product.lower() == "tdp":
+        input_table.meta["Exposure Start"] = 0.000000
```
stsci-hack: The exposure start time and exposure time for the total product can be obtained from the drizzle product header as a correctly computed values (expstart and texptime, respectively).  I am not sure why they are set to 0.0 here.  The total exposure time can be useful in interpreting the flux values.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
+    keyword_dict["proposal_id"] = imghdu[0].header["PROPOSID"]
+    keyword_dict["image_file_name"] = imghdu[0].header['FILENAME']
+    keyword_dict["target_name"] = imghdu[0].header["TARGNAME"] 
+    keyword_dict["date_obs"] = imghdu[0].header["DATE-OBS"]
+    keyword_dict["instrument"] = imghdu[0].header["INSTRUME"]
+    keyword_dict["detector"] = imghdu[0].header["DETECTOR"]
+    keyword_dict["target_ra"] = imghdu[0].header["RA_TARG"]
+    keyword_dict["target_dec"] = imghdu[0].header["DEC_TARG"]
+    keyword_dict["expo_start"] = imghdu[0].header["EXPSTART"]
+    keyword_dict["expo_time"] = imghdu[0].header["EXPTIME"]
```
stsci-hack: I would suggest using 'texptime' here instead, since it was defined as the 'total exposure time' for the product.  In most cases, it has the same value as 'exptime', but not always.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
+    keyword_dict["ccd_gain"] = imghdu[0].header["CCDGAIN"]
+    keyword_dict["filter_tdp"] = imghdu[0].header["FILTER"]
+    keyword_dict["filter1"] = imghdu[0].header["FILTER1"]
```
stsci-hack: You will want logic to guard against using this on WFC3 data which does NOT have a FILTER1 or FILTER2 keyword which would cause this line to throw an exception (or at least I would expect it to).

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
...
+    keyword_dict["filter2"] = imghdu[0].header["FILTER2"]
+
+    # Get the HSTWCS object from the first extension
+    keyword_dict["wcs_name"] = imghdu[1].header["WCSNAME"]
```
stsci-hack: It may be helpful to also include the brand-new 'wcstype' keyword here as well as a human-readable interpretation of 'wcsname'.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -315,14 +330,14 @@ def compute_background(image, box_size=50, win_size=3, nsigma=5., threshold=None
...
             bkg_rms_mean = bkg.background.mean() + nsigma * bkg_rms.std()
             bkg_dao_rms = bkg.background_rms
-            if threshold is None:
+            if threshold_flag is None:
                 threshold = default_threshold
-            elif threshold < 0:
-                threshold = -1 * threshold * default_threshold
+            elif threshold_flag < 0:
+                threshold = -1 * threshold_flag * default_threshold
                 log.info("{} based on {}".format(threshold.max(), default_threshold.max()))
```
stsci-hack: This print statement should have been revised weeks ago in the original code to say something like "background threshold set to {} based on {}".

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -178,7 +183,7 @@ def create_sextractor_like_sourcelists(source_filename, catalog_filename, param_
     # Regenerate the source catalog with presumably now only good sources
     seg_cat = source_properties(imgarr_bkgsub, segm, filter_kernel=kernel, wcs=imgwcs)
```
stsci-hack: From the photutils docs:  If you input the background image that was subtracted from the data into the background keyword of source_properties(), the background properties for each source will also be calculated.  This would be done by adding 'background=bkg.background, if I interpret everything correctly.

drizzlepac/hlautils/se_source_generation.py
```diff
@@ -387,46 +412,119 @@ def write_catalog(seg_cat, catalog_filename, product="tdp"):
         dd = Column(dec_icrs, name="Dec_icrs")
```
stsci-hack: From the astropy docs, it appears that you should be able to define the 'description', 'units' and even 'format' when you define Column and have those values get added to the output table metadata automatically.  This may eliminate the need to add the .description and .unit lines for these columns from below... if it actually works.

drizzlepac/hlautils/se_source_generation.py
mdlpstsci: I wanted to keep the information in the same order as the classic HLA output.  Until I determine how to format the metadata, if possible, this was the brute force way of maintaining the order.

drizzlepac/hlautils/se_source_generation.py
mdlpstsci: I can see the values in the white light, total filter, and individual filter FITS headers.  I put the EXPSTART and EXPTIME (or TEXPTIME)  into the total filter detection product catalog to match the example I pulled from MAST.  The "white light" catalog has these values set to 0.0 for some reason. I agree that each TDP should have the values and not zero.  I will fix.

drizzlepac/hlautils/se_source_generation.py
mdlpstsci: Yes, I keep forgetting about this.

drizzlepac/hlautils/se_source_generation.py
mdlpstsci: OK

drizzlepac/hlautils/se_source_generation.py
mdlpstsci: I will combine the various "if" statements as I have been too literal about producing the same product as HLA Classic.

mdlpstsci:
Updated code has been committed, but this is still a work in progress.  Two suggested changes have not yet been implemented, and there other improvements to be made.  This code runs and writes the total and filter detection product catalogs.  The two items not addressed here will be moved to a new issue.

PR #407, opened 2019-06-20
Refactoring of run_astrodrizzle subroutine, addition of segmap sourcelist generation to runhlaprocessing.py

dulude:
This commit has a version of runhlaprocessing.py that produces all required imagery and sourcelist products to minimum specifications.

Changelist:
- Major refactoring of runhlaprocessing.run_astrodrizzle(). Now produces images that are much closer to their HLA classic counterparts.
- Added call to processing_utils.py to update header info in all drizzled products to meet HAP specs
- Added segmap sourcelist generation functionality (See closed PR #402)


stsci-hack:
@dulude : One last detail.  A test run on 'j9es06.out' resulted in a pickle file remaining behind; namely, 'j9es06.pickle'.  This will need to be turned off as well before this code can be merged.

PR #408, opened 2019-06-20
[WIP] Add functionality for auto-determining kernel for source finding

stsci-hack:
Basic implementation of an algorithm which attempts to define the source finding kernel using the sources it finds in the image.  This initial attempt has not been optimized or made robust against some error conditions.  However, it should work for any type of image from any detector and provide a reasonable kernel that can be used to more accurately identify sources in that image while taking into account exposure-specific changes to the PSF due to breathing, focus changes, guiding errors, and so on.
For safety, should no suitable PSF be identified, it will provide the same default Gaussian2DKernel that has been used so far.


stsci-hack:
This computation has already been included in the PR #455 .

PR #412, opened 2019-06-27
Point-source source identification upgrade

dulude:
* Added code to better estimate an overall FWHM value based on PSFs of image sources
* Added code to better estimate DAOstarfinder input "threshold" value. 
* Upgraded code seems to preform much better than before, yielding much more sensible results for noisier images (ACS 10595_06) while still still generating sourcelists that are in line with what was previously produced for cleaner images (ACS 1065_01)

### DIFF
--- a/drizzlepac/hlautils/sourcelist_generation.py
+++ b/drizzlepac/hlautils/sourcelist_generation.py
@@ -64,14 +64,7 @@ def create_dao_like_coordlists(fitsfile,sourcelist_filename,param_dict,make_regi
     log.info("bkg sigma: {}".format(bkg_sigma))
 
     pd_fwhm = param_dict['dao']['TWEAK_FWHMPSF']/param_dict['astrodrizzle']['SCALE']
-    pd_thresh = param_dict['dao']['TWEAK_THRESHOLD']*bkg_sigma
     calc_thresh = bkgsig_sf * bkg_sigma
-
-    log.info("param_dict fwhm: {}".format(pd_fwhm))
-    log.info("calculated fwhm: {}\n".format("3.5"))
-    log.info("param_dict_thresh: {}".format(pd_thresh))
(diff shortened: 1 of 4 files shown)


drizzlepac/hlautils/sourcelist_generation.py
```diff
@@ -45,16 +52,77 @@ def create_dao_like_coordlists(fitsfile,sourcelist_filename,make_region_file=Fal
...
+        bkg_rms_mean = max(0.01, imgarr.min())
+        bkg_rms = bkg_rms_mean * 5
+
+
+
+    # Estimate FWHM from image sources
+    kernel = astrometric_utils.build_auto_kernel(image, wht_image, threshold=bkg_rms, fwhm=default_fwhm)
+    segm = detect_sources(image, detect_sources_thresh, npixels=param_dict["sourcex"]["source_box"], filter_kernel=kernel)
+    cat = source_properties(image, segm)
+    bad_srcs = np.where(astrometric_utils.classify_sources(cat) == 0)[0] + 1
```
stsci-hack: This actually removes what it thinks are cosmic-rays or single-pixel events from the catalog.  I am not sure whether this is something that conforms to what the HLA-classic source identification would do.  Just wondering...

dulude: Commenting out these two lines (117 and 118) doesn't have any impact the final coord lists produced when I ran some quick comparison tests, so I just removed those two lines in commit [94bb84d](https://github.com/spacetelescope/drizzlepac/pull/412/commits/94bb84dd98a6fface8ad10e9199512ab305e5321).

dulude:
This branch ([dulude:pointsource_sorucelist_upgrade](https://github.com/dulude/drizzlepac/tree/pointsource_sorucelist_upgrade)) is obsolete and doesn't not need to be merged. I am closing the PR without merging it

PR #416, opened 2019-06-27
Customize stsci-bot behavior

pllim:
@mcara , I think this will make the bot does what you want.


pllim:
Err... I don't think CI failures about missing Numpy is related...

rendinam:
A small change and the failures appear unrelated. Merging.

PR #421, opened 2019-07-10
Create manifest file

stsci-hack:
A requirement for pipeline operations of this code is that a manifest file be generated to report the filenames of all the products produced from the input file.  This manifest needs to include not only the drizzle products, but also any headerlet files and source catalogs created as well.


stsci-hack:
Test failures will be addressed in a separate PR.

PR #426, opened 2019-07-22
Add support for HSTWCS inputs to astrodrizzle

stsci-hack:
Overload the 'final_refimage' and 'single_refimage' parameters to allow HSTWCS or astropy.wcs.WCS objects as inputs through the Python (not TEAL) API.  

While I was at it, I also started cleaning up some PEP8 style issues in the old code (not all, but a good start).


drizzlepac/util.py
```diff
@@ -510,10 +510,14 @@ def verifyRefimage(refimage):
 
     # start by trying to see whether the code can even find the file
     if is_blank(refimage):
-        valid=True
+        valid = True
```
mcara: I would propose to simply `return True`

mcara: I would propose to simply `return True`

drizzlepac/util.py
```diff
@@ -510,10 +510,14 @@ def verifyRefimage(refimage):
...
         return valid
 
-    refroot,extroot = fileutil.parseFilename(refimage)
+    if isinstance(refimage, astropy.wcs.WCS):
+        valid = True
+        return valid
```
mcara: Same here: `return True` should be enough and for the next block `return False`

drizzlepac/wcs_functions.py
```diff
@@ -30,44 +29,43 @@
...
         self.rot = None
         self.scale = None
 
-    def checkWCS(self,obj,name):
+    def checkWCS(self, obj, name):
         try:
             assert isinstance(obj, pywcs.WCS)
         except AssertionError:
-            print(name +' object needs to be an instance or subclass of a PyWCS object.')
+            print(name + ' object needs to be an instance or subclass of a PyWCS object.')
```
mcara: Maybe it is time to say bye-bye to `PyWCS`?

CHANGELOG.rst
```diff
@@ -14,6 +14,12 @@ The following notes provide some details on what has been revised for each
 version in reverse chronological order (most recent version at the top
 of the list).
 
+3.1.0 (unreleased)
```
mcara: I am not sure what is going on here (currently github says it is a merge conflict), but both `3.1.0` and `3.0.2` are "unreleased". I think `3.0.2` should have been updated with the release date when it was being planned to be released but if it wasn't, then maybe it should be done now?

stsci-hack: The code was 'released' starting 15-July (or thereabouts), but the change log did not get updated to reflect that...until now.

drizzlepac/wcs_functions.py
stsci-hack: Good idea...done.

stsci-hack:
Merging based on approval by Mihai (prior to resolving the file conflict).

PR #427, opened 2019-07-26
OO refactor source code 

mdlpstsci:
Three new modules have been created which will eventually supersede existing modules, but in the interim allow for testing both the original implementation, as well as new new object-oriented implementation in the same context seamlessly.  The new modules are hapsequencer.py (not a class at this time, but an overall driver), hlautils/product.py (classes representing the total, filter, and exposure products), and hlautils/poller_utils.py (the poller interpreter which also generates the product objects).  This is still WIP in terms of reorganization of functionality, refinement of functionality, and validation of solutions.  This PR has been created to allow others on Burgundy team to access the code easily.


drizzlepac/hlautils/poller_utils.py
```diff
@@ -0,0 +1,315 @@
...
+            # Append filter object to the list of filter objects for this specific total detection product
+            tdp_obj.add_product(filt_obj)
+
+            # Add the FilterProduct to the list of FilterProducts
+            filt_list.append(filt_obj)
+
+        # Add the TotalProduct to the list of TotalProducts
+        tdp_list.append(tdp_obj)
+
+    # Just for debugging.
```
stsci-hack: Debugging statements should either be removed or set to 'log.debug' prior to merging.

drizzlepac/hlautils/product.py
```diff
@@ -0,0 +1,222 @@
...
+        pass
+
+    def drizzle_product(self, meta_wcs, configobj):
+        """
+        Create the drizzle-combined total image using the meta_wcs as the reference output
+        """
+        edp_filenames = [element.full_filename for element in self.edp_list]
+        astrodrizzle.AstroDrizzle(input=edp_filenames,
+                                  output=self.drizzle_filename,
+                                  final_refimage=meta_wcs,
```
stsci-hack: Remember to explicitly set 'final_outnx' and 'final_outny' based on the values from the meta_wcs definition.

drizzlepac/hlautils/poller_utils.py
mdlpstsci: Removed debug cruft.

mdlpstsci:
- WIP
  Object-oriented implementation of the Hubble Advanced Products (HAP) portion
  of the package: main sequencing routine and the mosaic products.  Refined
  and/or created product.py, hapsequencer.py, and poller_utils.py.
  The hapsequencer.py and poller_utils.py currently exist independently from the
  corresponding modules runhlaprocessing.py, and pipeline_poller_utils.py.
  The runhlaprocessing.py and pipeline_poller_utils.py are deprecated by the
  OO implementation.  In essence the TotalProduct objects are equivalent to
  the obs_dict_info.

  The module processing_utils.py has changed in two routines:
  refine_product_headers, and update_hdrtab in that the obs_info_dict
  parameter has been replace by a TotalProduct object.

  This update removes the debugging cruft, and has added docstrings though
  the code is still WIP.  However, the code produces the same results as the
  the non-OO code (regression test of complicated poller file ib4606.out).

  This check-in is to allow others on Burgundy Team to layer their new OO
  code (configuration management and catalog management) into the new
  framework.  Once all is working from end-to-end, the deprecated routines
  will be removed.

  Important WIP:
  - Replace temporary total reference image with use of meta_wcs.
  - Improve and complete the creation of the point and segmented catalogs.
  - Migrate the alignment code to OO.
~

drizzlepac/hlautils/product.py
mdlpstsci: Currently using a reference image instead of the meta_wcs.  However, I would like to use the meta_wcs instead.  This method has been updated to use the final_[outnx|outny].

mdlpstsci:
I updated this PR.  I cleaned up a bit more code (removed cruft, added corrected comment), and more importantly, I removed the runhlapipeline.py and pipeline_poller_utils.py from the repo.  Since I updated processing_utils.py because I changed the function signatures of two functions, the old runhlapipeline.py would not run to successful completion anyway.

PR #429, opened 2019-08-14
Correct singleton CTE processing in runastrodriz

stsci-hack:
Pipeline testing under HSTDP 2019.3b uncovered problems with how singleton observations which are CTE-corrected are processed that resulted in FLC and FLT versions of the calibrated data having different WCSs.  
The base cause for this came down to how runastrodriz tried to include the FLC file for processing for singletons, then subsequently added another list that also included the FLC filename for additional processing.  Combined with STWCS v1.5.1, this resulted in the FLC file WCS getting reset by stwcs.updatewcs while the FLT image did not.  

The changes here will resolve this (long-standing hidden) problem with runastrodriz, while also adding a fast, simple regression test to insure it is fixed and that it stays fixed.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -366,7 +366,7 @@ def process(inFile, force=False, newpath=None, inmemory=False, num_cores=None,
 
             # Write the perform_align log to the trailer file...(this will delete the _alignlog)
             shutil.copy(_alignlog, _alignlog_copy)
-            _appendTrlFile(_trlfile, _alignlog)
+            _appendTrlFile(_trlfile, _alignlog_copy)
 
             # Append messages from this calling routine post-perform_align
             ftmp = open(_tmptrl, 'w')
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -275,6 +275,7 @@ def process(inFile, force=False, newpath=None, inmemory=False, num_cores=None,
     _drizfile = _trlroot + '_pydriz'
     _drizlog = _drizfile + ".log"  # the '.log' gets added automatically by astrodrizzle
     _alignlog = _trlroot + '_align.log'
+    _alignlog_copy = _alignlog.replace('.log', '_copy.log')
     if dcorr == 'PERFORM':
         if '_asn.fits' not in inFilename:
             # Working with a singleton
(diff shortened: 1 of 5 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -305,10 +305,14 @@ def process(inFile, force=False, newpath=None, inmemory=False, num_cores=None,
             _cal_prodname = inFilename
             _new_asn.extend(_inlist)  # kept so we can delete it when finished
 
-        # check to see whether FLC files are also present, and need to be updated
-        # generate list of FLC files
+            # check to see whether FLC files are also present, and need to be updated
+            # generate list of FLC files
+            _calfiles_flc = [f.replace('_flt.fits', '_flc.fits')
+                             for f in _calfiles
+                             if os.path.exists(f.replace('_flt.fits', '_flc.fits'))]
```
mcara: For improved efficiency it is better to avoid replacing `flt`->`flc` _twice_. I think this could be done in the following way:
```python
_calfiles_flc = [x for x in [f.replace('_flt.fits', '_flc.fits') for f in _calfiles] if os.path.exists(x)]
```

mcara:
@stsci-hack However, it would be nice if tests were passing before this is merged.

stsci-hack:
The new regression test added with this PR now passes, and runs using real data pulled using astroquery.  The only failures now stem from work being done to update other packages and only impact the Windows builds.

drizzlepac/runastrodriz.py
stsci-hack: As we discussed, the nested list comprehension calls is elegant, but not necessarily any clearer or easier to maintain.  So, unless there is a compelling reason, the current implementation will remain for the sake of clarity in the code.

jhunkeler:
@stsci-hack Ok, we're past the error with `spherical_geometry`... but there's drizzlepac-specific problem now:

```python 
        download_dir = None
        for file, file_status in zip(manifest['Local Path'], manifest['Status']):
            if file_status != "LOCAL":
                # Identify what sub-directory was created by astroquery for the
                # download
                if download_dir is None:
                    file_path = file.split(os.sep)
>                   file_path.remove('.')
E                   ValueError: list.remove(x): x not in list

D:\a\1\s\drizzlepac\hlautils\astroquery_utils.py:116: ValueError
---------------------------- Captured stdout call -----------------------------
INFO: ['iaaua1n4q_flc.fits']
Downloading URL https://mast.stsci.edu/api/v0.1/Download/file?uri=mast:HST/product/iaaua1n4q_flc.fits to ./mastDownload/HST/iaaua1n4q/iaaua1n4q_flc.fits ... [Done]
```

PR #439, opened 2019-08-22
add config object instantiation

dulude:
Added some code to hapsequencer.py to instantiate configuration objects into each product object defined by hlautils.poller_utils.interpret_obset_input(). These configuration objects contain all necessary pipeline parameter settings that will be used later on in the hapsequencer run.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -343,7 +343,7 @@ def run_hla_processing(input_filename, result=None, debug=False):
             full_product_list += total_item.edp_list
 
         # Next, generate output parameter file
-        param_out_filename = input_filename.replace(".out", ".config")
+        param_out_filename = input_filename.replace(".out", "_config.json")
 
         # Finally, instantiate configuration object in each product object
         for item in full_product_list:
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -315,14 +315,15 @@ def create_drizzle_products(obs_info_dict, total_list, meta_wcs):
 # ----------------------------------------------------------------------------------------------------------------------
 
 
-def run_hla_processing(input_filename, result=None, debug=False):
+def run_hla_processing(input_filename, result=None, debug=False, use_defaults_configs=True,
+                       input_custom_pars_file=None, output_custom_pars_file=None):
     # This routine needs to return an exit code, return_value, for use by the calling
     # Condor/OWL workflow code: 0 (zero) for success, 1 for error condition
     return_value = 0
 
     starting_dt = datetime.datetime.now()
     log.info("Run start time: {}".format(str(starting_dt)))
-    product_list = []
+
     try:
         # Parse the poller file and generate the the obs_info_dict, as well as the total detection 
         # product lists which contain the ExposureProduct, FilterProduct, and TotalProduct objects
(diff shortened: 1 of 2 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -334,6 +335,23 @@ def run_hla_processing(input_filename, result=None, debug=False):
         # instrument_programID_obsetID_manifest.txt (e.g.,wfc3_b46_06_manifest.txt)
         manifest_name = total_list[0].manifest_name
 
+        # Set up all pipeline parameter values that will be used later on in the run.
+        # First, build temp.list of all product objects
+        full_product_list = total_list.copy()
```
stsci-hack: You should work directly on the total_list.  This will allow the newly added '.pars' attribute from line 350 to become available to the rest of the steps.  Each step can then use the '.pars.get_pars('<processing step>')' syntax to retrieve the parameters for the processing step on-the-fly without any additional file I/O or bookkeeping for cfg filenames.

drizzlepac/hapsequencer.py
```diff
@@ -334,6 +335,23 @@ def run_hla_processing(input_filename, result=None, debug=False):
...
+
+        # Next, generate output parameter file
+        param_out_filename = input_filename.replace(".out", "_config.json")
+
+        # Finally, instantiate configuration object in each product object
+        for item in full_product_list:
+            item.pars = config_utils.HapConfig(item, output_custom_pars_file=param_out_filename, use_defaults=True)
+
+        # Lastly, delete temp.list of all product objects 
+        del full_product_list
```
stsci-hack: To go along with the previous comment about line 340, we do not want to work on a copy of the product list, so there will be nothing to delete, so this line can be removed.

drizzlepac/hapsequencer.py
```diff
@@ -334,6 +335,23 @@ def run_hla_processing(input_filename, result=None, debug=False):
...
 
+        # Set up all pipeline parameter values that will be used later on in the run.
+        # First, build temp.list of all product objects
+        full_product_list = total_list.copy()
+        for total_item in total_list:
+            full_product_list += total_item.fdp_list
+            full_product_list += total_item.edp_list
+
+        # Next, generate output parameter file
+        param_out_filename = input_filename.replace(".out", "_config.json")
```
stsci-hack: This may end up only being used in 'debug' mode (however that gets defined) to insure the HapConfig class correctly generates the parameters for each step.

drizzlepac/hapsequencer.py
```diff
@@ -334,6 +335,23 @@ def run_hla_processing(input_filename, result=None, debug=False):
...
+
+        # Finally, instantiate configuration object in each product object
+        for item in full_product_list:
+            item.pars = config_utils.HapConfig(item, output_custom_pars_file=param_out_filename, use_defaults=True)
```
stsci-hack: The HAPProduct definition in hlautils/product.py should be updated to change "self.param_dict={}" to "self.pars=None".  This will document that the pars object will be attached to the product once it gets instantiated by this line, rather than magically adding an attribute out of nowhere (which Python happily allows a developer to do much to the confusion of anyone trying to maintain that code later...).

drizzlepac/hapsequencer.py
dulude: Resolved by commit [33fc25c](https://github.com/spacetelescope/drizzlepac/pull/439/commits/33fc25c59307379fe90008fea1e7de121c4f0c93)

drizzlepac/hapsequencer.py
dulude: Resolved by commit [33fc25c](https://github.com/spacetelescope/drizzlepac/pull/439/commits/33fc25c59307379fe90008fea1e7de121c4f0c93)

drizzlepac/hapsequencer.py
dulude: Resolved by commit [33fc25c](https://github.com/spacetelescope/drizzlepac/pull/439/commits/33fc25c59307379fe90008fea1e7de121c4f0c93)

drizzlepac/hapsequencer.py
dulude: Resolved by commit [0a74d37](https://github.com/spacetelescope/drizzlepac/pull/439/commits/0a74d37a7d37f5e0bb7a86f373d69ca0bbd9bca9)

PR #440, opened 2019-08-26
Changes to alignment code to make it more flexible [WIP] [skip ci]

stsci-hack:
The alignment code was originally developed as a monolithic functional black-box that did everything (very IRAF-like).  However well it supported the initial pipeline use case, it did not lend itself to supporting additional use cases.  These changes implement a new version of the code which is based on objects that is designed to be more flexible in how it can be called to support additional use cases with less trouble.


stsci-hack:
This PR will result in (at least) the following changes in API for code related to alignment in this package:
  - `align_table` in hapsequencer now an object with previous astropy table results as an attribute `.filtered_table`
  - added `compute_2d_background` function to astrometric_utils
  - added `good_fwhm` and `num_fwhm` parameters to `build_auto_kernel` to control iterations to find a suitable source 
  - added parameters to `extract_sources` to better control selection of sources
  - changed second parameter for `compute_photometry` to define default value
  - added several functions to `astrometric_utils` to compute image and focus verification
  - Updated `TotalProduct` in `product.py` to define reference drizzle product for verification purposes
  - updated `FilterProduct` in `product.py` to use new O-O defined alignment interfaces
  - added parameters to `FilterProduct.align_to_gaia` to record what catalog it should align to.
  - updated `astrodrizzle` to create `single_sci.fits` files even with `in_memory` turned on (needed for verification processing)
  - added code to `devutils` to generate multi-page PDF showing alignment results for new pipeline processing
  - Finally, replaced 'runastrodriz' as used in the pipeline with the new processing code developed as 'runaligndriz` while keeping the `runastrodriz` name.

PR #447, opened 2019-09-09
bifurcation of json configuration files

dulude:
- Split the "parameters" and "default_values" subsections of all the .json configuration files in hap_pars/*/... into individual files. "parameters" subsections were written to hap_pars/user_parameters, and "default_values" subsections were written to hap_pars/default_parameters. The subdirectory structure formerly used to organize and store the old unsplit .json configuration files in hap_pars/ was preserved in hap_pars/user_parameters and hap_pars/default_parameters.

- Removed hap_pars/any, hap_pars/acs and hap_pars/wfc3 and all the now unused unsplit .json configuration files contained therein 

- Refactored hlautils/config_utils.py to utilize "split" configuration .json files found in hap_pars/user_parameters and hap_pars/default_parameters.

- created drizzlepac/devutils/ which contains useful scripts that may be useful at some future point in hapsequencer.py development, but are expressly NOT intended to be part of a general release to the wider world.


drizzlepac/devutils/fix_json_files.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python
```
stsci-hack: Does this file need to be provided to the end-user?  This seems to be a purely developmental function...

drizzlepac/devutils/json_splitter.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
```
stsci-hack: Does this file need to be provided to the end-user?  If not, it should be removed from this PR.

PR #450, opened 2019-09-11
Add cmdline interface

stsci-hack:
This provides a basic command line interface for use by end users or a processing pipeline.  The command defined will be:
    `runsinglehap <poller file>`
where the "<poller file>" will be a database query file with one line for each exposure to be included in the single visit products.


drizzlepac/runsinglehap.py
```diff
@@ -0,0 +1,86 @@
...
+
+:License: :doc:`LICENSE`
+
+USAGE: runsinglehap [-d] inputFilename
+
+Python USAGE:
+    python
+    from drizzlepac import runsinglehap
+    runsinglehap.process(inputFilename,debug=True)
+
```
mdlpstsci: Just a typo here: "process" should be "perform".

drizzlepac/runsinglehap.py
```diff
@@ -0,0 +1,86 @@
...
+The '-d' option will run this task in DEBUG mode producing additional outputs.
+
```
mdlpstsci: I would move this line up to be beneath the command line USAGE.

mdlpstsci:
Just to small changes to docstring.

drizzlepac/runsinglehap.py
stsci-hack: Fixed (along with another typo in the doc strings) in c569ae17

drizzlepac/runsinglehap.py
stsci-hack: Resolved in c569ae17

PR #452, opened 2019-09-13
Initial implementation of similarity index

stsci-hack:
The new function 'compute_similarity()' computes a similarity index between 2 images.  Initial testing with this function on drizzled images shows that this function returns distinct values for similar vs. non-similar images.  This index is loosely derived from the algorithm used by [AmphIdent as described online](https://onlinelibrary.wiley.com/doi/full/10.1002/ece3.3140). 

This function has been tested on a couple of test cases so far, specifically:
  - an image drizzled at a scale 5% different from the reference image (same inputs) returns an index of 0.014
  - an image drizzled from mis-aligned inputs compared to a drizzled image of the same inputs with correct relative alignment (same set of images as previous example) returned an index of 2.734.

Ultimately, this function could be used to compare a drizzle product prior to _a posteriori_ alignment (as the reference) to a drizzle product generated after _a posteriori_ alignment to validate the alignment in an automated manner.

### DIFF
--- a/drizzlepac/hlautils/astrometric_utils.py
+++ b/drizzlepac/hlautils/astrometric_utils.py
@@ -23,6 +23,7 @@ from distutils.version import LooseVersion
 
 import numpy as np
 from scipy import ndimage
+import scipy.stats as st
 from lxml import etree
 try:
     from matplotlib import pyplot as plt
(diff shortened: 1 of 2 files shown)


stsci-hack:
A new version of 'runastrodriz', now named 'runaligndriz', has been added to implement the new logic for applying this similarity index to the products generated after performing a posteriori alignment to GAIA.  This new code will completely back out all WCS solutions generated during processing.  Whatever WCS is active for the products which fail the similarity check will be deleted from the input files, replaced with the pipeline default solution again (using updatewcs to be thorough), and the badly-aligned drizzle products will be deleted and replaced with the pipeline default products generated as the comparison files.  All this gets documented in the trailer files so the user gets a record (albeit mostly hidden) of the failure to align.   In addition, the code now also includes a new parameter to let the user keep a recognized 'bad' solution so that they can investigate what went wrong or confirm how bad the alignment went.

stsci-hack:
This verification step will be moved into PR #440 in order to take advantage of the improvements in alignment to avoid as many errors cases as possible.

PR #454, opened 2019-09-18
Fix rounding problem with creating meta_wcs

stsci-hack:
This fixes a rounding problem with determining the size of the output undistorted WCS for a given set of distorted WCSs.


mcara:
What was the problem?

PR #458, opened 2019-10-04
[WIP] Updates to the Segmentation catalog implementation

mdlpstsci:
WIP Modifications for the implementation of the Segmentation catalog

- Update implementation of the Segmentation catalog by adding
  the concentration index computation.
  - Perform some clean-up.
- Update to JSON "catalog generation" files:
  - Add entries (nlevels, contrast, border) to support.
    segmentation catalog (sourcex) to eliminate any hard-coded
    values in code.
  - Move the entries which support the computation of the
    concentration index to the catalog-independent portion
    of the file.
  - Remove "bthresh" and "thresh" entries from [sourcex].
- Add comments and log message to wcs_drizzle_product() method
  of Product classes to explain needed override of configuration
  values.


drizzlepac/hlautils/catalog_utils.py
```diff
@@ -317,8 +331,45 @@ def __init__(self, image, param_dict, debug, tp_sources):
...
+        photflam = self.image.imghdu[1].header['photflam']
+        # FIX: replace the constants
+        self.ab_zeropoint = -2.5 * np.log10(photflam) - 21.10 - 5.0 * np.log10(photplam) + 18.6921
+
+        # Compute average gain
+        # FIX not four amps for all detectors
+        self.gain = self.image.imghdu[0].header['exptime'] * np.mean([self.image.imghdu[0].header['atodgna'],
+                                                                      self.image.imghdu[0].header['atodgnb'],
+                                                                      self.image.imghdu[0].header['atodgnc'],
+                                                                      self.image.imghdu[0].header['atodgnd']])
```
stsci-hack: Use of "header['atodgn*']" would create a list of whatever values are present in the header and use of list comprehension can generate the final values from this list without explicitly looking for all 4 amps.  Logic could even be used with this to only include keyword values that are non-zero since observations which only use 1 amp at a time set all the other keyword values to 0.0, which would ruin the interpretation of the gain entirely.  See 'ibvz01010' as an example of where this is done (0.0, 0.0, 1.559999, 0.0).

drizzlepac/hlautils/catalog_utils.py
```diff
@@ -317,8 +331,45 @@ def __init__(self, image, param_dict, debug, tp_sources):
...
+
+        # Convert photometric aperture radii from arcsec to pixels
+        self.aper_radius_arcsec = [self.param_dict['aperture_1'], self.param_dict['aperture_2']]
+        self.aper_radius_list_pixels = []
+        for aper_radius in self.aper_radius_arcsec:
+            self.aper_radius_list_pixels.append(aper_radius/self.param_dict['scale'])
```
stsci-hack: self.aper_radius_list_pixels = [aper_radius/self.param_dict['scale'] for aper_radius in self.aper_radius_arcsec]

drizzlepac/hlautils/catalog_utils.py
```diff
@@ -625,173 +649,169 @@ def transform_list_xy_to_ra_dec(self, list_of_x, list_of_y, drizzled_image):
...
 
-        # if processing filter product, use sources identified by parent total drizzle product identify_sources() run
-        if self.tp_sources:
+            try:
+                # Deblending is a combination of multi-thresholding and watershed
+                # segmentation. Sextractor uses a multi-thresholding technique.
+                # npixels = number of connected pixels in source
+                # npixels and filter_kernel should match those used by detect_sources()
+                segm_deblended_img = deblend_sources(imgarr_bkgsub, self.segm_img, npixels=self._size_source_box,
+                                                     filter_kernel=self.image.kernel, nlevels=self._nlevels, contrast=self._contrast)
```
stsci-hack: It may be better to drop 'contrast=...' to the next line...

drizzlepac/hlautils/catalog_utils.py
```diff
@@ -625,173 +649,169 @@ def transform_list_xy_to_ra_dec(self, list_of_x, list_of_y, drizzled_image):
...
+
+            # Clean up segments that are near or partially overlap the border of the image and relabel
+            # the segments to be sequential
+            self.segm_img.remove_border_labels(self._border, partial_overlap=True, relabel=True)
+
+            # The total product catalog consists only of X/Y and RA/Dec coordinates for the detected sources
+            # in the total drizzled image.  All the actual measurements are done on the filtered drizzled
+            # images using the coordinates determined from the total drizzled image.
+            self.source_cat = source_properties(imgarr_bkgsub, self.segm_img, background=self.image.bkg_background_ra,
+                                                filter_kernel=self.image.kernel, wcs=self.image.imgwcs)
```
stsci-hack: This operation seems unnecessary, since the same operation gets performed in 'measure_sources'.  Is there a reason 'self.source_cat' needs to be defined at this stage of processing, or can it actually wait until 'measure_sources' is called?  If it can wait, it would remove some unnecessary operations.

drizzlepac/hlautils/catalog_utils.py
mdlpstsci: Done.

drizzlepac/hlautils/catalog_utils.py
mdlpstsci: This was done here to support the following debug block which generates data used for ds9 overlay.  I will move this code into the debug block for now.  In a follow on Issue/PR for "clean up" this code will move into the write method to keep all the file generation code in one place.

drizzlepac/hlautils/catalog_utils.py
mdlpstsci: All places were the source used the configuration scale factor have been upgraded to use the scaling factor found in the actual data.

drizzlepac/hlautils/catalog_utils.py
mdlpstsci: Replaced the hard-coded search and evaluation of the four gain header keywords by the above suggestion of searching for the keywords using wildcards and list comprehension!

drizzlepac/hlautils/catalog_utils.py
mdlpstsci: CORRECTION.  The self.source_cat=source_properties()...) does need to be done here in order to get the centroid coordinates for the sources in the white light image.  The "total" catalogs (*.ecsv file) are written before the source properties are derived for the filter images.

mdlpstsci:
Added reviewer comments:
- Replaced use of the scale factor obtained from the configuration files with the scale value stored in the image as part of the WCS.
- Reworked the computation of the gain factor to accommodate an unknown number of FITS header keywords ('atodgn?') which may or may not contain viable values.

PR #461, opened 2019-10-17
[WIP] Updates to creating catalogs:

mdlpstsci:
- Removed hard-coded configuration variables in hapsequencer.py.
- Removed "scale" from JSON configuration files.
- Added code to the TotalProduct class to use Footprints in
  order to generate a mask for the total image contour.
- Moved writing of the Total Detection catalog until after the
  measurements have been done on the Filtered images and the
  Filter catalogs have been written.  A few columns from the
  Filter catalog are extracted and appended to the Total Detection
  catalog.
  - Supporting extracting and appending: "extract" and "combine"
    routines added to the appropriate classes.
  - Column names from the filter catalogs have the filter name
    added as a suffix when appended to the total catalogs.
- Upgraded the annotation which is inserted as part of the
  "header" information in the *.ecsv file to contain an
  official STScI publication statement.
- Added further information for the interpretation of the
  tabular data to the header annotation.


mdlpstsci:
This is still a WIP, but it does run to completion and creates catalogs (point and segment) for an instrument with multiple detectors and multiple filters for a detector.  The catalogs require further clean up.

mdlpstsci:
Previous PRs have been merged with the same bug that is only revealed on Windows so it does not appear to be prohibitive.  Will consult with original developer to understand the original intent of the code.

PR #463, opened 2019-10-24
Attempt to fix issue on Windows for log file access

mdlpstsci:
Accessing (os.remove(drizlog)) the drizzle log file generates an error on Window systems.  Cleaned up bits of code handling various log files in an attempt to fix this issue.


mdlpstsci:
Closing this PR without merging as it did not resolve the issue, and the original developer should be consulted to understand the intent of the code.  Having said this, the original developer merged other PRs with this "bug" so it is not a showstopper for progress.

PR #477, opened 2019-11-05
New PC size of 4.2, SC size of about 0.2

stsci-hack:
Update to the sky cell definitions to reduce the projection cell size to 4.2 degrees (from 5.0) and make the sky cell size 0.2deg (from 0.1).
  
**This should not be merged until we have agreement on what the new values should be and that the file provides those correctly.**


stsci-hack:
Testing by Matt suggested that some exposures may not be entirely contained by the sky cells identified as the only overlapping cells.  Additional testing needs to be performed on those cases to insure that the sky cell definitions, and more importantly, the code used to identify overlaps works accurately.

PR #483, opened 2019-11-07
Revise logging

stsci-hack:
These changes implement a single file as the full log of all processing that takes place for a single visit through `runsinglehap`.  This logging relies entirely on Python's basic logging functionality, with standardized, Python-recommended setup being performed by `stsci.tools.logutil`.

In addition, I addressed some syntax (PEP8) issues and removed some extraneous import statements as well.

### DIFF
--- a/drizzlepac/updatehdr.py
+++ b/drizzlepac/updatehdr.py
@@ -6,8 +6,7 @@
 
 """
 import re
-import math
-import warnings
+import sys
 
 from astropy.io import fits
 import numpy as np
@@ -15,7 +14,7 @@ import numpy as np
 from astropy import wcs as pywcs
 
 from stsci.tools import fileutil, logutil
-from stwcs import wcsutil, updatewcs
+from stwcs import wcsutil
 from stwcs.wcsutil import wcscorr
 from stsci.skypac.utils import get_ext_list, ext2str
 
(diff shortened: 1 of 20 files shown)


stsci-hack:
The drizzle log files can remain separate files since they pertain specifically to just the product they are generating.  Besides, a trailer file is expected for each product as generated by drizzle, so we can leave those as extra trailer files.  My testing indicated that the total product log file naturally includes all the separate drizzle log comments as well.  I still need to verify that the same holds true for the hla_flag comments as well, in that they are automatically included in the total log file.  However, a separate log file does NOT need to be generated for that step, so I will see what is going on there and include the appropriate revision in this PR.

stsci-hack:
Latest commit turned off use of 'utils.with_logging' for hla_flag_filter to be consistent with the rest of the single hap processing.  The total trailer file now get generated with the proper name for all processing performed by runsinglehap.

drizzlepac/updatehdr.py
```diff
@@ -292,9 +289,9 @@ def linearize(wcsim, wcsima, wcsref, imcrpix, f, shift, hx=1.0, hy=1.0):
         wcsref.wcs_pix2world(p.astype(np.float64), 1), 1).astype(np.longdouble)
 
     # derivative with regard to x:
-    u1 = ((p[1] - p[4]) + 8 * (p[3] - p[2])) / (6*hx)
+    u1 = ((p[1] - p[4]) + 8 * (p[3] - p[2])) / (6 * hx)
     # derivative with regard to y:
-    u2 = ((p[5] - p[8]) + 8 * (p[7] - p[6])) / (6*hy)
+    u2 = ((p[5] - p[8]) + 8 * (p[7] - p[6])) / (6 * hy)
```
mdlpstsci: This can wait, but we should work to remove magic/hard-coded numbers and indices from the code.

PR #496, opened 2019-11-12
Update hla tests

mdlpstsci:
Created a parameterized test for processing poller files and generating Single Visit Mosaic output.


tests/hla/test_run_svmpoller.py
```diff
@@ -0,0 +1,141 @@
...
+
+    try:
+
+        # The dataset variable is a fully qualified poller file (CSV) with no header
+        # Copy the poller file to the current working directory
+        shutil.copy2(dataset, ".")
+
+        # Get the poller file path, as well as its simple name
+        poller_path = "/".join(dataset.split("/")[:-1])
+        poller_file = dataset.split("/")[-1]
```
stsci-hack: It is usually better to be a little more platform agnostic when parsing filenames and split on 'os.sep' instead of the magic literal '/'.  In this case, the tests all run in the same type of Linux environment, but that may not always be the case.

tests/hla/test_run_svmpoller.py
```diff
@@ -0,0 +1,141 @@
...
+        # The fully qualified path designated in the poller file is the shared cache
+        file_list = table["col8"].tolist()
+
+        # Check if the files to be processed are in the same directory as poller
+        # directory, otherwise they need to be copied from the on-line cache
+        for full_filename in file_list:
+            filename = full_filename.split("/")[-1]
+            log.info("Looking for file {}".format(filename))
+            if os.path.exists(poller_path + "/" + filename):
+                shutil.copy2(poller_path + "/" + filename, ".")
```
stsci-hack: Generating a new filename would be a little more transparent with the use of 'os.path.join(poller_path, filename)'.

tests/hla/test_run_svmpoller.py
mdlpstsci: Using os methods to be OS independent.

tests/hla/test_run_svmpoller.py
mdlpstsci: /Users/mdelapena/repos/drizzlepac/drizzlepac

Issue #502, opened 2019-11-14
Review all single-visit code to define log message levels


stsci-hack:
All log messages generated by the single-visit processing need to be reviewed to set all messages to the appropriate log level: DEBUG, INFO, WARNING, ERROR, CRITICAL. 

This will then set up the trailer file (log file) output from the processing to include only those messages relevant to the log level set upon processing (through #501).

stsci-hack:
Developers will start with modification to the following modules:
@mdlpstsci : product, catalog_utils, and hapsequencer
@dulude : runsinglehap, hla_flag_filter, config_utils and related modules
@stsci-hack : alignment modules

stsci-hack:
The format of the log messages can be made Splunk/Kibana compatible alongside all other DMS-generated pipeline logs using:

```
MSG_DATEFMT = '%Y%j%H%M%S'
SPLUNK_MSG_FORMAT = '%(asctime)s %(levelname)s src=%(name)s- %(message)s'
log = logutil.create_logger(__name__, level=logutil.logging.INFO, stream=sys.stdout, 
                            format=SPLUNK_MSG_FORMAT, datefmt=MSG_DATEFMT)
```

Issue #501, opened 2019-11-14
Add log_level parameter to runsinglehap


stsci-hack:
A new parameter, `--log_level` and `-l`, should be added to the `runsinglehap` task to set the Python logging level to one of the pre-defined values: DEBUG, INFO, WARNING, ERROR, CRITICAL and so on. 

This parameter will then need to be used to reset the global Python logging level as the first operation in the `perform()` function, if defined by the user at the command line.

s-goldman:
Looks like this work was completed in #506

PR #506, opened 2019-11-15
Add command-line input "log_level" to runsinglehap.py, hapsequencer.run_hap_processing

dulude:
- This PR allows the user to specify the logging level (CRITICAL/ERROR/ Etc...) as a command-line input to runsinglehap.py. (See issue #501)
The runsinglehap.perform() subroutine and the the main logging definition at the top of the code were modified to accommodate a user-specified log level.
- The hapsequencer.run_hap_processing()  subroutine and the the main logging definition at the top of the code were modified to accommodate a user-specified log level.
- Fellow Burgundy team members can now use log_level for their work towards issue #502.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -277,7 +277,7 @@ def run_hap_processing(input_filename, diagnostic_mode=False, use_defaults_confi
     # This routine needs to return an exit code, return_value, for use by the calling
     # Condor/OWL workflow code: 0 (zero) for success, 1 for error condition
     return_value = 0
-    log.level = log_level
+    log.setLevel(log_level)
     # Define trailer file (log file) that will contain the log entries for all processing
     if isinstance(input_filename, str):  # input file is a poller file -- easy case
         logname = input_filename.replace('.out', '.log')
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -25,7 +25,7 @@ from stwcs import wcsutil
 
 __taskname__ = 'hapsequencer'
 log_level = logging.ERROR
-log = logutil.create_logger(__name__, level=log_level, stream=sys.stdout)
+log = logutil.create_logger(__name__, stream=sys.stdout)
 
 __version__ = 0.1
 __version_date__ = '07-Nov-2019'
@@ -228,7 +228,7 @@ def create_drizzle_products(total_list):
 
 def run_hap_processing(input_filename, debug=False, use_defaults_configs=True,
                        input_custom_pars_file=None, output_custom_pars_file=None, phot_mode="both",
-                       log_level=logutil.logging.WARNING):
+                       log_level=logutil.logging.INFO):
     """
     Run the HST Advanced Products (HAP) generation code.  This routine is the sequencer or
     controller which invokes the high-level functionality to process the single visit data.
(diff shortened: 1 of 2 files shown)


drizzlepac/runsinglehap.py
```diff
@@ -53,13 +53,27 @@ def perform(input_filename, **kwargs):
...
+        kwargs['log_level'] = logutil.logging.CRITICAL
+    elif kwargs['log_level'] == "error":
+        kwargs['log_level'] = logutil.logging.ERROR
+    elif kwargs['log_level'] == "warning":
+        kwargs['log_level'] = logutil.logging.WARNING
+    elif kwargs['log_level'] == "INFO":
+        kwargs['log_level'] = logutil.logging.INFO
+    else:
+        kwargs['log_level'] = logutil.logging.DEBUG
+
```
mdlpstsci: Do we really want the default logging mode to be DEBUG?  I would have thought the default should be INFO.

drizzlepac/hapsequencer.py
```diff
@@ -23,8 +24,8 @@
 from stwcs import wcsutil
 
 __taskname__ = 'hapsequencer'
-log_level = logging.INFO
-log = logutil.create_logger(__name__, level=log_level, stream=sys.stdout)
+log_level = logging.ERROR
```
mdlpstsci: I am not sure what this assignment is actually doing.

drizzlepac/runsinglehap.py
dulude: issue resolved by commit [889def1](https://github.com/spacetelescope/drizzlepac/pull/506/commits/889def19be2f0eb53bd402492bd61354d0a03474)

drizzlepac/runsinglehap.py
```diff
@@ -53,13 +53,29 @@ def perform(input_filename, **kwargs):
...
+        kwargs['log_level'] = logutil.logging.ERROR
+    elif kwargs['log_level'] == "warning":
+        kwargs['log_level'] = logutil.logging.WARNING
+    elif kwargs['log_level'] == "info":
+        kwargs['log_level'] = logutil.logging.INFO
+    elif kwargs['log_level'] == "debug":
+        kwargs['log_level'] = logutil.logging.DEBUG
+    else:
+        kwargs['log_level'] = logutil.logging.INFO
+
```
stsci-hack: The logic could be simplified to only 1 if_else statement (no elif's) with the use of a dict which translates between user-input values and logging.[CRITICAL|ERROR|...] values.  Also, user-input should be converted to all lower case for the comparison to avoid the code missing what the user 'intended' in case they try to replicate logging level names (all upper-case).

drizzlepac/hapsequencer.py
dulude: I think you are looking at an older version. These lines are different in the latest version.

drizzlepac/runsinglehap.py
dulude: Resolved by commit [623790e](https://github.com/spacetelescope/drizzlepac/pull/506/commits/623790ee0698e3d9b632399993f5f30c5f3266a6)

dulude:
> I would just like some feed back on my two comments. I think Warren should weigh in on the default log level.

Hey @mdlpstsci , I've processed your comments into the code. The default logging level is 'info' everywhere, and I've removed the 'log_level = logging.ERROR' line (formerly line 27) in hapsequencer.py.

drizzlepac/hapsequencer.py
```diff
@@ -276,7 +277,7 @@ def run_hap_processing(input_filename, diagnostic_mode=False, use_defaults_confi
     # This routine needs to return an exit code, return_value, for use by the calling
     # Condor/OWL workflow code: 0 (zero) for success, 1 for error condition
     return_value = 0
-
+    log.level = log_level
```
mdlpstsci:
```suggestion
    log.setLevel(log_level)
```

PR #512, opened 2019-11-19
Revise logging more

stsci-hack:
These changes address a number of issues with the single-visit processing:

- Fixed the problem with processing icnw34.out: find_fwhm ended up with a NaN in the kernel's fit flux results
- Update's align.py to better match (not exactly match) the results of alignimages (which does not use the OO alignment code in align_utils).
- Revises some logging statements to use the new format for the prefix (date stamped lines)
- Exposes code that reads in default configuration parameter index file to functional calls 
- Updates disp_flag_by_source to add function to write out separate flag region files to disk for use in DS9 directly
This is not a complete update of all logging in the alignment-related code, so more can be done there as others see fit.


stsci-hack:
X-Center and Y-Center precision in output tables now reset to 3 digits.

drizzlepac/hlautils/config_utils.py
```diff
@@ -190,15 +189,8 @@ def _determine_conditions(self, prod_obj):
...
-        base_dir = os.path.dirname(os.path.dirname(code_dir))
-        self.pars_dir = os.path.join(base_dir, "pars/hap_pars")
-        cfg_index_fileanme = self.inst_det + "_index.json"
-        cfg_index_filename = os.path.join(self.pars_dir, cfg_index_fileanme)
-
-        with open(cfg_index_filename) as jsonFile:
-            json_string = jsonFile.read()
-            self.full_cfg_index = json.loads(json_string)
+        self.full_cfg_index, self.pars_dir = read_index(self.instrument, 
+                                                        self.detector)
```
mdlpstsci: Class attributes should at the least be declared in the init of the class and initialized to a default value (e.g., None). In this way the reader knows there is a class attribute which has yet to be computed.  When feasible, attributes computed by methods of a class should return the values rather than setting the values in the method in order to avoid "hiding" the setting of the values.

drizzlepac/pars/hap_pars/default_parameters/acs/wfc/acs_wfc_alignment_all.json
```diff
@@ -17,13 +17,13 @@
...
     "generate_source_catalogs":
       {
-        "box_size":50,
+        "box_size":27,
         "win_size":3,
         "nsigma":5.0,
         "centering_mode": "starfind",
-        "bkg_estimator": "MedianBackground",
+        "bkg_estimator": "SExtractorBackground",
         "rms_estimator": "StdBackgroundRMS",
```
mdlpstsci: This is the only configuration file which has the bkg_estimator set to SExtractorBackground.  I just want to make sure this is intended.

stsci-hack: Actually, that is an oversight, as WFC3/UVIS should probably also use the same estimator.  This will be addressed shortly in a patch commit.

stsci-hack:
Rebased to pick up PRs that were merged in ahead of this...

drizzlepac/devutils/disp_sources_by_flag.py
```diff
@@ -46,14 +47,15 @@ def display_regions(imgname,reg_dict,flag_counts,n_sources):
...
-                                                                                           padding2*".",
-                                                                                           flag_counts[ctr],
-                                                                                           100.0*(float(flag_counts[ctr])/float(n_sources))))
+        print("Frame {}: Bit value {}{}{}{}{} sources flagged ({}% of all sources)".format(ctr + 1,
+                            bit_val, "." * padding1,
+                            flag_meanings[ctr],
+                            padding2 * ".",
+                            flag_counts[ctr],
+                            100.0 * (float(flag_counts[ctr]) / float(n_sources))))
         if ctr != 0:
```
mdlpstsci: Is it possible for n_sources to be zero?

stsci-hack: This routine gets called directly (mostly interactively) by the user when they have sources they want to display, and 'n_sources' is one of the parameters provided by the user.  So, I would say it is not likely, simply due to the fact that if 'n_sources' was 0, there would be no need to call this code in the first place.

drizzlepac/runastrodriz.py
```diff
@@ -102,11 +103,11 @@
...
-              "WFC3/UVIS": {'sigma': 1.5, 'good_bits': ~14588},
-              "ACS/WFC": {'sigma': 1.5, 'good_bits': ~14588},
+              "WFC3/UVIS": {'sigma': 1.5, 'good_bits': 1360},
+              "ACS/WFC": {'sigma': 1.5, 'good_bits': 1360},
               "ACS/SBC": {'sigma': 2.0, 'good_bits': 0},
-              "ACS/HRC": {'sigma': 1.5, 'good_bits': ~14588},
-              "WFPC2/PC": {'sigma': 1.5, 'good_bits': ~14588}}
+              "ACS/HRC": {'sigma': 1.5, 'good_bits': 1360},
+              "WFPC2/PC": {'sigma': 1.5, 'good_bits': 1360}}
 sub_dirs = ['OrIg_files', 'pipeline-default']
```
mdlpstsci: Add a comment which shows the individual components of 1360 and the associated meanings.  Also show the meaning of 4096 and 512.

stsci-hack:
> I noticed that the 'X-Center' and 'Y-Center' values in the point catalogs still have 13 significant digits. This should be reduced to 3 or 4.

@dulude : This has been addressed in 25277a6 .

drizzlepac/runastrodriz.py
stsci-hack: @mdlpstsci : Addressed in 591760f and 3c24be8.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -584,16 +586,44 @@ def extract_sources(img, dqmask=None, fwhm=3.0, kernel=None, photmode=None,
         imgarr = img
 
     segm = detect_sources(imgarr, segment_threshold, npixels=source_box,
-                          filter_kernel=kernel)
+                          filter_kernel=kernel, connectivity=4)
+
+    log.debug("Creating segmentation map for {} ".format(outroot, kernel.shape))
+    if kernel is not None:
```
mdlpstsci: Do you really intend to print both `outroot` and `kernel.shape` here?

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -584,16 +586,44 @@ def extract_sources(img, dqmask=None, fwhm=3.0, kernel=None, photmode=None,
...
+        log.debug("   based on a default kernel.")
+
+    num_brightest = 10 if len(segm.areas) > 10 else len(segm.areas)
+    mean_area = np.mean(segm.areas)
+    max_area = np.sort(segm.areas)[-1 * num_brightest:].mean()
+    # This section looks for crowded fields where segments run into each other
+    # By reducing the size of the kernel used for segment detection, this can be minimized
+    # in crowded fields.  Also, mean area is used to try to avoid this logic for fields with
+    # several large extended sources in an otherwise empty field.
+    if max_area > 3422 and mean_area > (kernel_area / 2):  # largest > 33-pix radius source
```
mdlpstsci: Define the magic number 3422.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -615,36 +645,40 @@ def extract_sources(img, dqmask=None, fwhm=3.0, kernel=None, photmode=None,
     # convert segm to mask for daofind
     if centering_mode == 'starfind':
         src_table = None
+
         # daofind = IRAFStarFinder(fwhm=fwhm, threshold=5.*bkg.background_rms_median)
-        log.info("Setting up DAOStarFinder with: \n    fwhm={}  threshold={}".format(fwhm, dao_threshold))
-        daofind = DAOStarFinder(fwhm=fwhm, threshold=dao_threshold)
+        log.debug("Setting up DAOStarFinder with: \n    fwhm={}  threshold={}".format(fwhm, dao_threshold))
+        # daofind = DAOStarFinder(fwhm=fwhm, threshold=dao_threshold)
```
mdlpstsci: There is a debug message here, but the line of code(Line 651)  is actually commented out.  I do see this computation at Line 680.  Should the debug statement be moved down to Line 680?

stsci-hack: Corrected in 219bfa6.

drizzlepac/hlautils/astrometric_utils.py
stsci-hack: The magic number was moved to the top of the module with a slight re-definition and explanation in 219bfa6.

drizzlepac/hlautils/astrometric_utils.py
stsci-hack: Corrected in 219bfa6.

PR #531, opened 2019-12-02
Added function to verify GAIA WCSNAMEs

stsci-hack:
This change works entirely within runastrodriz (pipeline processing) to insure that any exposure taken after 10/1/2017 will have '-GSC240' at the end of the apriori-corrected WCSNAME since it will not be getting that from the astrometry database.


stsci-hack:
@mdlpstsci: This should be ready to review now...

Issue #521, opened 2019-11-21
Sort SVM images before alignment


mdlpstsci:
For SVM processing, the images in the poller file or input list should be sorted so the reference (first image) has a maximum number of potential sources.  The images be sorted in the following priority:
- redder filters with priority over bluer filters
- wider filters with priority over narrower filters
- longer exposures with priority over shorter exposures

Use FITS keywords photplam, photbw, and the exposure time for the algorithm.

mdlpstsci:
Jen suggests
To determine which filter to align first, use the number of sources in each catalog. (This gives a simple way to approximate the combined effect of bandwidth, exposure time, and spectral energy distribution on the signal to noise for each observation).

stsci-hack:
Sorting can be more precisely done by computing the intersection of the GAIA DR2 passband with a flat source spectrum computed for the observation based on PHOTFLAM, PHOTPLAM, PHOTBW and EXPTIME. 

-   The [GAIA DR2 passband](https://www.cosmos.esa.int/documents/29201/1645651/GaiaDR2_Revised_Passbands_ZeroPoints.zip/54db454f-69cb-ea0c-15be-f1b1f597f191) can be found online and added to the code as an parameter file. 
-   This intersection can be computed using [pysynphot's FlatSpectrum() for the observation](https://pysynphot.readthedocs.io/en/latest/spectrum.html#flat) and [pysynphot's custom spectrum tutorial for the GAIA DR2 passband](https://pysynphot.readthedocs.io/en/latest/tutorials.html#tutorial-10-spectrum-from-custom-text-file).

The count rate from this Composite spectrum can then be used as the primary sorting mechanism for observations taken in filters which overlap the GAIA passband.  For observations outside the GAIA passband (WFC3/IR, UV filters), the total integrated counts for the observations spectra sorted by distance from the central wavelength from the GAIA passband will need to be used instead.  This may simply be estimated as the area of the flat passband (PHOTFLAM*EXPTIME*PHOTBW) sorted by abs(delta(PHOTPLAM - GAIA_PEAK_LAMBDA)).

stsci-hack:
Successful sorting of inputs will then be used to update the input poller list for creation of the total and filter products. 
[...]

PR #553, opened 2020-01-10
Implement v1.0 of single fit for entire visit

stsci-hack:
This implements the initial version of a single WCS fit for all exposures in a single visit for a single instrument (across all detectors used for the visit) as described in #521.  This does NOT implement the actual input exposure sorting described in #521 that will be expected by this PR.  

- User values and default values for tolerance for 'match_relative_fit' were lowered to 0.5.  This relies on the assumption that the errors in the proper motions for the GAIA sources are going to be within 20-25 mas depending on the time of the observation for enough sources to get a good alignment to GAIA. 
-  The fit uses the 'relative' fit method to align all exposures to each other THEN perform a single uniform fit to GAIA for the entire mosaic and relies on background and source determinations performed for image-to-image alignment.  Image-to-image alignment retained in order to serve as the default alignment of the visit. 
- A new label has been defined (EVM) to indicate that only an image-to-image alignment could be performed and that the final fit to GAIA was unsuccessful.  
- The alignment code was pulled into it's own function that gets called by hapsequencer to try to isolate any further changes to alignment processing. 
- Fixes logging for align_utils as it needed to be set upon getting called by `product.py`.  

`
NOTE:
In addition to these alignment based changes, 'hlautils/cell_utils' was updated to include an implementation of the Ckmeans algorithm for experimentation with grouping exposures by date for use in the multi-visit processing. 
`

This code successfully aligned all the data from ib4606 (both UVIS and IR) into a single alignment to GAIA.

### DIFF
--- a/drizzlepac/hlautils/align_utils.py
+++ b/drizzlepac/hlautils/align_utils.py
@@ -61,6 +61,7 @@ class AlignmentTable:
                           plot=False, vmax=None, deblend=False
         """
         log.setLevel(log_level)
+
         # Register fit methods with the class
         self.fit_methods = {'relative': match_relative_fit,
                             '2dhist': match_2dhist_fit,
@@ -193,6 +194,9 @@ class AlignmentTable:
 
     def perform_fit(self, method_name, catalog_name, reference_catalog):
         """Perform fit using specified method, then determine fit quality"""
+        inputs = [img.meta['name'] for img in self.imglist]
+        log.info("Performing {} astrometric fit to {} for: \n    {}".format(method_name,
+                                                                            reference_catalog, inputs))
         imglist = self.fit_methods[method_name](self.imglist, reference_catalog,
                                                 **self.fit_pars[method_name])
 
@@ -496,10 +500,10 @@ class HAPImage:
         # from one image to a single source in the
         # other or vice-versa.
         # Create temp DQ mask containing all pixels flagged with any value EXCEPT 256
-        non_sat_mask = bitfield_to_boolean_mask(dqarr, ignore_flags=256+2048)
+        non_sat_mask = bitfield_to_boolean_mask(dqarr, ignore_flags=256 + 2048)
 
         # Create temp DQ mask containing saturated pixels ONLY
-        sat_mask = bitfield_to_boolean_mask(dqarr, ignore_flags=~(256+2048))
+        sat_mask = bitfield_to_boolean_mask(dqarr, ignore_flags=~(256 + 2048))
 
         # Ignore sources where only a couple of pixels are flagged as saturated
         sat_mask = ndimage.binary_erosion(sat_mask, iterations=1)
@@ -562,6 +566,8 @@ def match_relative_fit(imglist, reference_catalog, **fit_pars):
     log.info("{} (match_relative_fit) Cross matching and fitting {}".format("-" * 20, "-" * 27))
     # 0: Specify matching algorithm to use
     match = tweakwcs.TPMatch(**fit_pars)
+    log.debug("Relative fit configured with: \n  {}".format(fit_pars))
+
     # match = tweakwcs.TPMatch(searchrad=250, separation=0.1,
     #                          tolerance=100, use2dhist=False)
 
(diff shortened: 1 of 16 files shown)


stsci-hack:
Commit 9b37de3 implements:

- Revision of how to measure saturated stars for alignment in 'astrometric_utils.extract_sources' by looking for more than 3 pixels within 10% of the max value, then using segmentation.center_of_mass for position instead of daofind.  
- Updated parameters for 'relative_alignment' to use separation=4.0 which eliminates(?) problem with matching multiple sources in one catalog to another and allows 'expand_refcat' to work more reliably. 
- Simplified SVM alignment to GAIA to replace original 'image-by-image' alignment with a custom FilterProduct object that includes all exposures in the total image and performs a single 'relative' fit with 'fit_label=SVM'.  Default 'fit_label' reset to 'EVM' to allow for recognition of incomplete alignment to GAIA.

This commit was used to successfully align ib4606 to GAIA.

drizzlepac/hlautils/align_utils.py
```diff
@@ -489,20 +493,9 @@ def build_dqmask(self, chip=None):
         # source match algorithm from trying to match multiple sources
         # from one image to a single source in the
         # other or vice-versa.
-        # Create temp DQ mask containing all pixels flagged with any value EXCEPT 256
-        non_sat_mask = bitfield_to_boolean_mask(dqarr, ignore_flags=256+2048)
-
-        # Create temp DQ mask containing saturated pixels ONLY
-        sat_mask = bitfield_to_boolean_mask(dqarr, ignore_flags=~(256+2048))
+        # Create temp DQ mask containing all pixels flagged with any value EXCEPT 512 (bad ref pixel)
```
mdlpstsci: Update "grow out" comment as it refers to saturated pixels only.

drizzlepac/hlautils/astrometric_utils.py
```diff
@@ -700,38 +701,49 @@ def extract_sources(img, dqmask=None, fwhm=3.0, kernel=None, photmode=None,
...
-                    # Add row for detected source to master catalog
-                    # apply offset to slice to convert positions into full-frame coordinates
-                    seg_table['xcentroid'] += seg_xoffset
-                    seg_table['ycentroid'] += seg_yoffset
-                    src_table.add_row(seg_table[max_row])
+
+                # Add logic to remove sources which have more than 8 pixels
+                # within 10% of the max value in the source segment, a situation
+                # which would indicate bleeding from a saturated source
+                if (detection_img > detection_img.max() * 0.9).sum() > 3:
```
mdlpstsci: Which is incorrect here - the "3" in the code or the "8" in the comment?  Or is this doing something I do not understand?

drizzlepac/pars/hap_pars/user_parameters/wfc3/uvis/wfc3_uvis_alignment_all.json
```diff
@@ -36,9 +36,9 @@
       },
     "match_relative_fit":
       {
-        "searchrad": 75,
-        "separation": 0.1,
-        "tolerance": 2,
+        "searchrad": 25,
```
mdlpstsci: Was the searchrad here intentionally changed to 25 from 75?  I ask as this is the only searchrad value that changed, and it is very different than all the other values?

stsci-hack: Restored to a value of 75 in 6b78487

drizzlepac/hlautils/astrometric_utils.py
stsci-hack: This comment was corrected in 6b78487 to report on the value of '3' being used as a radius of 0.5 pixels area where all pixels are 'the same' to indicate a saturated source.

drizzlepac/hlautils/align_utils.py
stsci-hack: The comment was completely removed since it referred to code that had been removed entirely.  Fixed in 6b78487 .

Issue #554, opened 2020-01-16
Populate RMS values for WCS


stsci-hack:
Keywords need to be added to the Primary (active) WCS to reflect the level of error for the fit used to determine the WCS.  These keywords should be: 

  - RMS_RA
  - RMS_DEC

These values should come from the headerlet for the active WCS.  This requires that these keywords be correctly populated in the headerlet themselves, and effort needs to be done to insure that these keywords are populated in the headerlets.

stsci-hack:
Resolved with #555 on 17-Jan-2020.

PR #555, opened 2020-01-17
Add rms to headers

stsci-hack:
The RMS_RA, RMS_DEC, and NMATCHES keywords were being added to the headerlets created during the a posteriori alignment step of runastrodriz.  However, those headerlets were not getting copied over from the aposteriori processing sub-directory for use as the final headerlet.  

These changes insure that the headerlets get copied as the working headerlets.  In addition, the aposteriori aligned FLC/FLT images get updated with the values of these keywords as well with these code changes.  Between the two sets of changes, all critical fit information finally makes it into the outputs generated and used by the pipeline.  

This addresses #554.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -551,7 +551,7 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
         # Generate headerlets for each updated FLT image
         hlet_msg = _timestamp("Writing Headerlets started")
         for fname in _calfiles:
-            print("Creating new headerlet from {}".format(fname))
+            hlet_msg += "Creating new headerlet from {}".format(fname)
             frootname = fileutil.buildNewRootname(fname)
             hname = "%s_flt_hlet.fits" % frootname
             # Write out headerlet file used by astrodrizzle, however,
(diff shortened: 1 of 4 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -636,6 +638,8 @@ def run_driz(inlist, trlfile, calfiles, mode='default-pipeline', verify_alignmen
         _pyd_err = trlfile.replace('.tra', '_pydriz.stderr')
 
         try:
+            print("Running AstroDrizzle on: {}".format(infile))
+
```
mdlpstsci: Convert this to a log message or delete.

drizzlepac/runastrodriz.py
```diff
@@ -550,6 +551,7 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
         # Generate headerlets for each updated FLT image
         hlet_msg = _timestamp("Writing Headerlets started")
         for fname in _calfiles:
+            print("Creating new headerlet from {}".format(fname))
```
mdlpstsci: Convert this to a log message or delete.

drizzlepac/runastrodriz.py
stsci-hack: Addressed in 8514ba1

drizzlepac/runastrodriz.py
stsci-hack: Addressed in 8514ba1

PR #561, opened 2020-01-22
Close open files in align upon error

stsci-hack:
Errors in the alignment can often cause the a posteriori alignment to exit before determining a fit; sometimes as the result of an Exception, sometimes as a result of the data (too few sources,...).  Processing in the pipeline, though, requires that all open files be closed before exiting the a posteriori alignment step.  This includes both logging files as well as any open input or output files.  

The changes here explicitly close all the input files anytime the alignment code returns before full completion (an oversight in the original implementation).  There are also changes to define default values upon returning early to avoid subsequent Exceptions as well due to lack of a "best fit".  

These changes were tested using the singleton `ibbq01n0q` as reported by DSB during their nightly auto tests of this code in preparation for operational testing.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -816,6 +816,7 @@ def verify_alignment(inlist, calfiles, calfiles_flc, trlfile,
                         trlstr = "Could not align {} to absolute astrometric frame\n"
                         trlmsg += trlstr.format(row['imageName'])
                         print(trlmsg)
+                        _updateTrlFile(trlfile, trlmsg)
                         return None
             except Exception:
                 # Something went wrong with alignment to GAIA, so report this in
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -71,7 +71,7 @@ from collections import OrderedDict
 import datetime
 try:
     from psutil import Process
-except:
+except ImportError:
     Process = None
 
 # THIRD-PARTY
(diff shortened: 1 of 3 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -813,12 +815,14 @@ def verify_alignment(inlist, calfiles, calfiles_flc, trlfile,
                     else:
                         trlstr = "Could not align {} to absolute astrometric frame\n"
                         trlmsg += trlstr.format(row['imageName'])
+                        print(trlmsg)
```
mdlpstsci: Should this be _updateTrlFile(trlfile, trlmsg) and no print statement?

drizzlepac/runastrodriz.py
```diff
@@ -813,12 +815,14 @@ def verify_alignment(inlist, calfiles, calfiles_flc, trlfile,
...
                         return None
             except Exception:
                 # Something went wrong with alignment to GAIA, so report this in
                 # trailer file
                 _trlmsg = "EXCEPTION encountered in align...\n"
                 _trlmsg += "   No correction to absolute astrometric frame applied!\n"
+                print(_trlmsg)
```
mdlpstsci: Since the message is going to the Trailer file, do we need a print statement here?

drizzlepac/runastrodriz.py
stsci-hack: A call to _updateTrlFile( ) would actually be helpful here (see 7a46930) to insure that these comments actually make it to the trailer file as a 'permanent' record of the processing.  However, this information does not make it to stdout for the user or processing system to see unless we add the 'print()' statement, where it can be logged by the pipeline separately.  These are both necessary here since the next line causes the code to exit early before adding this information to any trailer file or stdout.

drizzlepac/runastrodriz.py
stsci-hack: Unfortunately, these comments may be recorded in the trailer file, but the end user (or pipeline) does not know why the code quits early without sending these comments to stdout through this 'print()' statement.

Issue #537, opened 2019-12-05
New keywords for SVM products


stsci-hack:
A set of new keywords has been identified as being potentially useful for evaluating the single-visit products; namely, 

- Fraction of pixels with data
- Mean exposure time in the pixels with data
- Median exposure time in the pixels with data
- Mean number of exposures in the pixels with data
- Median number of exposures in the pixels with data

Names for these keywords which are FITS compliant need to be defined, and these values need to be computed based on `drizzlepac.hlautils.cell_utils` products.  The keywords will then need to be added to the PRIMARY header of the FILTER SVM products and TOTAL SVM product at the least, and perhaps also to the single EXPOSURE SVM products as well.

Additional keywords may still be identified for addition by INS as part of this development.

nden:
Resolved in #569

PR #569, opened 2020-01-23
Add SVM footprint keywords to product headers

stsci-hack:
Addresses #537.

### DIFF
--- a/drizzlepac/hlautils/product.py
+++ b/drizzlepac/hlautils/product.py
@@ -29,8 +29,8 @@ log = logutil.create_logger(__name__, level=logutil.logging.NOTSET, stream=sys.s
 MASK_KWS = {"NPIXFRAC": [None, "Fraction of pixels with data"],
             "MEANEXPT": [None, "Mean exposure time per pixel with data"],
             "MEDEXPT": [None, "Median exposure time per pixel with data"],
-            "MEANNEXP": [None, "Mean exposures per pixel with data"],
-            "MEDNEXP": [None, "Median exposures per pixel with data"],
+            "MEANNEXP": [None, "Mean number of exposures per pixel with data"],
+            "MEDNEXP": [None, "Median number of exposures per pixel with data"],
             }
 
 class HAPProduct:
--- a/drizzlepac/hlautils/product.py
+++ b/drizzlepac/hlautils/product.py
@@ -60,11 +60,56 @@ class HAPProduct:
         # this attribute is updated in the hapsequncer.py module (run_hla_processing()).
         self.configobj_pars = None
 
+        # Initialize attributes for use in generating the output products
+        self.meta_wcs = None
+        self.mask = None
+        self.mask_kws = MASK_KWS.copy()
+
     # def print_info(self):
         # """ Generic print at this time to indicate the information used in the
         #     construction of the object for debug purposes.
         # """
         # print("Object information: {}".format(self.info))
+
+    def generate_footprint_mask(self):
+        """ Create a footprint mask for a set of exposure images
+
+            Create a mask which is True/1/on for the illuminated portion of the image, and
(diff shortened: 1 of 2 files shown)


drizzlepac/hlautils/product.py
```diff
@@ -23,6 +24,14 @@
 log = logutil.create_logger(__name__, level=logutil.logging.NOTSET, stream=sys.stdout,
                             format=SPLUNK_MSG_FORMAT, datefmt=MSG_DATEFMT)
 
+# Define keywords to be added to SVM products to describe the overlap of
+# exposures onto the output WCS footprint as determined from the SkyFootprint masks
+MASK_KWS = {"NPIXFRAC": [None, "Fraction of pixels with data"],
+            "MEANEXPT": [None, "Mean exposure time per pixel with data"],
+            "MEDEXPT": [None, "Median exposure time per pixel with data"],
+            "MEANNEXP": [None, "Mean exposures per pixel with data"],
+            "MEDNEXP": [None, "Median exposures per pixel with data"],
```
mdlpstsci: Clarify the commentary part of the keyword for MEANNEXP and MEDNEXP - "[Mean|Median] **_number of_** exposures..."

stsci-hack: Fixed in b421222.

drizzlepac/hlautils/product.py
```diff
@@ -154,9 +169,17 @@ def generate_footprint_mask(self):
         """
         footprint = cell_utils.SkyFootprint(self.meta_wcs)
         exposure_names = [element.full_filename for element in self.edp_list]
-        footprint.build(exposure_names)
+        footprint.build(exposure_names, scale=True)
         self.mask = footprint.total_mask
 
+        # Compute footprint-based SVM-specific keywords for product image header
+        good_pixels = self.mask > 0
+        self.mask_kws['NPIXFRAC'][0] = good_pixels.sum()
```
mdlpstsci: This looks to be the sum of pixels with good data versus a fraction (good_pixels/all_pixels).

drizzlepac/hlautils/product.py
```diff
@@ -272,10 +303,35 @@ def wcs_drizzle_product(self, meta_wcs):
...
+        self.mask = footprint.total_mask
+
+        # Compute footprint-based SVM-specific keywords for product image header
+        good_pixels = self.mask > 0
+        self.mask_kws['NPIXFRAC'][0] = good_pixels.sum()
+        self.mask_kws['MEANEXPT'][0] = np.mean(footprint.scaled_mask[good_pixels])
+        self.mask_kws['MEDEXPT'][0] = np.median(footprint.scaled_mask[good_pixels])
+        self.mask_kws['MEANNEXP'][0] = np.mean(self.mask[good_pixels])
+        self.mask_kws['MEDNEXP'][0] = np.median(self.mask[good_pixels])
+
```
mdlpstsci: It looks like _generate_footprint_mask()_  and _wcs_drizzle_product()_ are basically the same for the _TotalProduct_ and _FIlterProduct_ classes.  It is clear you do not want the _ExposureProduct_ class to have these methods.  You could put these method in the _HAPProduct_ base class and then override the methods in the _ExposureProduct_ class with "pass" which is a cheapo solution. Alternatively, you could do what is discussed at this URL which I found compelling:
https://medium.com/@maouu/sorry-but-youre-wrong-aea1b88ffc03

stsci-hack: The alternative definitions using __getattribute__ and __dict__ for ExposureProduct was implemented successfully(?) in 871bdef.  Thanks for the pointer!

drizzlepac/hlautils/product.py
stsci-hack: Also fixed in 871bdef.

Issue #582, opened 2020-02-04
2dhist requires integer number of bins


srodney:
In imgclasses.py the _estimate_2dhist_shift() function passes the argument searchrad to the _xy_2dhist() function as 'r' which is then passed to the numpy histogram2d function as the number of bins, using 2*r+1.  In numpy v1.18.1 (and probably earlier) this argument must be an int, and it raises an exception if it is not a data type that can be safely converted to an int.

The searchrad will not generally be defined as an integer, so I propose that the best fix is to use int(round(searchrad)) in _estimate_2dhist_shift() to force r to be an int.      I'll issue a pull request, but I'd welcome careful review (@mcara maybe?) that this will not break the fixes for issue #278

mcara:
@stsci-hack A slightly different approach compared to the one implemented in #583 is used in https://github.com/spacetelescope/tweakwcs/blob/98b7f9aea69b347375a86f834da939ed81b0764c/tweakwcs/matchutils.py#L277

Somehow I forgot to "back port" this `tweakwcs` implementation to `drizzlepac`. I propose to switch to `tweakwcs` implementation as it a little bit "safer".

mcara:
Fixed via #587

PR #583, opened 2020-02-04
proposed bug fix for 2dhist int bin number requirement

srodney:
* see issue #582 and #278


stsci-hack:
This addresses the issue that 'np.histogram2d' requires an 'int' for the number of bins parameter, a parameter derived directly from the value of 'r' being passed in by this fix.  I see no problems with this simple fix.

PR #586, opened 2020-02-06
Modified the handling of the filter values to accommodate SBC names

mdlpstsci:
SBC filter names are currently malformed (e.g., "F140LP;N/A;F140LP" in the SVM poller manifests.  The third token delimited by ";" should not be present. Also, the "N/A" needs
to be replaced with "none" as the slash cannot be part of the output filename.  The code has been modified to handle this situation.

### DIFF
--- a/drizzlepac/hlautils/analyze.py
+++ b/drizzlepac/hlautils/analyze.py
@@ -246,6 +246,7 @@ def analyze_data(input_file_list, log_level=logutil.logging.NOTSET):
         no_proc_value = None
         do_process = True
         # Imaging vs spectroscopic or coronagraphic
+        print("******OBSTYPE: {}".format(obstype))
         if obstype != 'IMAGING':
             no_proc_key = OBSKEY
             no_proc_value = obstype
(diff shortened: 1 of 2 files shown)


drizzlepac/hlautils/poller_utils.py
```diff
@@ -298,23 +298,34 @@ def determine_filter_name(raw_filter):
...
         if 'clear' not in filt:
             output_filter_list.append(filt)
 
     if not output_filter_list:
         filter_name = 'clear'
+    elif output_filter_list[0].startswith('pol'):
+        output_filter_list.reverse()
     else:
-        if output_filter_list[0].startswith('pol'):
+        if output_filter_list[0].startswith('none'):
```
stsci-hack: In [1009]: rawfilt = 'POL120V;F555W'                                                                                                

In [1010]: determine_filter_name(rawfilt)                                                                                           
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-1010-230fd5bd86d9> in <module>
----> 1 determine_filter_name(rawfilt)

<ipython-input-998-a241b87a1552> in determine_filter_name(raw_filter)
     25         filter_name = delimiter.join(output_filter_list)
     26 
---> 27     return filter_name

UnboundLocalError: local variable 'filter_name' referenced before assignment

mdlpstsci:
Changed to state of "Do not merge" until some issues are clarified as to the proper naming convention ultimately to be used.

drizzlepac/hlautils/analyze.py
```diff
@@ -246,6 +246,7 @@ def analyze_data(input_file_list, log_level=logutil.logging.NOTSET):
         no_proc_value = None
         do_process = True
         # Imaging vs spectroscopic or coronagraphic
+        print("******OBSTYPE: {}".format(obstype))
```
stsci-hack: Should this be a log message (log.debug? log.info?) or removed altogether?

mdlpstsci: Unnecessary print statement.

Issue #590, opened 2020-02-12
error identifying primary extension number in sci.fits files


srodney:
Line 445 in catalogs.py has a check on the extname tuple, designed to catch the case where the wcs object has an empty string for the extension name, and None for the extension number: self.wcs.extname=('',None).   I've been tripping on a bug that manifests when the wcs object has self.wcs.extname=('',1).   The fits file that provided the WCS in this case has the WCS info (and data array) in the primary HDU, and the primary HDU has extver=1.    In my case, this fits file is a drz_sci.fits file (i.e., the output of AstroDrizzle using build=False).   I think it is probably a general enough case that it warrants adding an extra check to catch and correct.   

https://github.com/spacetelescope/drizzlepac/blob/0fe80f657f3830234a5b232d07f5b864cc17f67b/drizzlepac/catalogs.py#L445

mcara:
> I've been tripping on a bug that manifests when the wcs object has `self.wcs.extname=('',1)`

I am curious, how can this be? I do not believe our software should be creating this kind of files. This just seems wrong from the [FITS standard](https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf) point of view:

> **EXTVER keyword**. The value field shall contain an integer to be used to **distinguish among different extensions in a *FITS* file with the *same type and name***, i.e., the same values for `XTENSION` and `EXTNAME`.

In my reading of the above, an `EXTNAME` is required for the `EXTVER` to make sense and to be present. 

I think the real bug is in https://github.com/spacetelescope/stwcs/blob/b7c7aaa96c9c3c9dd8c9e93dbdbf90996dfe52c8/stwcs/wcsutil/hstwcs.py#L156-L157 that allowed this to happen/be supported.

If your primary header does not have `EXTNAME` at all but it does have `EXTVER=1`, does

```python
w = stwcs.wcsutils.HSTWCS('filename.fits', ext=('', 1))
```
even work? Does it load the correct WCS?

Also, if using `astropy.io.fits`, can it find extension `('', 1)` for such a file:

```python
from astropy.io import fits
h = fits.open('filename.fits')
h[('', 1)].header
```

I doubt it works. I do not think we should be supporting non-standard FITS files.

CC: @stsci-hack

srodney:
Well, I don't disagree.   To answer your questions @mcara : 
This does work: 

`w = stwcs.wcsutils.HSTWCS('filename.fits', ext=('', 1))`

But this does not: 

```
[...]

PR #591, opened 2020-02-12
new check to catch wcs from unnamed primary hdu with extver=1

srodney:
Addresses issue #590

### DIFF
--- a/drizzlepac/catalogs.py
+++ b/drizzlepac/catalogs.py
@@ -442,7 +442,7 @@ class ImageCatalog(Catalog):
         super().__init__(wcs, catalog_source, **kwargs)
         extind = self.fname.rfind('[')
         self.fnamenoext = self.fname if extind < 0 else self.fname[:extind]
-        if self.wcs.extname == ('',None):
+        if (self.wcs.extname == ('',None) or self.wcs.extname == ('',1)):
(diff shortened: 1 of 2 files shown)


stsci-hack:
A more fundamental solution has been implemented in #954, where astrodrizzle correctly populates the headers with EXTNAME and EXTVER in all cases.  This admittedly does not help with data already processed, except to allow the users to reprocess their drizzled products to clean up the headers and finally make them FITS-compliant.  This PR should now either be closed without merging to rely on #954 or rebased onto the latest version of the code so that it can finally be merged in.

mcara:
Alternative fix that would create a WCS object with the correct `extname` property (`==` a value that would allow `hdulist[extname]` to work as expected) is in https://github.com/spacetelescope/stwcs/pull/179. That PR fixes the underlying issue.

mcara:
Alternative fix that would create a WCS object with the correct `extname` property (`==` a value that would allow `hdulist[extname]` to work as expected) is in https://github.com/spacetelescope/stwcs/pull/179. That PR fixes the underlying issue.

CC: @stsci-hack

mcara:
fixed via https://github.com/spacetelescope/stwcs/pull/179

PR #592, opened 2020-02-14
Segmented Catalog Updates

mdlpstsci:
Modified the segmented catalog to address bugs, make improvements, and to align better with the way the Point catalog is produced.
- Use option "columns" on to_table() method so only desired values are computed.
- Need to provide both the segmented image, as well as the total detection catalog to
  the filter product object.
- Change some methods to modify the passed table parameter object as if "passed by
  reference" rather than return a new table.
- Use list indices for addressing column "good" or "bad" rows.
- Put X-/Y-centroids and RA/Dec coordinates, as captured from the total detection
  table, for corresponding segments in the filter catalog which could not be measured.
- Create columns of the output filter table and fill values with nans.  For any
  measurments that cannot be made, a nan will be present.

### DIFF
--- a/drizzlepac/hlautils/catalog_utils.py
+++ b/drizzlepac/hlautils/catalog_utils.py
@@ -8,7 +8,7 @@ import copy
 import astropy.units as u
 from astropy.io import fits as fits
 from astropy.stats import sigma_clipped_stats
-from astropy.table import Column, MaskedColumn, Table, join
+from astropy.table import Column, MaskedColumn, Table, join, vstack
 from astropy.coordinates import SkyCoord
 import numpy as np
 from scipy import ndimage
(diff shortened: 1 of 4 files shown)


drizzlepac/hlautils/catalog_utils.py
```diff
@@ -824,6 +825,13 @@ def __init__(self, image, param_dict, param_dict_qc, diagnostic_mode, tp_sources
         self._border = self.param_dict["sourcex"]["border"]
         self._nsigma = self.param_dict["nsigma"]
 
+        # Columns to include from the computation of source properties to save
+        # computation time from computing values which are not used
+        self.include_filter_cols = ['background_at_centroid', 'bbox_xmax', 'bbox_xmin', 'bbox_ymax', 'bbox_ymin',
```
stsci-hack: This list seems a little too spartan.  Properties such as 'area' hold obvious use for the end-user and correspond to a column provided by the HLA catalog.  Granted, the point of this was to limit the computations, but we also do not want to end up leaving out information valuable for the end-user.  I would suggest that this list be reviewed by key stakeholders prior to merging to insure the proper balance.

mdlpstsci: I only used the values generated by Photutils which look like they correspond to the values generated in the Source Extractor catalog for the HLA project.  Note there are additional columns which contain photometry data as produced by the photometry_tools.iraf_style_photometry().  I will ask users for their input.

mdlpstsci: Email requesting input on further values to be included in the Segmentation catalog has been sent to: Jenn, Varun, Roberto, Matt, Hien, Rick, and Steve.

mdlpstsci: Until more requests, added "area" and "source_sum_error" (aka FluxIsoError).

PR #593, opened 2020-02-18
Fix HSTDP 2019.5.1 bugs

stsci-hack:
These commits fix numerous bugs reported as a result of initial reprocessing after installing HSTDP 2019.5.1 (quick-fix).   Multiple bugs will be addressed starting with:

- **jcwr03030** : Processing encountered recursion error (exceeded maximum limit) caused by FWHM being measured for 'sharpest' focus object in field that was too close to the edge of a chip.  
- **j8uw01020** : One (of the 2) exposures in the association had NGOODPIX=0, yet it was still being passed to AstroDrizzle() as if it was valid, causing the wrong parameters to be set for processing and resulting in 'too many pixels rejected' error. 

The datasets listed with each error not only the datasets reported in reprocessing as having errors, but were also used to demonstrate the original problem for debugging and used to verify that the problem was corrected by the code changes.  Additional bug fixes will be included in this PR in later commits as they are developed for the known list of error cases from reprocessing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ The following notes provide some details on what has been revised for each
 version in reverse chronological order (most recent version at the top
 of the list).
 
-3.1.5 (unreleased)
+3.1.7 (unreleased)
 ==================
 
 - Fix a crash in ``tweakreg`` when ``2dhist`` is enabled and ``numpy``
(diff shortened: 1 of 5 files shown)


stsci-hack:
- **icrw02010** : Reprocessing ran into 'small array too large' error. This was also fixed by same commit that fixed the problem with **jcwr03030**.

stsci-hack:
- jcx552010 : One of the SBC images reports as good (EXPTIME and NGOODPIX > 0), yet is blank.  This is now treated in 'max_overlap_diff' as a 'bad' measure allowing this ASN to process.

stsci-hack:
Commit b07ce98 addresses the failure to process idn117010 caused by a typo in computing the overlap regions.  This bug showed up due to the combined image having a range in X >> range in Y.

stsci-hack:
Commit 4bca0da and 56d257f address problems with the 2D histogram needing integer input values when they are originally defined as floats.

stsci-hack:
Commit 53c7297 addresses problems with table generation when no sources are found.

stsci-hack:
Commit [f6585a4](https://github.com/spacetelescope/drizzlepac/pull/593/commits/f6585a4e37aa761286c66a01562ee9e4bfb36d93) includes:
  - Changes made for #594 for making sure alignment processing can be turned on/off using environment variables
  -  Fix logic that did not pass along FLCs to final drizzle processing after alignment

stsci-hack:
- jcx552010 : One of the SBC images reports as good (EXPTIME and NGOODPIX > 0), yet is blank.  This is now treated in 'max_overlap_diff' as a 'bad' measure allowing this ASN to process.

PR #616, opened 2020-03-11
Fix various bugs

stsci-hack:
Update the alignment_viewer module to produce JSON products as an additional code-readable summary of the processed datasets.

### DIFF
--- a/drizzlepac/align.py
+++ b/drizzlepac/align.py
@@ -29,9 +29,9 @@ from .hlautils import config_utils
 __taskname__ = 'align'
 
 MIN_CATALOG_THRESHOLD = 3
-MIN_OBSERVABLE_THRESHOLD = 10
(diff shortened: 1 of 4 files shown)


stsci-hack:
Changes to the pipeline alignment were also included to address cases where data should have aligned to GAIA or recognized as being a successful apriori solution.  These updates to the pipeline alignment were verified to successfully result in apriori or aposteriori solutions for:
    - icz006010
    - ibfuw1080
    - j9fj85080
    - iczg12030
    - ic1r37010
    - id9m60fuq
    - iblf01d8q
    - ibsl31030

In addition, the alignment_viewer code was modified to include more detailed information about the observations in the JSON summary file to allow for better determination of whether or not the observation SHOULD be aligned.  This information includes keywords (primarily from the first row/image of the HDRTAB):  obstype, fgslock, mtflag, filters, and obsmode, amongst other keywords.

PR #619, opened 2020-03-20
Insure background subtraction is consistently applied

stsci-hack:
It turned out that background subtraction was not being done consistently for each of the iterations of the drizzling during standard pipeline processing.  This led to images which are mostly blank to fail similarity comparison due to differing background levels resulting in many datasets not getting updated WCS solutions when they were appropriate.  

In addition, datasets which can not be aligned in the pipeline with an aposteriori solution (such as prism or grism datasets), were not being handled correctly and an unnecessary exception was being thrown.  This also prevented the apriori solution from being applied in some cases where it should have been.

These changes address both of these issues and were tested on ALL 209 datasets that only had pipeline default solutions (or NONE) from the 2020-03-19 test run, with a majority of them now getting improved WCS solutions.

### DIFF
--- a/drizzlepac/align.py
+++ b/drizzlepac/align.py
@@ -226,28 +226,33 @@ def perform_align(input_list, archive=False, clobber=False, debug=False, update_
         else:
             log.warning("WARNING: Unable to display Git repository revision information.")
 
+    # Initialize key variables
+    filtered_table = None
+
+    # 1: Interpret input data and optional parameters
+    log.info("{} STEP 1: Get data {}".format("-" * 20, "-" * 66))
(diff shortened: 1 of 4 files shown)


stsci-hack:
Testing on all 209 'failed' results from the 2020-03-19 astrometry testing done under Condor resulted in these changes allowing all datasets to get either apriori or aposteriori where conditions in the observations allowed.  Only those which were 'functionally blank' (single source or only faint extended source(s)), completely blank(including EXPTIME=0 data), moving target obs and HRC/ACQ did not get updated WCS solutions.

PR #621, opened 2020-03-27
Fix fitmask exceptions for blank images 

stsci-hack:
This addresses problems seen in extended testing of 1000 datasets where an Exception gets thrown when 1 image in a set did not have a successful relative fit while the others did.  This showed up as the Exception:
`
KeyError: "fitmask"
`
In addition, problems with blank images not generating a source catalog was causing 'amutils.build_wcscat' to throw the following Exception:
`  
File "/home/mburger/anaconda3/envs/2019.5.clone/lib/python3.6/site-packages/drizzlepac/hlautils/astrometric_utils.py", line 559, in extract_point_sources
    if high_sn is not None and len(srcs) > high_sn:
TypeError: object of type 'NoneType' has no len()
`

Both of these issues were corrected in this PR based on testing of ibc305yvq, iaac240d0, ibg414010, and ibh707010.

### DIFF
--- a/drizzlepac/hlautils/align_utils.py
+++ b/drizzlepac/hlautils/align_utils.py
@@ -686,9 +686,11 @@ def interpret_fit_rms(tweakwcs_output, reference_catalog):
     for group_id in group_ids:
         input_mag = None
         for item in tweakwcs_output:
+            tinfo = item.meta['fit_info']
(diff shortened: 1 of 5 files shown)


stsci-hack:
This also includes changes to use the source finding routine from 'tweakreg' to generate the sources for the quality_analysis module, instead of relying on IRAFStarFinder.

PR #622, opened 2020-03-27
Insure src catalog always has same number of columns

stsci-hack:
For datasets with no detectable point-sources, the quality analysis code ended up getting only 4 empty columns returned by the source finding code 'tweakutils.ndfind'.  This resulted in the following Exception being thrown as detected through testing:
`
File "/home/mburger/anaconda3/envs/caldp/lib/python3.6/site-packages/drizzlepac/hlautils/astrometric_utils.py", line 561, in extract_point_sources
    nbright=nbright)
ValueError: not enough values to unpack (expected 7, got 4)
`

This simple fix insures that the returned values always have the same number of values to unpack (7).  This fix was verified using ibh705w8q (original error case) and iaac240d0 (case with detected sources).

### DIFF
--- a/drizzlepac/hlautils/astrometric_utils.py
+++ b/drizzlepac/hlautils/astrometric_utils.py
@@ -558,7 +558,8 @@ def extract_point_sources(img, dqmask=None, fwhm=3.0, kernel=None,
     x, y, flux, src_id, sharp, round1, round2 = ndfind(img, 
                                                      sigma*threshold, 
                                                      fwhm, bkg[1],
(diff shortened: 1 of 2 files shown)


stsci-hack:
The Exception handling for aposteriori alignment also needed to be cleaned up just to make it clear there was no fatal Exceptions, merely a failure to align.  The log messages were updated in the latest commit (c1ae2e9) in runastrodriz to clarify that everything was handled correctly and no Exception was triggered that needs to be evaluated.

PR #629, opened 2020-04-09
Compare RA/Dec of cross-matches sources in Point and Segment Catalogs

dulude:
* See [Jira ticket HLA-293](https://jira.stsci.edu/browse/HLA-293)

**Summary**
The new function _svm_quality_analysis.compare_ra_dec_crossmatches()_ finds sources common to both the point and segment catalogs, and uses the matched RA and Dec values. The function computes the on-sky seperation of the cross-matched sources and reports both unclipped and 3x3 sigma-clipped statistics on the separation values, along with details of the cross-match and catalogs containing the cross-matched point and segment RA and dec values.

**Code changes/updates**
- drizzlepac/hlautils/svm_quality_analysis.py
     - added compare_ra_dec_crossmatches() function
- drizzlepac/hlautils/diagnostic_utils.py
      - modified code so that hap_obj.primary_header info is not permanently modified
- drizzlepac/hapsequencer.py
      - added call to execute svm_quality_analysis.compare_ra_dec_crossmatches()


drizzlepac/hlautils/svm_quality_analysis.py
```diff
@@ -132,7 +140,134 @@ def compare_num_sources(catalog_list, drizzle_list, log_level=logutil.logging.NO
...
+    matching_lines_ref, matching_lines_img = csl.getMatchedLists(slNames, imgNames, slLengths, log_level=log_level)
+    json_results_dict['number of cross-matches'] = len(matching_lines_ref)
+    if len(matching_lines_ref) == 0 or len(matching_lines_img) == 0:
+        log.critical("*** Comparisons cannot be computed. No matching sources were found. ***")
+        return ("ERROR")
+    # 2: Create masks to remove missing values or values not considered "good" according to user-specified good bit values
+    # 2a: create mask that identifies lines any value from any column is missing
+    missing_mask = csl.mask_missing_values(point_data, seg_data, matching_lines_ref, matching_lines_img, columns_to_compare)
+    # 2b: create mask based on flag values
+    matched_values = csl.extractMatchedLines("FLAGS", point_data, seg_data, matching_lines_ref, matching_lines_img)
```
mdlpstsci, line 206: The column name in the *ecsv file is "Flags".  There is no checking for case in the code, so this line will not work as expected.

drizzlepac/hlautils/svm_quality_analysis.py
```diff
@@ -132,7 +140,134 @@ def compare_num_sources(catalog_list, drizzle_list, log_level=logutil.logging.NO
...
+    log.info("\n")
+    # 2: Run starmatch_hist to get list of matched sources common to both input sourcelists
+    slLengths = [len(point_data['RA']), len(seg_data['RA'])]
+    json_results_dict['point catalog length'] = slLengths[0]
+    json_results_dict['segment catalog length'] = slLengths[1]
+    matching_lines_ref, matching_lines_img = csl.getMatchedLists(slNames, imgNames, slLengths, log_level=log_level)
+    json_results_dict['number of cross-matches'] = len(matching_lines_ref)
+    if len(matching_lines_ref) == 0 or len(matching_lines_img) == 0:
+        log.critical("*** Comparisons cannot be computed. No matching sources were found. ***")
+        return ("ERROR")
```
mdlpstsci, line 201: The function call in hapsequencer does not expect a return value.  I do not think it is a critical error to not have any matches as this is diagnostic information, but we perhaps should discuss this as a team.  I would add the names of the catalogs to the output message to be more informative.

drizzlepac/hlautils/svm_quality_analysis.py
dulude: This is true, but it's not a problem. I wrote compare_sourcelists.py to handle different types of sourcelists (.ecsv hap sourcelists, and .sexphot.txt and daophot.txt HLA sourcelists) which have different names for the same columns (like "Flags" vs. "FLAGS). Compare_sourcelists.py has code that determines the sourcelist type, and maps the columns in the sourcelist to a set of in-program column names that are always the same, regardless of sourcelist type, so it's always an apples to apples comparison.
"FLAGS" as seen on line 206 is the in-program column name. All flag data point_data and seg_data is stored under this name, so it's actually working the way it should.

mdlpstsci: ahhh.

drizzlepac/hlautils/svm_quality_analysis.py
dulude: Issue resolved by commit [62fdb04](https://github.com/spacetelescope/drizzlepac/pull/629/commits/62fdb043854d67cf0a9f6c915ddc8c2d7a97deb4)

Issue #630, opened 2020-04-10
Initialize align_dicts variable before usage


mdlpstsci:
In testing a bug was raised that the align_dicts (runastrodriz.py) was not initialized before use and caused the routine to return an error code to OWL.  

> 16:07:33 (08-Apr-2020): Finished processing ib8t01010_pipeline_asn.fits
> 2020099160733-I-----astrodrizzle completed -------------------------------------
> 
> local variable 'align_dicts' referenced before assignment
> ERROR: Cannot run astrodrizzle on -i.
> 2020099160733-I-INFO-md_xxx_OWL.sh-runastrodriz -i  ib8t01010_asn returned 1
> Deactivated calibration software env caldp-master
> OWL python /ifs/pipeline/dms/rhel7/pkgs/miniconda3/envs/owl-2020.1.0 re-established.

It is true the align_dicts variable is not initialized, but it seems the problem still never should have been seen under the circumstances the data was being processed.

> 2020099160719-I-INFO-md_xxx_OWL.sh-ASTROMETRY_APPLY_APRIORI and ASTROMETRY_COMPUTE_APOSTERIORI are being set to OFF
> 2020099160719-I-INFO-md_xxx_OWL.sh-Active calib s/w env using 'source calib_sw_env_on.sh'
> Activated calibration software env caldp-master
> ...
> 2020099160720-I-INFO-md_xxx_OWL.sh-runastrodriz -i  ib8t01010_asn
> 16:07:25 (08-Apr-2020): Calibration pipeline processing of ib8t01010_asn started.
> *** astrodrizzle Processing Version 2.2.0(06-Mar-2020)***

In evaluating this issue, I was not able to reproduce the bug.    While the variable can be initialized, it should be understood how this bug was raised.  This needs to be fixed as it will hold up testing in DMS.

mdlpstsci:
The error is raised when both of the astrometry environment variables are set to OFF, ASTROMETRY_APPLY_APRIORI and ASTROMETRY_COMPUTE_APOSTERIORI, and quality analysis environment variable is set to ON, PIPELINE_QUALITY_TESTING.   Error is here:

```
if qa_switch:
        # Generate quality statistics for astrometry if specified
        calfiles = _calfiles_flc if _calfiles_flc else _calfiles
        json_file = qa.run_all(inFile, calfiles, fit_dict=align_dicts)
 
```

PR #632, opened 2020-04-14
Reset initialization for pipeline QA test

stsci-hack:
The QA test for standard pipeline processing originally passed in the alignment dictionary generated during processing.  However, in cases where all astrometry updates get turned off, this dict is uninitialized.  Furthermore, it turned out the QA test did not use that input in the first place (planned for, but not used).  

This set of changes addresses both situations as documented in JIRA ticket [HLA-317](https://jira.stsci.edu/browse/HLA-317).


stsci-hack:
This resolves #630 and JIRA ticket HLA-317.

PR #635, opened 2020-04-15
Enable control over fitgeom in alignment

stsci-hack:
Not all datasets have enough (good quality) sources to perform a full 'general' fit, which is the default for tweakwcs.  These changes expose control over the 'fitgeom' parameter in `tweakwcs.align_wcs()` to enable the code to set the value instead of relying solely on the default value.  Additional PRs/tickets will be needed to describe work to define the value of 'fitgeom' for each instrument if that turns out to be necessary.

The SVM processing was updated to set the value of 'fitgeom' to 'rscale' for all SVM processing under the assumption that higher order fitting done in standard pipeline processing has accounted for any skew in the observations being combined in SVM processing.   

This addresses [JIRA Ticket HLA-316](https://jira.stsci.edu/browse/HLA-316).

### DIFF
--- a/drizzlepac/hlautils/align_utils.py
+++ b/drizzlepac/hlautils/align_utils.py
@@ -174,6 +174,7 @@ class AlignmentTable:
     def configure_fit(self):
         # Convert input images to tweakwcs-compatible FITSWCS objects and
         # attach source catalogs to them.
+        
 
         self.imglist = []
         for group_id, image in enumerate(self.process_list):
@@ -195,8 +196,11 @@ class AlignmentTable:
         """
         return list(self.fit_methods.keys())
 
-    def perform_fit(self, method_name, catalog_name, reference_catalog):
+    def perform_fit(self, method_name, catalog_name, reference_catalog, 
+                    fitgeom='general'):
         """Perform fit using specified method, then determine fit quality"""
+        # Updated fits_pars with value for fitgeom
+        self.fit_pars[method_name]['fitgeom'] = fitgeom
         imglist = self.fit_methods[method_name](self.imglist, reference_catalog,
                                                 **self.fit_pars[method_name])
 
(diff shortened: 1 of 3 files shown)


stsci-hack:
This was tested on SVM datasets iboa31 and id5o27 to verify that the code uses 'rscale' fitting as expected and that the 'rscale' fit resolved the problem with skewed output products seen when using the default 'general' fit.

drizzlepac/hlautils/align_utils.py
```diff
@@ -195,8 +196,11 @@ def get_fit_methods(self):
         """
         return list(self.fit_methods.keys())
 
-    def perform_fit(self, method_name, catalog_name, reference_catalog):
+    def perform_fit(self, method_name, catalog_name, reference_catalog, 
+                    fitgeom='general'):
         """Perform fit using specified method, then determine fit quality"""
+        # Updated fits_pars with value for fitgeom
+        self.fit_pars[method_name]['fitgeom'] = fitgeom
```
mdlpstsci, line 203: There should be an informational message here that a value in a parameter file driving functionality has been overridden by a new value.

drizzlepac/hlautils/product.py
```diff
@@ -265,7 +265,7 @@ def align_to_gaia(self, catalog_name='GAIADR2', headerlet_filenames=None, output
                     align_table.perform_fit(method_name, catalog_name, ref_catalog)
                     align_table.select_fit(catalog_name, method_name)
                     align_table.apply_fit(headerlet_filenames=headerlet_filenames,
-                                         fit_label=fit_label)
+                                         fit_label=fit_label, fitgeom=fitgeom)
```
mdlpstsci, line 268: OK.  I guess I am crazy.  I guess the perform_fit() here does not need the fitgeom in this particular invocation (?), but I do not see that apply_fit() even accepts a fitgeom parameter, and it does not do anything with the value.  Do I just not have the Python voodoo?

drizzlepac/hlautils/align_utils.py
stsci-hack: Good idea.   Implemented in e78adcc.

drizzlepac/hlautils/product.py
stsci-hack: No, you were not going crazy.  My edits happened over 2 days and I overlooked the mismatch on where this parameter was getting set.  This has been corrected in e78adcc.

Issue #631, opened 2020-04-10
Initialize variable rv before use


mdlpstsci:
When trying to run the "runsinglehap" processing with the SVM diagnostic tests activated, an issue was raised due to an uninitialized variable.  The variable "rv" in get_git_rev_info.py needs to be initialized.

PR #638, opened 2020-04-17
Ensure return variable is initialized.

mdlpstsci:
Initialize the return variable of the function get_rev_id() so the code does not crash.

### DIFF
--- a/drizzlepac/hlautils/get_git_rev_info.py
+++ b/drizzlepac/hlautils/get_git_rev_info.py
@@ -92,6 +92,10 @@ def get_rev_id(local_repo_path):
     full git revision ID of the specified repository if everything ran OK, and "FAILURE" if something went
     wrong.
     """
+
+    # Ensure the return variable is initialized
+    rv = "FAILURE: git revision info not found"
+
     start_path = os.getcwd()
     try:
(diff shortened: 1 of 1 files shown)


drizzlepac/hlautils/get_git_rev_info.py
```diff
@@ -92,6 +92,10 @@ def get_rev_id(local_repo_path):
     full git revision ID of the specified repository if everything ran OK, and "FAILURE" if something went
     wrong.
     """
+
+    # Ensure the return variable is initialized
+    rv = "FAILURE: git revision info not found"
```
stsci-hack, line 97: This specific change looks to be just what was needed.  However, it duplicates line 111 which could just be replaced with a 'pass' statement instead.  Also, the lone 'raise' in line 109(?) is not strictly proper Python based on Section 8.4 of the Python documentation.  Instead, it should be something like 'raise ValueError("Git version not found")'.  This code will work, but it does not strictly follow good Python practice as it is now.

mdlpstsci: Addressed issues.

stsci-hack:
Closes #631

Issue #645, opened 2020-04-21
hapsequencer.py not running svm_quality_analysis.characterize_gaia_distribution()


dulude:

dulude:
I simply forgot to call it in hapsequencer.py

PR #636, opened 2020-04-16
Add SVM Q+A function to statistically characterize the distribution of GAIA sources identified within a given HAP final product image footprint

dulude:
* See [JIRA ticket #311 ](https://jira.stsci.edu/browse/HLA-311)

**Summary**
- Added function drizzlepac/hlautils/svm_quality_analysis.characterize_gaia_distribution() which computes and writes the following statistical measures about the distribution of GAIA sources in the image footprint to a json file:
    - Number of GAIA sources
    - X centroid location
    - Y centroid location
    - X offset of centroid from image center
    - Y offset of centroid from image center
    - X standard deviation
    - Y standard deviation
    - minimum closest neighbor distance
    - maximum closest neighbor distance
    - mean closest neighbor distance
    - standard deviation of closest neighbor distances
- drizzlepac/hlautils/svm_quality_analysis.py: Moved high-level code dealing with GAIA catalog creation to a new function generate_gaia_catalog(). This streamlines things a bit, as functions find_gaia_sources() and characterize_gaia_distribution() can now call the same piece of code to get GAIA catalogs.
- Added function drizzlepac/hlautils/astrometric_utils.get_catalog_from_footprint() which only identifies GAIA sources within a specified footprint (a list of RA, Dec coords that define the  corners of the bounding box)

PR #646, opened 2020-04-21
Add call to execute svm_quality_analysis.characterize_gaia_distribution() in hapsequencer.py

dulude:
See Issue #645. 
I realized after merging PR #636 that I had never added a call to execute svm_quality_analysis.characterize_gaia_distribution() in hapsequencer.py. This PR fixes the issue.


stsci-hack:
These changes only enable the analysis code to be run from the command line in a hard-coded manner based on the settings of True/False which can be useful for some applications (e.g., dev/debugging).  It is not clear how this code, though, would be called from within hapsequencer.  As a result, I would suggest turning the code now in the '__main__' section at the end into it's own callable function which would also allow the end-user/pipeline to call it with a custom set of quality tests turned on instead of hard-coding them as they are currently.

dulude:
> These changes only enable the analysis code to be run from the command line in a hard-coded manner based on the settings of True/False which can be useful for some applications (e.g., dev/debugging). It is not clear how this code, though, would be called from within hapsequencer. As a result, I would suggest turning the code now in the '**main**' section at the end into it's own callable function which would also allow the end-user/pipeline to call it with a custom set of quality tests turned on instead of hard-coding them as they are currently.

Hey @stsci-hack, I implemented your suggestions in commit [679524b](https://github.com/spacetelescope/drizzlepac/pull/646/commits/679524b044ef2f96b469a58485bc3accd322b29d)

Issue #644, opened 2020-04-21
Need updates to DGEO warning message


samanthalh:
Hello! I'm working on a WFPC2 DrizzlePac ticket (INC0154128, for reference) and the user asked about the lengthy warning message below. I feel this needs to be updated for two reasons: 1) it provides users with an IRAF hedit command, and IRAF is deprecated, and 2) it should explain that for WFPC2 and other archival instruments, it is **fine** and **expected** to type 'c' and continue running drizzle. I don't believe that's clear at that moment. Thanks!!

> 
> A 'DGEOFILE' keyword is present in the primary header but 'NPOLFILE' keyword was not found.
> This version of the software uses a new format for the residual distortion DGEO files.
> Please consult the instrument web pages for which reference files to download.
> A small (new style) dgeofile is needed ('_npl.fits' extension) and possibly a
> detector to image correction file ('_d2i.fits' extension).
> The names of these files must be added to the primary header either using the task XXXX
> or manually, for example:
> 
> hedit u2pq0505t_c0m.fits[0] npolfile fname_npl.fits add+
> hedit u2pq0505t_c0m.fits[0] d2imfile fname_d2i.fits add+
> 
> where fname_npl.fits is the name of the new style dgeo file and fname_d2i.fits is
> the name of the detector to image correction. After adding these keywords to the
> primary header, updatewcs must be run to update the science files:
> 
> from stwcs import updatewcs
> updatewcs.updatewcs("u2pq0505t_c0m.fits")
> 
> Alternatively you may choose to run astrodrizzle without DGEO and detector to image correction.
> 
> To stop astrodrizzle and update the dgeo files, type 'q'.
> To continue running astrodrizzle without the non-polynomial distortion correction, type 'c':

stsci-hack:
@samanthalh Please check out #647 to see whether this will make the warning more appropriate for the users.

stsci-hack:
Resolved with #647 on 22-Apr-2020.

PR #647, opened 2020-04-22
Update wfpc2 warning

stsci-hack:
This should adequately address the concerns raised in #644.  The warning message was rephrased to direct users on the astropy.io.fits syntax for adding the NPOL and D2IM reference filenames to the image header.  The options were also rephrased to make it less troubling to type 'c' to continue if they have not manually added the reference files or run updatewcs.  This leaves the user open, though, to a greater chance that the code will crash.

### DIFF
--- a/drizzlepac/processInput.py
+++ b/drizzlepac/processInput.py
@@ -1065,17 +1065,19 @@ def checkDGEOFile(filenames):
 
     msg = """
             A 'DGEOFILE' keyword is present in the primary header but 'NPOLFILE' keyword was not found.
-            This version of the software uses a new format for the residual distortion DGEO files.
+            This version of the software uses the common format 'NPOLFILE' that replaced
+            the origina DGEO distortion files which describe the for the residual distortion.
             Please consult the instrument web pages for which reference files to download.
-            A small (new style) dgeofile is needed ('_npl.fits' extension) and possibly a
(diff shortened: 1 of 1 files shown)


stsci-hack:
Will close #644.

rjavila:
Should there be a specific mention that this message can be ignored for WFPC2 images?

stsci-hack:
@samanthalh  @rjavila This message is essentially for no one else but WFPC2 users since STIS, ACS and WFC3 data do not use the DGEOFILE.  Some WFCP2 datasets do not have a DGEOFILE specified (value = 'N/A'), while others do.  For those data datasets with DGEOFILE = 'N/A', the user can hit 'c' and it would work (most of the time).   However, for those datasets where the 'DGEOFILE' is defined, the user should be given the choice to use that calibration, which is what this message does.  Now, I could (most likely) change the code to not bring up this message at all if 'DGEOFILE' is 'N/A', although an NPOLFILE may still be define for that data.  Otherwise, you are simply suggesting removing this warning altogether without providing the user any guidance on how to correctly apply the distortion calibrations available to their data, which could be done.  Please confirm what you would like to be done.

stsci-hack:
Since there was no response to my question about how to finish PR, I will close this until they raise it again as an issue.

PR #655, opened 2020-05-01
Resolve HSTDP2020.2 WCS exceptions

stsci-hack:
HSTDP 2020.2 testing encountered exceptions being generated, and worked around, when pipeline processing j9l929010 and 9 other datasets.  It turns out that these datasets still have incomplete IDC_* solutions in the astrometry database, and trying to restore one of them as the active PRIMARY WCS caused subsequent processing to throw an Exception (missing wcs.idcscale).  

These changes address this issue as well as tries to make the names for the new aposteriori headerlets consistent with what is stored in the database.


stsci-hack:
This will also need to be pulled onto the release branch used for 2020.2 (3.1.8x?).

PR #661, opened 2020-05-07
Prototype Bokeh graphic to support SVM quality assessment data

mdlpstsci:
Prototype graphics generation code for the SVM quality assessment using Bokeh.  This module is specifically designed for reading the photometry data and generating two plots, and also incorporates a preliminary JSON harvester written by Warren.  This module serves as a "proof of concept" experiment and is addressing JIRA HLA-322.

### DIFF
--- a/drizzlepac/hlautils/photometry_graphics.py
+++ b/drizzlepac/hlautils/photometry_graphics.py
@@ -1,0 +1,112 @@
+# This routine is designed specifically to support plotting the
+# SVM data quality output.  As such, it expects the input Pandas
+# Dataframe to have a particular structure and the data to contain
+# specific measurements.
+# This is NOT a general plotting utility.
+#
+# SVM Photometry file has the following data values where
+# differences are POINT - SEGMENT.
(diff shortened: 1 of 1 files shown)


drizzlepac/hlautils/photometry_graphics.py
```diff
@@ -0,0 +1,263 @@
...
+from bokeh.plotting import figure, output_file, show
+from bokeh.models import ColumnDataSource, Label
+import csv
+import glob
+import json
+import os
+import pandas as pd
+
+
+def compute_global_stats(value_array):
```
stsci-hack, line 48: Style comment: 'compute_global_stats' implies the computation of multiple statistics terms, not just the mean.  This is confusing unless it was intended to serve as place-holder for computing additional stats later (std, median, ...).  No changes needed here, but it would be useful to keep this in mind in the final implementation.

mdlpstsci: It is meant to be a placeholder to compute multiple stats.  The function should have a better description or a better name depending upon the circumstances!

Issue #634, opened 2020-04-15
Remove unnecessary files from SVM manifest


mdlpstsci:
In the current SVM manifest the names of all the _potential_ headerlet filenames are included.  This is a mistake in that the manifest should only contain the names of files actually generated during the SVM processing.  This file is used by DSB to ensure all the files listed in the manifest are present on disk.

If the dataset is unable to apply a correction to align to the absolute astrometric reference frame of GAIA, then no headerlet file will be created.

PR #664, opened 2020-05-13
Initial fix for manifest file

stsci-hack:
This simple update replaces the expected list of headerlet filenames that currently gets added to the manifest file with the list of actually generated headerlet filenames.   Initial testing indicates it does not interfere with generating the manifest file when all expected headerlets are written out (verified using iboa31).


stsci-hack:
Merging this will close #634 .

stsci-hack:
Testing with visit iacs01' demonstrates that these changes results in a manifest file which only contains the names of headerlet files which were actually written to disk (2 instead of the 3 expected files).

PR #667, opened 2020-05-15
Add json harvester script

dulude:
- See [JIRA ticket 314](https://jira.stsci.edu/browse/HLA-314).
- drizzlepac/hlautils/diagnostic_json_harvester.py:
   - Initial functional commit.
   - This script extracts all the information stored in the json files produced by drizzlepac/hlautils/svm_quality_analysis.py stores it in a Pandas DataFrame, and writes the final Pandas DataFrame to a comma-seperated .csv file.
   - The information is organized by image name. Each row contains all information generated by all svm_quality_analysis tests run on, or related to a particular image, so there is typically information from multiple .json files per row.
- drizzlepac/hapsequencer.py:
   - removed unneeded call to svm_quality_analysis.report_wcs().
- drizzlepac/devutils/comparison_tools/compare_sourcelists.py:
   - couple of bug fixes.


drizzlepac/hapsequencer.py
```diff
@@ -484,9 +484,6 @@ def run_hap_processing(input_filename, diagnostic_mode=False, use_defaults_confi
             log.info("SVM Quality Assurance statistics have been requested for this dataset, {}.".format(input_filename))
             svm_qa.run_quality_analysis(total_obj_list, log_level=log_level)
 
-            # Compute WCS differences between the primary WCS and the alternates, except for OPUS WCS
-            svm_qa.report_wcs(total_obj_list)
-
         # Write out manifest file listing all products generated during processing
```
mdlpstsci, line 487: Just noting the removal of 487-489 has already been checked into master with PR#668.

drizzlepac/hlautils/diagnostic_json_harvester.py
```diff
@@ -24,11 +25,62 @@
...
+    Returns
+    -------
+    a version of input dictionary *dd* that has been flattened by one layer
+    """
+    return {prefix + separator + k if prefix else k: v
+            for kk, vv in dd.items()
+            for k, v in flatten_dict(vv, separator, kk).items()
+            } if isinstance(dd, dict) else {prefix: dd}
+
+# ------------------------------------------------------------------------------------------------------------
```
mdlpstsci, line 80: Just FYI - there is a Python package, flatten-dict, which will both _flatten_ dict-like objects and _unflatten_ those objects!

drizzlepac/hlautils/diagnostic_json_harvester.py
```diff
@@ -24,11 +25,62 @@
...
+
+
+def get_json_files(search_path=os.getcwd(), log_level=logutil.logging.INFO):
     """use grep to create a list of json files to harvest
```
mdlpstsci, line 85: Minor typo: You meant to say "glob" and not "grep" in the description.

drizzlepac/devutils/comparison_tools/compare_sourcelists.py
```diff
@@ -1149,8 +1149,8 @@ def comparesourcelists(slNames=None, imgNames=None, good_flag_sum = 255, plotGen
         # Get platescale
         plate_scale = wcsutil.HSTWCS(imgNames[0], ext=('sci', 1)).pscale
         matched_values = extractMatchedLines("X", refData, compData, matching_lines_ref, matching_lines_img, bitmask=bitmask)
-        if output_json_filename:  # Add matched values to diag_obj
-            diag_obj.add_data_item(matched_values,"X MATCHED VALUES")
+        if output_json_filename and matched_values.shape[1] > 0:  # Add matched values to diag_obj
+            diag_obj.add_data_item(Table([matched_values[0], matched_values[1]], names=('reference', 'comparison')),"X MATCHED VALUES")
             diag_obj.add_update_info_item("header", "plate_scale", plate_scale)
     if matched_values.shape[1] > 0:
         formalTitle = "X Position"
```
stsci-hack, line 1156: The following is a suggestion only, not a request for a change...  Having multiple copies of essentially the same code for each column being extracted comes with a couple of problems; namely, more places for errors to be made, more code for Python to interpret and more difficult to understand what is being done in the code.  It works, but it would be better to define a single function that could be called for each column ("X", "Y", ...) instead.  I could see 'extract_column(colname, formalTitle, input_json_filename, output_json_filename, regressionTestResults)' that returns the 'matchedValues' for plotting later.  I would also argue that 'colTitles' is simply 'regressionTestResults.keys()' and is not strictly necessary now that dicts are ordered, although defining 'regressionTestResults' as an OrderedDict to start with would resolve that as well.  This would let lines 1145-1164 be pulled into a single function that would then replace the same lines for all the subsequent columns.   A list of column names [('X','X Position'), ('Y', 'Y Position'),...] could then serve to call this function for all the columns, making it more obvious what is being done.  I may be overlooking some interaction between each of these sections of code, but initially this could eliminate 14 nearly identical sections of code if the unique handling of the outputs can be handled separately.  **PLEASE** consider whether this encapsulation is possible and **if not, disregard this comment!**

drizzlepac/hapsequencer.py
dulude: Noted!

drizzlepac/hlautils/diagnostic_json_harvester.py
dulude: oh cool! i'll have to look into that.

dulude: I'll incorporate this into my development towards ticket [HLA-343](https://jira.stsci.edu/browse/HLA-343).

drizzlepac/hlautils/diagnostic_json_harvester.py
dulude: Good Catch! Fixed in commit [1ac1a17](https://github.com/spacetelescope/drizzlepac/pull/667/commits/1ac1a175e9780ca9e692f1bf183aba0dff46221a)

drizzlepac/devutils/comparison_tools/compare_sourcelists.py
dulude: There's just enough complexity and non-uniqueness between different comparisons (X position, Y position, RA/DEC positions, etc.) that encapsulating this would be complicated. Best leave it alone for right now.

PR #669, opened 2020-05-19
Address problems from testing svm0.2.5

stsci-hack:
Single-visit processing tests using svm0.2.5 still resulted in some datasets showing problems.  One dataset which showed mis-alignment within each visit is 'j8cw03' and 'j8cw53' (both of the Tadpole galaxy).   This PR addresses the alignment problems reproduced interactively, along with fixing the logic for supporting user provided lists of filenames as inputs.  

Additional problems seen during testing will also be addressed in this PR as the individual issues are isolated.


stsci-hack:
A major change in the processing was implemented in 84f3ab5.  This commit examines the active WCS to see whether or not it is an aposteriori solution applied from the astrometry database, and replaces it with the 'best' true apriori solution; namely, (in order of priority), the HSC WCS, the GSC WCS, finally the IDC_* solution.  This allows the pipeline to completely recompute the aposteriori solution instead of simply starting from a previous solution, which included errors when first computed during the initial computation of the new astrometry solutions.  This new process was demonstrated to work using 'ie6d07030'.  This replacement only occurs when applying the apriori solutions in 'runastrodriz.py' pipeline processing and does not affect any other steps in the pipeline astrometry determination.

stsci-hack:
It turns out some of the problems with consistency of WCS keywords comes back to how STWCS applies the keywords from the headerlet as the primary WCS.  This will require a change in STWCS, with one suggested change being made in [STWCS PR #122](https://github.com/spacetelescope/stwcs/pull/122).  If this works, this code would require that updated version of STWCS, and the dependencies in drizzlepac would need to be updated to reflect this.

PR #670, opened 2020-05-20
Fix a crash when finding sources in very large images

mcara:
Source finding in very large images fails due to convolution image containing `NaN`s. This is due to a bug in `scipy.signal.convolve2d` - see https://github.com/scipy/scipy/issues/10761 

Both `scipy.signal.fftconvolve` and `scipy.ndimage.convolve` work correctly.

This PR replaces the call to `scipy.signal.convolve2d` with `scipy.ndimage.convolve`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,7 +17,10 @@ of the list).
 3.1.7 (unreleased)
 ==================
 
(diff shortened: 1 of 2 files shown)


mcara:
Thanks @stsci-hack @larrybradley

PR #673, opened 2020-05-29
Implement cross-correlation as default fitting option for SVM processing

stsci-hack:
Many observations are difficult to align due any number of factors, including lack of point sources or crowding or too many saturated sources or all of the above.  Registration based on cross-correlation can be used when there are problems with point-source based alignment.  

The changes here implement a new fitting function based on scikit-image's feature module to use cross-correlation to determine the relative alignment to within a 1pixel.  This initial correction then gets passed to 'tweakwcs' for final alignment with suitable input parameters (searchrad=1.0).  This new function has also been made the default fitting routine for single-visit mosaic (SVM) processing. 

Classification of sources in the alignment catalog also uses scikit-image's hough line transforms to look for linear sources and flags them as non-PSFs so they will be ignored when being selected for use in alignment.  

This algorithm still needs to be extensively tested, but does work well on ib3m53.

### DIFF
--- a/drizzlepac/align.py
+++ b/drizzlepac/align.py
@@ -349,6 +349,7 @@ def perform_align(input_list, archive=False, clobber=False, debug=False, update_
         best_fit_status_dict = {}
         best_fit_qual = 5
         best_fit_label = [None, None]
+        relax_consistency = False
         # create pristine copy of imglist that will be used to restore imglist back so it always starts exactly the same
         # for each run.
         orig_imglist = copy.deepcopy(imglist)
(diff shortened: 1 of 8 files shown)


stsci-hack:
**NOTE: This PR is NOT intended to be merged immediately.  Instead, it should be tested on most of the datasets identified by INS on the 'Single-Visit Mosaic Testing Datasets' confluence page before making any decision to merge.**

stsci-hack:
**DO NOT MERGE**:  This new method still does not work to correctly align data from j8cw03 or j8cw53 (Tadpole observations).  It is not clear why, so until this method can work on this data, the PR should NOT be merged.

stsci-hack:
This algorithm is not developed enough for integration in the pipeline environment at this time, so this ticket will be closed until it can be (if ever) be improved to be suitable for pipeline use.

PR #674, opened 2020-06-02
Add new parameter for pipeline use of custom rules files

stsci-hack:
Rules for how to blend the header keyword values for all the input exposures into a single header for the drizzle products are specified using a `rules_file` as defined in and by the `fitsblender` package.  These rules files, though, are only defined per instrument, which causes problems during pipeline processing of single-visit mosaics (SVM) as each visit copies a custom HAP rules file from the `drizzlepac/pars` directory into the local directory for use by `fitsblender`.  Unfortunately, with a generic name, the pipeline processing removed that file for each visit, as another visit was being processed in the same directory expecting to use it as well.  

These changes add a new parameter, `rules_file`, to AstroDrizzle and updated the `hapsequencer` and `processing_utils` modules to define dataset-specific names for the rules for local use.  This will allow pipeline processing to process any number of unique visits in the same directory without problems anymore.  

These default rules files have also been renamed to 'hap.rules' files instead of the older 'hla.rules' files.


drizzlepac/adrizzle.py
```diff
@@ -371,6 +372,7 @@ def drizFinal(imageObjectList, output_wcs, configObj,build=None,wcsmap=None,proc
         paramDict['crbit'] = configObj['crbit']
         paramDict['proc_unit'] = configObj['proc_unit']
         paramDict['wht_type'] = configObj[final_step]['final_wht_type']
+        paramDict['rules_file'] = configObj['rules_file'] if configObj['rules_file'] != "" else None
```
mdlpstsci, line 376: I seem to recall the default value for a string should be an empty string and not the value None.  Same comment to line 345 above.  Making these equal to "" would also match the comment at line 171 in the astrodrizzle.help.

drizzlepac/hlautils/processing_utils.py
```diff
@@ -24,16 +24,22 @@
...
-        shutil.copy(rules_filename, os.getcwd())
+    def_rules_name = "{}_header_hap.rules".format(instrument.lower())
+    new_rules_name = "{}_header_hap.rules".format(rootname)
+    rules_filename = os.path.join(base_dir, 'pars', def_rules_name)
+    new_rules_filename = os.path.join(os.getcwd(), new_rules_name)
+    
+    if new_rules_name not in os.listdir('.'):
+        shutil.copy(rules_filename, new_rules_filename)
+        #shutil.copy(rules_filename, os.getcwd())
+    return new_rules_name
```
mdlpstsci, line 42: Just making sure here you want to return `new_rules_name` and not `new_rules_filename`.

drizzlepac/hlautils/product.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, prop_id, obset_id, instrument, detector, filename, filetype,
         self.instrument = instrument
         self.detector = detector
         self.filetype = filetype
+        self.rules_file = None
```
mdlpstsci, line 52: Should self.rules_file default initialization be an empty string?

drizzlepac/hlautils/processing_utils.py
stsci-hack: Good question!  It turns out that the 'fitsblender' package wants to look for a file with 'new_rules_name' in the current directory for use, so we do not need the full filename with path as stored in 'new_rules_filename'.

drizzlepac/outputimage.py
```diff
@@ -172,7 +172,8 @@ def set_units(self,units):
         self.units = units
 
     def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
-                versions=None, overwrite=yes, blend=True, virtual=False):
+                versions=None, overwrite=yes, blend=True, virtual=False,
+                rules_file=None):
```
mdlpstsci, line 176: I am just commenting everywhere that I see if a string should be set to None or an empty string.  None means the variable has not been set to a default value and "" means the default is an empty string.  There are just some places in code where None will not work for a string.  Pardon my repetition on this.

drizzlepac/outputimage.py
```diff
@@ -680,7 +683,7 @@ def cleanTemplates(scihdr,errhdr,dqhdr):
                 if keyword not in dqhdr:
                     dqhdr[keyword]= scihdr[keyword]
 
-def getTemplates(fnames, blend=True):
+def getTemplates(fnames, blend=True, rules_file=None):
```
mdlpstsci, line 686: Same None vs "".

drizzlepac/outputimage.py
```diff
@@ -743,7 +749,8 @@ def deleteDistortionKeywords(hdr):
             pass
 
 
-def writeSingleFITS(data,wcs,output,template,clobber=True,verbose=True):
+def writeSingleFITS(data,wcs,output,template,clobber=True,verbose=True, 
+                    rules_file=None):
```
mdlpstsci, line 753: Same None vs "".

drizzlepac/adrizzle.py
stsci-hack: You point out something which would normally not be necessary, since it is not current good practice.  However, the 'fitsblender' package was written (many years ago using pre-PEP 8 Python) with an explicit dependence on 'None' as the default value for the 'rules_file' parameter.  Therefore, in order to get the correct behavior, we need to pass in None instead of an empty string.  Could we change `fitsblender`?  Yes.  Would that open up many other new issues as that package was updated?  Almost assuredly.  Is it worth it at this time?  I would argue that at this time, it is not worth the effort.

mdlpstsci: Agreed

drizzlepac/hlautils/processing_utils.py
mdlpstsci: OK.

PR #676, opened 2020-06-04
Update poller_utils to interpret MVM poller files

stsci-hack:
Support for multi-visit mosaic (MVM) processing starts with being able to correctly interpret the list of input files specified to be processed.  For MVM inputs, that list has several differences from the single-visit input files; namely, the addition of skycell ID(s) for each input and the addition of a flag to specify whether it has already been added to a skycell or not (is new). 

This code updates the poller_utils code to add support for interpreting these MVM input (poller) files and to return a modified tree which only contains 'filter' products with each one corresponding to a separate layer in a separate skycell.  The output tree will contain one entry for every skycell that the input files overlap, with the same file potentially contributing to multiple products.  Each layer corresponds to a unique detector/filter/calendar year/exposure time combination.   This code also generates an additional layer for each skycell/filter/year which includes ALL the exposures for the layer regardless of exposure time.  

An initial version of the sequencer for multi-visit processing is also included as a means for testing the processing of the MVM input poller files.   It is a copy of the SVM sequencer which has had all single exposure and total exposure processing and catalog generation removed, since the MVM processing only results in skycell layers.


stsci-hack:
Initial testing has only been done using a list of filenames, not a real MVM input file.  An MVM file still needs to be mocked up for use in testing.

stsci-hack:
Testing was completed with a hand-crafted, simulated MVM poller file for 'ic0s17'.  The output from 'interpret_mvm_input()' is now identical whether the simulated MVM poller file or a simple list of filenames were provided as input as a result of the changes in 7651a5e.  At this point, the poller file interpreter code could be considered ready for use in MVM processing.

stsci-hack:
This is only a first draft of this functionality.  Additional work is expected to get this code to successfully interpret actual MVM poller files once they are provided by DSB.  The processing of the data will also require many updates to the code to get that right, but this version at least serves as a proof of concept that allows for continued development and testing.

PR #680, opened 2020-06-08
Update relative astrometric resids to use HapDiagnostics

stsci-hack:
Updating the QA test for pipeline relative astrometry to use the new HapDiagnostic class for use in the new QA framework.


stsci-hack:
This initial version still requires interactive testing and debugging before it can be merged...

stsci-hack:
This PR is ready for review.  Interactive testing has been performed by running 'svm_quality_analysis.compare_photometry()' on the SVM products generated for 'iboa31', as well as running the 'quality_analysis.run_all()' on the pipeline products for 'ib3m53020'.  Reviews of the quality of the pipeline json file data itself will be addressed separately.

drizzlepac/haputils/poller_utils.py
```diff
@@ -868,8 +868,9 @@ def build_poller_table(input, log_level, poller_type='svm'):
                             poller_row['skycell_obj'] = scell_obj
                             # append new row to table
                             new_poller_table.add_row(poller_row[0])
-
-    return new_poller_table
+        poller_table = new_poller_table
```
mdlpstsci, line 871: Getting confused between all these PRs.  This line needs to be indented.  I also think Mike addressed this in his PR.

mdlpstsci: Never mind.  This is OK.

stsci-hack:
Testing with the JSON harvester from #682 uncovered problems with how the data in the output JSON files were being packaged.  The array data needed to be passed to the HapDiagnostic object in one call to 'add_data_item' while the scalar values needed to be added in a separate call.  These changes will need to be made and tested before this can be merged.

PR #685, opened 2020-06-14
Generate a Bokeh graphic which can be used as an example for other plotting routines

mdlpstsci:
This PR is actually an update of an earlier Bokeh plotting routine.  This version has various enhancements:

- Cleaned photometry_graphics.py and added a main.
- Modified svm_quality_analysis.py to add checking and handling in two tests,find_gaia_sources() and characterize_gaia_distribution(), when no GAIA sourcesare found - make the tests exit nicely.
- Modified svm_quality_analysis.py to add checking and handine for compare_photometry() when no cross-matches between catalogs are found so the test exists nicely.
- Modified svm_quality_analysis.py to add log text which indicates a when a data quality analysis test starts and to report the name of the test.
- Cleaned up output JSON text so variable names are more succinct.
- Updated pandas_utils.py to accommodate reading and use of HDF5 which is storing a Pandas dataframe of svm data quality output.
- Modified diagnostic_json_harvester.pu to remove extra period from output default filename.


drizzlepac/haputils/photometry_graphics.py
```diff
@@ -63,21 +75,20 @@ def compute_global_stats(value_array):
...
-def load_data_to_arrays(csv_filename):
-    """Load the harvested data, stored in a CSV file, into local arrays.
+def get_data(storage_filename):
+    """Load the harvested data, stored in a storage file, into local arrays.
 
     Parameters
     ==========
-    csv_filename: str
-    Name of the CSV file created by generate_photometry_df.
+    storage_filename: str
```
stsci-hack, line 83:
```suggestion
    storage_filename : str
```
Just a small adjustment to allow Sphinx to properly format the docstring.

drizzlepac/haputils/photometry_graphics.py
```diff
@@ -63,21 +75,20 @@ def compute_global_stats(value_array):
...
+    Name of the storage file for the Pandas dataframe created by the harvester.
 
     Returns
     =======
-    (d_MagAp1_mean, d_MagAp1_std, d_MagAp1_median): tuple
-    Aperture 1 statistics
-
-    (d_MagAp2_mean, d_MagAp2_std, d_MagAp2_median): tuple
-    Aperture 2 statistics
+    phot_data: Pandas dataframe
```
stsci-hack, line 88:
```suggestion
    phot_data : Pandas dataframe
```
Another minor docstring cleanup...

drizzlepac/haputils/pandas_utils.py
```diff
@@ -0,0 +1,108 @@
...
+
+        return column_data
+
+    def get_columns_HDF5(self, column_names):
+        """ Method to do the actual reading of dataframe and get the data in the
+            specified columns.
+
+            Parameters
+            ----------
+            column_names: list
```
stsci-hack, line 84:
```suggestion
            column_names : list
```

drizzlepac/haputils/pandas_utils.py
```diff
@@ -0,0 +1,108 @@
...
+            A list of the column names which specify the desired data.
+
+            Returns
+            -------
+            column_data: Pandas dataframe
```
stsci-hack, line 89:
```suggestion
            column_data : Pandas dataframe
```

drizzlepac/haputils/pandas_utils.py
```diff
@@ -0,0 +1,108 @@
...
+        # Lazy attribute creation
+        self._dataframe = None
+
+    def get_columns_CSV(self, column_names):
+        """ Method to do the actual reading of dataframe and get the data in the
+            specified columns.
+
+            Parameters
+            ----------
+            column_names: list
```
stsci-hack, line 59:
```suggestion
            column_names : list
```

drizzlepac/haputils/pandas_utils.py
```diff
@@ -0,0 +1,108 @@
...
+            A list of the column names which specify the desired data.
+
+            Returns
+            -------
+            column_data: Pandas dataframe
```
stsci-hack, line 64:
```suggestion
            column_data : Pandas dataframe
```

drizzlepac/haputils/photometry_graphics.py
```diff
@@ -86,178 +97,181 @@ def load_data_to_arrays(csv_filename):
...
 # Generate the actual plot for the "svm_graphic_type" data.
-def generate_graphic(ap1, ap2, means_ap1, means_ap2):
+def generate_graphic(phot_data, stat_Ap1, stat_Ap2, output_base_filename):
     """Generate the graphics associated with this particular type of data.
 
     Parameters
     ==========
-    ap1: tuple
-    Tuple containing the mean, std, and median statistics for Aperture 1
+    phot_data: Pandas dataframe
```
stsci-hack, line 133:
```suggestion
    phot_data : Pandas dataframe
```

drizzlepac/haputils/photometry_graphics.py
```diff
@@ -86,178 +97,181 @@ def load_data_to_arrays(csv_filename):
...
+    Dataframe consisting of the Magnitude statistics of mean, std, and median for Aperture 1
+    and Aperture 2
 
-    ap2: tuple
-    Tuple containing the mean, std, and median statistics for Aperture 2
-
-    means_ap1: tuple
+    stat_Ap1: tuple
```
stsci-hack, line 137:
```suggestion
    stat_Ap1 : tuple
```

drizzlepac/haputils/photometry_graphics.py
```diff
@@ -86,178 +97,181 @@ def load_data_to_arrays(csv_filename):
...
     Tuple containing the average of the means and the average of the medians for Aperture 1
 
-    means_ap2: tuple
+    stat_Ap2: tuple
```
stsci-hack, line 140:
```suggestion
    stat_Ap2 : tuple
```

drizzlepac/haputils/photometry_graphics.py
```diff
@@ -86,178 +97,181 @@ def load_data_to_arrays(csv_filename):
...
+    hover_p2.tooltips = [("Mean", "@{Ap2 Mean Differences}"), ("StdDev", "@{Ap2 Standard Deviation}"),
+                         ("Median", "@{Ap2 Median Differences}")]
+    p2.add_tools(hover_p2)
+
+    stat_text = ('<DeltaMagAp2_Mean>: {:6.2f}     <DeltaMagAp2_Median>: {:6.2f}'.format(stat_Ap2[0], stat_Ap2[1]))
     stat_label = Label(x=20, y=20, x_units='screen', y_units='screen', text=stat_text)
     p2.add_layout(stat_label)
 
     # Display!
     show(row(p1, p2))
```
stsci-hack, line 205: This will automatically bring up the browser to display the generate plot.  Is there a way to simply write out the HTML file without displaying it immediately?  This will be crucial in an automated environment like the pipeline testing or even when we run this code on our own server with 1000 datasets.

drizzlepac/haputils/photometry_graphics.py
```diff
@@ -86,178 +97,181 @@ def load_data_to_arrays(csv_filename):
...
-    """Driver to load the data from the CSV file and generate the graphics.
+def photometry_graphics_driver(storage_filename, output_base_filename='photometry_graphics', log_level=logutil.logging.INFO):
+    """Driver to load the data from the storage file and generate the graphics.
 
     Parameters
     ==========
-    csv_filename: str
-    Name of the CSV file created by generate_photometry_df.
-    """
+    storage_filename: str
```
stsci-hack, line 213:
```suggestion
    storage_filename : str
```

drizzlepac/haputils/photometry_graphics.py
```diff
@@ -86,178 +97,181 @@ def load_data_to_arrays(csv_filename):
...
+    sourceDF = ColumnDataSource(phot_data)
+    num_of_datasets = len(phot_data.index)
     print('Number of datasets: {}'.format(num_of_datasets))
 
     # Define a figure object
     p1 = figure()
 
     # Add the glyphs
-    p1.circle(x_index, ap1[0], fill_alpha=0.5, line_alpha=0.5, size=10, color='green',
+    p1.circle(x='x_index', y=NEW_PHOT_COLUMNS[0], source=sourceDF, fill_alpha=0.5, line_alpha=0.5, size=10, color='green',
```
stsci-hack, line 166: I found that the Pandas DataFrame automatically provides a column named 'index' which appears to be identical to what you create as 'x_index'.  If so, would it be simpler for your code to just rely on 'index' instead?

mdlpstsci: If I understand your comment, dataframes always have an index, but the index is not actually a column in the data.  I see the "index" on the rows when the file is written/opened as a CSV, but I do not see this when the file is written/opened with HDFStore.  I can follow up on this more, but in the interests of moving on, I will keep "x_index" for now (unless there is further objection).

drizzlepac/haputils/photometry_graphics.py
mdlpstsci: Fixed in next commit.

drizzlepac/haputils/photometry_graphics.py
stsci-hack: The `ColumnDataSource` object contains a full `index` column as seen here (using `pdb.set_trace()`):

```
(Pdb) fitDF.column_names
['index', 'gen_info.instrument', 'gen_info.detector', 'gen_info.filter', 'gen_info.imgname', 'header.DATE-OBS', 'header.RA_TARG', 'header.DEC_TARG', 'header.GYROMODE', 'header.EXPTIME', 'fit_results.aligned_to', 'fit_results.rms_x', 'fit_results.rms_y', 'fit_results.xsh', 'fit_results.ysh', 'fit_results.rot', 'fit_results.scale', 'fit_results.rot_fit', 'fit_results.scale_fit', 'fit_results.nmatches', 'fit_results.skew', 'x_index']
(Pdb) fitDF.data['index']
array([0, 1, 2])
```
This was what made me think that the `x_index` column was not needed.  If this will not be sufficient, then please keep the `x_index` column as-is.

mdlpstsci: I think the issue here is the difference between a dataframe which has been stored as CSV versus HDF5.  For the CSV file, the dataframe read from the file lists a running index column when just printing the dataframe. However, there is no index column when printing the column names.  Once the a ColumnDataSource object is created from the "CSV dataframe", there is an index column containing what you would expect (e.g, 1, 2, 3, etc.).  This is **not** true when reading the data from an HDF5 file.

When just printing the "HDF dataframe", there is no running index column printed, and there is no index column when printing the column names. There is a column which is printed that contains source file names (e.g., hst_11665_06_wfc3_ir_f110w_ib4606), and this column has no name.  Once the a ColumnDataSource object is created from the "HDF dataframe", there is an index column called 'index', but it contains the source file names and not a running numerical index.  

I checked the HDFStore  to see if there was a parameter on the read of the file to create a numeric index, but this is not the case.

PR #688, opened 2020-06-16
Add HAPLEVEL to ExposureProduct file headers

stsci-hack:
Pipeline processing of HAP products requires the use of the HAPLEVEL keyword in order for the system to categorize all the products correctly and successfully archive the results.  This small fix adds this keyword to the HAP input files.


stsci-hack:
This PR should not really be merged until there is agreement on what value should be set for this keyword for the ExposureProduct (hst*_flt.fits/hst*_flc.fits) files.

PR #689, opened 2020-06-17
Updated report_wcs() in svm_quality_analysis.py

mdlpstsci:
- Fixed report_wcs() to generate its JSON output in a more efficient manner in terms of space saving.  This comes at the cost of more JSON files being produced - one per exposure in a visit. However, this makes the WCS data in the final Pandas dataframe more accessible.
- Made the output JSON file name conform to the naming convention
adopted for the SVM products.

### DIFF
--- a/drizzlepac/haputils/svm_quality_analysis.py
+++ b/drizzlepac/haputils/svm_quality_analysis.py
@@ -830,28 +830,25 @@ def report_wcs(total_product_list, json_timestamp=None, json_time_since_epoch=No
     """
     log.setLevel(log_level)
     log.info('\n\n*****     Begin Quality Analysis Test: report_wcs.     *****\n')
+    apriori_suffix_list = ['HSC30', 'GSC240']
+    aposteriori_list = ['FIT', 'SVM', 'MVM']
 
-    # Generate a separate JSON file for each total product (a total product
-    # consists of a single detector.  Each total product consists of one or
-    # more ExposureProduct objects.
+    # Generate a separate JSON file for each ExposureProduct object
     for total_product in total_product_list:
         detector = total_product.detector
-        ipppss = total_product.edp_list[0].exposure_name[0:6]
-
-        # Set up the diagnostic object
-        diagnostic_obj = du.HapDiagnostic()
-        diagnostic_obj.instantiate_from_hap_obj(total_product,
-                                                data_source="{}.report_wcs".format(__taskname__),
-                                                description="WCS information",
-                                                timestamp=json_timestamp,
-                                                time_since_epoch=json_time_since_epoch)
-
-        # Construct the output JSON filename
-        json_filename = '_'.join([ipppss, detector, 'svm_wcs.json'])
 
         # Loop over all the individual exposures in the list which comprise the total product
         for edp_object in total_product.edp_list:
-            edp_filter = edp_object.filters
+            # Set up the diagnostic object
+            diagnostic_obj = du.HapDiagnostic()
+            diagnostic_obj.instantiate_from_hap_obj(edp_object,
+                                                    data_source="{}.report_wcs".format(__taskname__),
+                                                    description="WCS information",
+                                                    timestamp=json_timestamp,
+                                                    time_since_epoch=json_time_since_epoch)
+
+            # Construct the output JSON filename
+            json_filename = '_'.join([edp_object.product_basename, 'svm_wcs.json'])
 
             # For exposures with multiple science extensions (multiple chips),
             # generate a combined WCS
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -830,28 +830,25 @@ def report_wcs(total_product_list, json_timestamp=None, json_time_since_epoch=No
     """
     log.setLevel(log_level)
     log.info('\n\n*****     Begin Quality Analysis Test: report_wcs.     *****\n')
+    apriori_suffix_list = ['HSC30', 'GSC240']
```
stsci-hack, line 833: This will work for now.  Unfortunately, if additional apriori solutions are defined, this code would need to be updated to correctly recognize them, which would probably not happen until problems start showing up (in the archive again?).  The absence of any of the aposteriori terms in the WCS and presence of the '-' in the WCSNAME would be a sufficient universal check of whether a WCSNAME is apriori.  This would work due to the fact that the aposteriori terms are fixed and will not change, ever, as far as I can imagine.

drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -940,61 +939,81 @@ def report_wcs(total_product_list, json_timestamp=None, json_time_since_epoch=No
                 # Restore an alternate to be the primary WCS
                 for key, value in dict_of_wcskeys_names.items():
                     if key != ' ':
-                        wcsutil.altwcs.restoreWCS(edp_object.full_filename, ext=extname_list, wcskey=key)
                         alt_key = key
-                        alt_wcs_name = value
+                        alt_wcs_name = value.strip()
+
+                        if len(alt_wcs_name) == 13 and alt_wcs_name.startswith('IDC_'):
```
stsci-hack, line 945: We currently have 9 character rootnames for the IDCTAB's leading to 13 character long IDC_* WCSNAME.  However, that relies on an assumption which could be broken even if only by the end user.  I believe a safer way to detect the default WCS would be to look for:

    '-' not in alt_wcs_name and alt_wcs_name.startswith('IDC_')

That makes no assumptions about filename length while eliminating OPUS and apriori/aposteriori solutions which add the '-*' terms at the end.

mdlpstsci: Much better solution.  Fix seen on next commit.

drizzlepac/haputils/svm_quality_analysis.py
mdlpstsci: Fix found in next commit.  Added comment for examples of WCS naming convention for default, a priori, and a posteriori solutions.

drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -937,15 +936,19 @@ def report_wcs(total_product_list, json_timestamp=None, json_time_since_epoch=No
...
                     if key != ' ':
                         alt_key = key
                         alt_wcs_name = value.strip()
 
-                        if len(alt_wcs_name) == 13 and alt_wcs_name.startswith('IDC_'):
+                        if '-' not in alt_wcs_name and alt_wcs_name.startswith('IDC_'):
                             alt_name = 'AlternateWCS_default'
                             delta_name = 'DeltaWCS_default'
-                        elif list(filter(alt_wcs_name.endswith, apriori_suffix_list)) != []:
+                        elif '-' in alt_wcs_name and 'FIT' not in alt_wcs_name:
```
stsci-hack, line 951: This logic would appear, to me at least, to allow 'IDC_12345678q-SVM_REL_GAIADR2' to be identified as an apriori solution.  Unfortunately, if there is an '-' in the WCSNAME, any of the aposteriori_list values could be present, not just 'FIT'.

mdlpstsci: I tried to look for a listing of all the possible "a_posteriori" names in order to setup the logic properly for the "a priori" discriminant, and I found this

<Starting WCS>-FIT_<REL|IMG>_<Astrometric Catalog>

which is why I used the logic you see.  Looking at some of the SVM WCSNAMES and poking in the source code, it seems the "a posteriori" WCS is

<Starting WCS>-FIT_<REL|IMG|SVM>_<Astrometric Catalog>

I guess I did not piece together the definitive list.  Obi-Wan, where is a definitive list?

stsci-hack: The definitive list should be in the Drizzlepac readthedocs, something which has not been updated with all the latest options.  This part of the WCSNAME gets defined during SVM/MVM processing using the `align_to_gaia()` method of the `Product` class through the use of the fit_label parameter which gets passed to the `AlignmentTable.apply_fit()` method.   Now that we have this implemented, your experience confirms that we really need to update our documentation, including the docstrings for these methods.

mdlpstsci: No worries.

PR #691, opened 2020-06-18
Expand search for JSON files in harvester

stsci-hack:
The current version of the JSON harvester will only search for JSON files in the currently specified directory.  This works perfectly well when all the data is in a single directory, perhaps as will be the case for pipeline processing.  Unfortunately, testing on large numbers of datasets, each in their own directory, makes this code impossible to collate all the JSON files since it would need to be called separately for each dataset/directory.  

This small revision simply adds logic to look in the underlying directories for the desired JSON files.  The result is that a single call pointing to the parent directory for all the datasets directories can now find all the desired JSON files in one step resulting in a collated Pandas DataFrame file with all the results.

### DIFF
--- a/drizzlepac/haputils/diagnostic_json_harvester.py
+++ b/drizzlepac/haputils/diagnostic_json_harvester.py
@@ -82,6 +82,11 @@ def get_json_files(search_path=os.getcwd(), log_level=logutil.logging.INFO):
     for search_pattern in search_patterns:
         search_string = os.path.join(search_path, search_pattern)
         search_results = glob.glob(search_string)
+        if len(search_results) == 0:
+            # Try another directory lower
+            search_string = os.path.join(search_path, '*', search_pattern)
+            search_results = glob.glob(search_string)
+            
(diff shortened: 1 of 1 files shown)


stsci-hack:
The revision with the first commit was successfully used to generate the combined Pandas DataFrame for all the pipeline astrometry JSON files from the results_2020-06-17 test run.

drizzlepac/haputils/diagnostic_json_harvester.py
```diff
@@ -82,9 +82,14 @@ def get_json_files(search_path=os.getcwd(), log_level=logutil.logging.INFO):
     for search_pattern in search_patterns:
```
mdlpstsci, line 82: Please update docstring for search path as now the search will be performed one level deeper if no JSON files are found.

I also think the docstring for the entire routine should be updated to indicate this routine (unless it is changed) does not do a recursive search of the specified search path, but rather only searches the specified path and one level deeper, as deemed necessary.

PR #692, opened 2020-06-18
Update QA plots with useful Bokeh tools

stsci-hack:
The initial version of many of the QA plots did not include any Bokeh tools to help the user navigate around the sometimes very crowded plots.  These updates add those tools which are most useful to the plots.


stsci-hack:
No need to merge this in immediately, since other changes to make the plots more useful for end-users may still need to be made after INS starts looking at them.

PR #693, opened 2020-06-18
Fix cal QA errors found during testing

stsci-hack:
A series of errors were uncovered during the 06-17-2020 tests of the astrometry QA code.  These changes were based on the errors identified using 'jb7h07010' that reproduced the problems uncovered in the 06-17 tests.


stsci-hack:
This is affecting SVM testing as well, so getting this merged quickly...

PR #699, opened 2020-07-01
HAPFigure class in support of HAP graphics

mdlpstsci:
This is a work in progress, but I want other members of Burgundy team to see the code easily.  Intend to bring the graph_utils module to a stable state, and then I will begin to migrate the graphics in svm_quality_analysis to using the new class.  This will allow fleshing out and debugging of HAPFigure.


drizzlepac/haputils/graph_utils.py
```diff
@@ -0,0 +1,258 @@
...
+log = logutil.create_logger(__name__, level=logutil.logging.NOTSET, stream=sys.stdout,
+                            format=SPLUNK_MSG_FORMAT, datefmt=MSG_DATEFMT)
+
+# Data columns from the Pandas dataframe which are to be used as common display value
+# for all hover tooltips.
+HOVER_BASIC_TIPS = [('Inst/Det', '@{gen_info.inst_det}'),
+                    ('Proposal ID', '@{gen_info.proposal_id}'),
+                    ('ASN ID', '@{header.ASN_ID}'),
+                    ('Filter', '@{gen_info.filter}'),
+                    ('ImageName', '@{gen_info.imgname}')]
```
stsci-hack, line 29: I would actually swap 'ImageName' and 'Inst/Det' in this list.  When reviewing the plot, the 'Inst/Det' will already be understood through the definition of the glyph color, while the user will want to know what image it corresponds with to look for any supplemental plots that may be available for that image.

drizzlepac/haputils/graph_utils.py
```diff
@@ -0,0 +1,258 @@
...
+                       'triangle': HAPFigure.build_triangle_glyph}
+
+        glyph_types[glyph_name](self)
+
+
+    # "Shape" glyphs
+    def build_circle_glyph(self):
+
+        # hacky
+        if self.glyph_color is 'colormap':
```
stsci-hack, line 192: I am wondering whether this 'if...else' statement could simply define the values of 'legend_group' and 'legend_label' that would then get passed to a single call to 'self.fig.circle'.  Just a thought...

drizzlepac/haputils/wcs_graphics.py
```diff
@@ -216,91 +218,87 @@ def generate_graphic(wcs_dataDF, output_base_filename, log_level):
...
+                        y_label = 'Delta Orientation (degrees)',
+                        grid_line_color = 'white',
+                        hover_tips = wcs_tips)
 
     # Figure 1 delta_crpix1 vs delta_crpix2
     # Figure 2 delta_crval1 vs delta_crval2
-    wcs_components = ['del_def_wcsname', 'del_apriori_wcsname', 'del_apost_wcsname']
-    wcs_type_colors = ['blue', 'green', 'purple']
+    # Figure 3 delta_scale vs delta_orientation
     wcs_type_names = ['default', 'apriori', 'aposteriori']
```
stsci-hack, line 246: I have to wonder whether these should be module level variables (WCS_TYPE_NAMES?) instead of hidden, hard-coded options.

mdlpstsci:
The routines of interest: pandas_utils.py, graphic_utils.py, wcs_graphics.py and svm_quality_graphics.py.

drizzlepac/haputils/graph_utils.py
stsci-hack: I see this has been updated in the code.

drizzlepac/haputils/graph_utils.py
mdlpstsci: You can only pass one of the possible "legend" parameters (of which there are three that exist) to the glyph creation functionality.  I am looking for alternatives to get rid of the hacky code.

stsci-hack: Not a blocker, merely a potential feature to be implemented or used.

PR #707, opened 2020-07-06
Add vector plots to pipeline astrometric Q&A results

### LINKED ISSUE #700: PIPELINE Q&A: Only generate plots for a posteriori solutions
_Issue [HLA-379](https://jira.stsci.edu/browse/HLA-379) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

INS noticed that observations with only a priori solutions were also being summarized in the Q&A plots being generated.  They have requested that only observations with a posteriori solutions be included.

stsci-hack:
Vector plots of the relative alignment of observations within an association can provide useful information about the quality of the alignment and any potential problems with the processing.  

These changes:

- implement vector plots for each exposure as a separate Bokeh plot (HTML page)
- only generates astrometric JSON results files for data with a posteriori solutions

This code was used to generate the initial set of residuals plots for the 2020-06-24 test.  

When merged, this PR will close #700 along with Jira Ticket HLA-379.

### DIFF
--- a/drizzlepac/haputils/align_utils.py
+++ b/drizzlepac/haputils/align_utils.py
@@ -601,20 +601,25 @@ def match_relative_fit(imglist, reference_catalog, **fit_pars):
     """
     log.info("{} (match_relative_fit) Cross matching and fitting {}".format("-" * 20, "-" * 27))
     if 'fitgeom' in fit_pars:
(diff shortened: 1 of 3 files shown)


stsci-hack:
This work should really be implemented as part of #699 instead.

PR #711, opened 2020-07-10
Add pipeline Q&A vector plots

stsci-hack:
Pipeline Q&A vector plots were added to the set of plots being generated for the standard pipeline astrometry testing.  The code for creating vector plots in Bokeh comprises the bulk of the changes. 

In addition, though, the fitting was also updated to enforce a limit on the minimum number of objects to be included in the fit at 6, instead of the default value of 3 that comes with tweakwcs.


stsci-hack:
This code has already been used to generate the set of astrometry residual plots for the 06-24 and 07-07 test results.

PR #671, opened 2020-05-28
Update docs to describe astrometry processing more fully

stsci-hack:
The documentation in this package needs to be updated to more fully explain the expanded astrometry processing now implemented for pipeline calibration of HST imaging data.  Updates to the documentation are intended to provide end-users with a better understanding of the logic and algorithms used to do what is possible to get their HST data aligned to GAIA as well as possible while retaining the relative alignment between all their exposures.

PR #713, opened 2020-07-10
Fix clumsy text in runastrodriz docs

stsci-hack:
Simple update to reword some very clumsy descriptions in the runastrodriz readthedocs documentation.


stsci-hack:
This addressed the comment made in the review of #671 that was originally intended to be included in that PR.

PR #715, opened 2020-07-13
Simplify logic for archiving WCS in tweakreg

### LINKED ISSUE #697: TweakReg may not backup primary WCS
Under certain circumstances, `TweakReg` may not back up the primary WCS. Specifically,
```python
TweakReg(..., updatehdr=True, wcsname='TWEAK', reusename=True)
```
will not back up the primary WCS but it will archive a copy of the _updated (tweaked) primary WCS_. This second part is the result of the behavior described in #690

mcara:
This PR greatly simplifies the logic for archiving primary WCS. Now, if `'reusename'=True` then primary WCS is archived if provided `'wcsname'`is different from the `'WCSNAME'` of the primary WCS in image headers.

If primary WCS' name is the same (case-insensitive) as the name provided by `'wcsname'` parameter and `'reusename'=True`, then primary WCS will be updated by `tweakreg` in-place without archiving primary WCS. If `'reusename'=False` then an exception will be raised.

This PR reduces the number of duplicate alternate WCS created by `tweakreg` and it also solves the issue of non-unique WCS names in image headers: the algorithm is designed not to create WCSes with the same name.

Fixes #697 

This PR **requires functionality from** https://github.com/spacetelescope/stwcs/pull/149 and https://github.com/spacetelescope/stwcs/pull/153

CC: @nden

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,9 @@ of the list).
 In addition to a few dozen bug fixes, the following updates to the algorithms
 were also implemented.
 
+- Simplified the logic in ``tweakreg`` for deciding how to archive primary WCS
+  resulting in a reduction of duplicate WCSes in image headers. [#715]
+
 - Added polynomial look-up table distortion keywords to the list of distortion
   keywords used by ``outputimage.deleteDistortionKeywords`` so that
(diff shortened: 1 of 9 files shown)


mcara:
Both required PRs in `stwcs` have been merged so this PR can be merged as well.

PR #718, opened 2020-07-14
Remove stsci.disttools and d2to1 from azure pipeline env.

rendinam:
Try removing these packages from the CI environment in an attempt to alleviate errors.


mcara:
Thank you @rendinam

Issue #702, opened 2020-07-03
SVM Q&A: Generate cross-filter point-source comparisons


stscijgbot-hstdp:
_Issue [HLA-381](https://jira.stsci.edu/browse/HLA-381) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The primary benefit of performing single-visit processing would be the alignment of the products for all filters to the same coordinate system (pixels).  Verification of this can be performed by cross-matching sources from each filter's point-source catalog, then computing the differences in positions.  The differences should all be computed relative to the filter which has the largest number of sources in the catalog. 

These results should be written out as JSON files using the standard interface for SVM QA results.

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-381?focusedCommentId=472903#comment-472903):_

Development branch URL: <https://github.com/dulude/drizzlepac/tree/HLA-381_crossmatch_filters](https://github.com/dulude/drizzlepac/tree/HLA-381_crossmatch_filters)>

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-381?focusedCommentId=474164#comment-474164):_

See [PR #&#x2060;719](https://github.com/spacetelescope/drizzlepac/pull/719)

stsci-hack:
Resolved with #719 on 15-Jul-2020.

PR #719, opened 2020-07-15
Generate cross-filter point-source comparisons

dulude:
**Linked issues/tickets**
- Git issue #702 
- JIRA ticket [HLA-381](https://jira.stsci.edu/browse/HLA-381)

**High-level summary**
-Added new test to svm_quality_analysis.py that quantifies the quality of alignment of filter products in multi-filter datasets. 
- This is accomplished for a given multi-filter dataset by:

1. independently identifying point sources in each filter product
2. convert X and Y coords of all identified sources to the common reference frame of the image that yielded the most sources.
3. Cross-match using the catalog with the most sources as the reference for all crossmatches
4. Compare the X, Y positions of the cross-matched sources in the various catalogs vs. the reference catalog and compute statistics on the differences.
5. For each interfilter comparison, write details of the cross-match, computed stats and the cross-matched catalogs to a .json file.
 
**Summary of changes to code**
- starmatch_hist.py:
     - added support for temporary *fxm.ecsv catalog files generated for the crossmatch by svm_quality_analysis.compare_interfilter_crossmatches().
-  hla_flag_filter.py:
     - added optional input parameter "origin" to subroutines xytord() and rdtoxy() so that the x, y coordinate origin can be specified.
- svm_quality_analysis.py:
     -  Assorted pep8 fixes throughout code
     - Added subroutines compare_interfilter_crossmatches(), find_hap_point_sources() and transform_coords() for the new test
     - added new optional command line input switch '-fxm' to run new interfilter crossmatch test

### DIFF
--- a/drizzlepac/haputils/svm_quality_analysis.py
+++ b/drizzlepac/haputils/svm_quality_analysis.py
@@ -31,6 +31,7 @@ import collections
 from datetime import datetime
 import glob
 import json
+import math
 import os
 import pdb
 import pickle
@@ -546,6 +547,7 @@ def compare_interfilter_crossmatches(total_obj_list, json_timestamp=None, json_t
     # Initiate logging!
     log.setLevel(log_level)
 
+    log.info('\n\n*****     Begin Quality Analysis Test: compare_interfilter_crossmatches.     *****\n')
     # Check to make sure there's at last 2 filter-level products. If not, return.
     num_filter_prods = 0
     for total_obj in total_obj_list:
@@ -560,31 +562,29 @@ def compare_interfilter_crossmatches(total_obj_list, json_timestamp=None, json_t
         return
     else:
         log.info("Found {} filter-level products for use in analysis of inter-filter cross matched source"
-                 " positions:".format(num_filter_prods))
-        ctr = 1
-        for total_obj in total_obj_list:
-            for filt_obj in total_obj.fdp_list:
-                log.info("{}: {}".format(ctr, filt_obj.drizzle_filename))
-                ctr += 1
+                 " positions".format(num_filter_prods))
 
         filtobj_dict = {}
         xmatch_ref_imgname = None
         max_sources = 0
-
+        ctr = 1
         for total_obj in total_obj_list:
             for filt_obj in total_obj.fdp_list:
-                log.info("{} {} {}".format(">"*20, filt_obj.drizzle_filename, "<"*20))
+                log.info("{} {}: {} {}".format(">"*20, ctr, filt_obj.drizzle_filename, "<"*20))
                 filtobj_dict[filt_obj.drizzle_filename] = find_hap_point_sources(filt_obj, log_level=log_level)
                 n_sources = len(filtobj_dict[filt_obj.drizzle_filename]['sources'])
                 log.info("Identified {} sources in {}".format(n_sources, filt_obj.drizzle_filename))
                 if n_sources > max_sources:
                     max_sources = n_sources
                     xmatch_ref_imgname = filt_obj.drizzle_filename
+                ctr += 1
+                log.info("")
         log.info("Crossmatch reference image {} contains {} sources.".format(xmatch_ref_imgname, max_sources))
         xmatch_ref_catname = xmatch_ref_imgname[:-8] + "point-cat-fxm.ecsv"
 
         # Perform coord transform and write temp cat files for cross match
         temp_cat_file_list = []
+        log.info("")
         for imgname in filtobj_dict.keys():
             filtobj_dict[imgname] = transform_coords(filtobj_dict[imgname],
                                                      xmatch_ref_imgname,
@@ -592,7 +592,7 @@ def compare_interfilter_crossmatches(total_obj_list, json_timestamp=None, json_t
             temp_cat_name = imgname[:-8] + "point-cat-fxm.ecsv"
             temp_cat_file_list.append(temp_cat_name)
             filtobj_dict[imgname]["sources"].write(temp_cat_name, format="ascii.ecsv")
-            log.info("Wrote source catalog {}".format(temp_cat_name))
+            log.info("Wrote temporary source catalog {}".format(temp_cat_name))
 
             # write out ds9 region files if log level is 'debug'
             if log_level == logutil.logging.DEBUG:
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -516,6 +516,304 @@ def compare_ra_dec_crossmatches(hap_obj, json_timestamp=None, json_time_since_ep
...
+                log.info("{}: {}".format(ctr, filt_obj.drizzle_filename))
+                ctr += 1
+
+        filtobj_dict = {}
+        xmatch_ref_imgname = None
+        max_sources = 0
+
+        for total_obj in total_obj_list:
+            for filt_obj in total_obj.fdp_list:
+                log.info("{} {} {}".format(">"*20, filt_obj.drizzle_filename, "<"*20))
```
stsci-hack, line 576: You could include the `ctr` from the previous loop in here and add it to this log line with something like:
    "{} Filter #{}: {}  {}".format(">"*20, ctr, filt_obj.drizzle_filename, '<'*20)

Functionally, this would not change anything, but it would limit the number of lines in the log file and still provide the same information while removing an extra loop over the same information.

drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -516,6 +516,304 @@ def compare_ra_dec_crossmatches(hap_obj, json_timestamp=None, json_time_since_ep
...
+
+        # Perform coord transform and write temp cat files for cross match
+        temp_cat_file_list = []
+        for imgname in filtobj_dict.keys():
+            filtobj_dict[imgname] = transform_coords(filtobj_dict[imgname],
+                                                     xmatch_ref_imgname,
+                                                     log_level=log_level)
+            temp_cat_name = imgname[:-8] + "point-cat-fxm.ecsv"
+            temp_cat_file_list.append(temp_cat_name)
+            filtobj_dict[imgname]["sources"].write(temp_cat_name, format="ascii.ecsv")
```
stsci-hack, line 594: It would simpler (and less error-prone?) to add an entry for `cat_file` to `filtobj_dict[imgname]` to store the value of `temp_cat_name` since it will be needed later in the processing (see Line 615).  This would make sure that the name only gets constructed in one place for use everywhere else, so that the rest of the code does not need to know how or what was used to make that name, it can just simply use the stored value.

drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -516,6 +516,304 @@ def compare_ra_dec_crossmatches(hap_obj, json_timestamp=None, json_time_since_ep
...
+                    # Generate tables containing just "xcentroid_ref" and "ycentroid_ref" columns with only
+                    # the cross-matched comparision sources
+                    matched_comp_coords = filtobj_dict[imgname]["sources"].copy()
+                    matched_comp_coords.keep_columns(['xcentroid_ref', 'ycentroid_ref'])
+                    matched_comp_coords = matched_comp_coords[matching_lines_comp]
+
+                    # store comparison matched sources catalog
+                    diag_obj.add_data_item(matched_comp_coords,
+                                           "Interfilter cross-matched comparison catalog",
+                                           descriptions={"xcentroid_ref": "xcentroid_ref",
```
stsci-hack, line 694: Should these column names include '_ref' since they refer to the comparison image positions?  We really do not want these columns confused (or possibly over-written by or over-writing) the same columns from the reference data.

drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -516,6 +516,304 @@ def compare_ra_dec_crossmatches(hap_obj, json_timestamp=None, json_time_since_ep
...
+                                                             "3x3 sigma-clipped standard deviation": "3x3 sigma-clipped standard deviation of differences"},
+                                               units={"Non-clipped min": "pixels",
+                                                      "Non-clipped max": "pixels",
+                                                      "Non-clipped mean": "pixels",
+                                                      "Non-clipped median": "pixels",
+                                                      "Non-clipped standard deviation": "pixels",
+                                                      "3x3 sigma-clipped mean": "pixels",
+                                                      "3x3 sigma-clipped median": "pixels",
+                                                      "3x3 sigma-clipped standard deviation": "pixels"})
+        # TODO: should stats be in instrument-specific pixels or instrument-generic arcseconds??
```
stsci-hack, line 738: This is a good question.  Getting the images to combine correctly with 'drizzlepac', though, relies on image alignment to be good to better than 0.1 output pixels regardless of the size of the pixel on the sky.  Given that we may be comparing filters from different detectors, it may make sense to include the plate scale for the 'reference frame' as part of the summary information on the cross-matched catalogs.

drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -516,6 +516,304 @@ def compare_ra_dec_crossmatches(hap_obj, json_timestamp=None, json_time_since_ep
...
+    # 2b: perform ra, dec -> x_centroid_ref, y_centroid_ref transform
+    xy_centroid_ref_values = hla_flag_filter.rdtoxy(ra_dec_values, xmatch_ref_imgname, fits_exten,
+                                                    origin=origin)
+
+    # 3: add new columns to filtobj_subdict['sources'] table
+    title_data_dict = collections.OrderedDict()
+    title_data_dict["ra"] = ra_dec_values[:, 0]
+    title_data_dict["dec"] = ra_dec_values[:, 1]
+    title_data_dict["xcentroid_ref"] = xy_centroid_ref_values[:, 0]
+    title_data_dict["ycentroid_ref"] = xy_centroid_ref_values[:, 1]
```
stsci-hack, line 807: It may be useful to add the xmatch_ref_imgname plate scale to this dictionary so it can be passed along to the JSON file later for interpretation of the pixel differences.

drizzlepac/haputils/svm_quality_analysis.py
dulude: Issue resolved in commit [b7fb0b3](https://github.com/spacetelescope/drizzlepac/pull/719/commits/b7fb0b33a8f7ad925bcb1451d9021de938d055c5)

drizzlepac/haputils/svm_quality_analysis.py
dulude: issue resolved by commit [f0bb5ad](https://github.com/spacetelescope/drizzlepac/pull/719/commits/f0bb5ad7c371aecf7bd615393618889fe2c1aa90)

drizzlepac/haputils/svm_quality_analysis.py
dulude: issue resolved by commit [f0bb5ad](https://github.com/spacetelescope/drizzlepac/pull/719/commits/f0bb5ad7c371aecf7bd615393618889fe2c1aa90)

PR #723, opened 2020-07-16
Use new API for pipeline Q&A astrometry plots

stsci-hack:
The pipeline astrometry Q&A plots were moved out of `quality_analysis.py` into its own (new) module `pipeline_graphics.py`.  All the plots now use the new plotting API from `graph_utils` as well.

In addition, the residuals plots were reset to use the same Y range for all plots, with a minimum range of -0.5 to 0.5.  The plots were also re-ordered on the page to match the order of plots generated by `tweakreg`.  Finally, the vector plot was re-centered so that all the points are visible (from 0 - range) by default.  These modifications to the plot characteristics were all implemented on request by INS.


drizzlepac/haputils/pipeline_graphics.py
```diff
@@ -0,0 +1,411 @@
...
+                   'fit_results.skew']
+
+RESIDS_COLUMNS = ['residuals.x',
+                  'residuals.y',
+                  'residuals.ref_x',
+                  'residuals.ref_y']
+TOOLSEP_START = '{'
+TOOLSEP_END = '}'
+    
+FIGURE_TOOLS = graph_utils.FIGURE_TOOLS_BASE
```
mdlpstsci, line 119: For the new interface, this should be removed from this file.  We decided to have a consistent set of tools and not allow the user to add more tools on-the-fly.  This entry should be ignored by the constructor, and I see I added a comment.  However, I did not eliminate the code!  I will fix.  In any case, this line of code should be eliminated.

drizzlepac/haputils/pipeline_graphics.py
```diff
@@ -0,0 +1,411 @@
...
+    num_of_datasets = len(fitCDS.data['index'])
+    print('Number of datasets: {}'.format(num_of_datasets))
+        
+    plot_list = []
+
+    # Data point figures
+    p1 = HAPFigure(title='RMS Values',
+                   x_label="RMS_X (pixels)",
+                   y_label="RMS_Y (pixels)",
+                   tools=FIGURE_TOOLS)
```
mdlpstsci, line 226: the "tools=FIGURE_TOOLS" should be removed from the parameter list in all places in this code.  You will get the default set of tools defined in graph_utils.py.

drizzlepac/haputils/pipeline_graphics.py
```diff
@@ -0,0 +1,411 @@
...
+HOVER_COLUMNS = ['gen_info.imgname',
+                 'gen_info.instrument',
+                 'gen_info.detector',
+                 'gen_info.filter',                
+                 'header.DATE-OBS',
+                 'header.RA_TARG',
+                 'header.DEC_TARG',
+                 'header.GYROMODE',
+                 'header.EXPTIME',
+                 'fit_results.aligned_to']
```
mdlpstsci, line 94: Maybe I need to add more docs.  You get all the "gen_info" and "header.*" columns available in the Pandas DataFrame for free in pandas_utils, as well as the 'inst_det' and 'colormap' columns.  While this list can serve the same purpose and are closely related, this is slightly different than generating the hover tooltips.  I did not protect against dups in the hover tools tips list.

You get a few hover tooltips for free in

drizzlepac/haputils/pipeline_graphics.py
```diff
@@ -0,0 +1,411 @@
...
+                   'fit_results.nmatches',
+                   'fit_results.skew']
+
+RESIDS_COLUMNS = ['residuals.x',
+                  'residuals.y',
+                  'residuals.ref_x',
+                  'residuals.ref_y']
+TOOLSEP_START = '{'
+TOOLSEP_END = '}'
+
```
mdlpstsci, line 118: Lines 112 - 117 should be removed from this code as you should not need them.  Build_tool_tips routine lives in graph_utils.py.

drizzlepac/haputils/pipeline_graphics.py
mdlpstsci: This does not seem to hurt anything even though there will be duplicate columns in the output dataframe.  I can clean this up in the pandas_utils.py.  This is really just a matter of different style of different folks coding. This is not a blocker.

drizzlepac/haputils/pipeline_graphics.py
stsci-hack: Removed in 057ba8a.

drizzlepac/haputils/pipeline_graphics.py
stsci-hack: Removed (in all locations) in 057ba8a.

drizzlepac/haputils/pipeline_graphics.py
stsci-hack: Fixed in 057ba8a.

drizzlepac/haputils/pipeline_graphics.py
stsci-hack: Fixed in 057ba8a.

stsci-hack:
In addition to addressing the review comments, I also took this opportunity to move the `get_pandas_data()` function from `pipeline_graphics` to `pandas_utils` so that it could be used as a general function.  This new version effectively merges the version from `svm_quality_graphics` and `pipeline_graphics` (formerly, `quality_analysis`).

drizzlepac/haputils/pandas_utils.py
```diff
@@ -35,6 +35,45 @@
 DETECTOR_LEGEND = {'UVIS': 'magenta', 'IR': 'red', 'WFC': 'blue',
                    'SBC': 'yellow', 'HRC': 'black'}
 
+def get_pandas_data(pandas_filename, data_columns, log_level=logutil.logging.NOTSET):
+    """Load the harvested data, stored in a CSV file, into local arrays.
```
mdlpstsci, line 39: Change CSV -> HDF5.
Please change CSV to HDF5 all locations as appropriate.

drizzlepac/haputils/pipeline_graphics.py
```diff
@@ -0,0 +1,379 @@
...
+    # TODO: include the date from the input data as part of the html filename
+    # Set the output file immediately as advised by Bokeh.
+    output_file(output)
+
+    num_of_datasets = len(fitCDS.data['index'])
+    print('Number of datasets: {}'.format(num_of_datasets))
+        
+    pipeline_tips = build_tooltips(tooltips_list, hover_columns, list(range(0, len(hover_columns))))
+        
+    plot_list = []
```
mdlpstsci, line 173: plot_list can be deleted.

drizzlepac/haputils/pandas_utils.py
stsci-hack: Addressed in 09f65ce9

drizzlepac/haputils/pipeline_graphics.py
stsci-hack: Addressed in 09f65ce9

PR #739, opened 2020-07-21
Update SVM point-to-segment catalog xmatch comparison plots

stsci-hack:
The new API was used to generate the point-to-segment-catalog cross-match comparison histograms.  

This set of changes also includes an update to the JSON harvester code to support user-provided search patterns.  Using only the SVM search pattern allows the harvester to work on the full set of SVM JSON files to successfully create a single master file. 

Finally, the `get_pandas_data` function has been replaced by an updated function with the same name in 'pandas_utils' in an effort to minimize the number of copies of at least this one utility function. 

This supports [Jira Ticket HLA-394](https://jira.stsci.edu/browse/HLA-394).

### DIFF
--- a/drizzlepac/haputils/diagnostic_json_harvester.py
+++ b/drizzlepac/haputils/diagnostic_json_harvester.py
@@ -56,7 +56,9 @@ def flatten_dict(dd, separator='.', prefix=''):
 # ------------------------------------------------------------------------------------------------------------
 
 
-def get_json_files(search_path=os.getcwd(), log_level=logutil.logging.INFO):
+def get_json_files(search_path=os.getcwd(), 
+                   search_patterns = ["*_svm_*.json", "*_mvm_*.json", "*_cal_qa_*.json"],
+                   log_level=logutil.logging.INFO):
     """use glob to create a list of json files to harvest
     
     This function looks for all the json files containing qa test results generated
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/diagnostic_json_harvester.py
```diff
@@ -205,11 +208,17 @@ def json_harvester(json_search_path=os.getcwd(), log_level=logutil.logging.INFO,
...
+                               log_level=log_level)
     master_dataframe = None
     # extract all information from all json files related to a specific Pandas DataFrame index value into a
     # single line in the master dataframe
-    for idx in json_dict.keys():
+    num_json = len(json_dict)
+    for n,idx in enumerate(json_dict.keys()):
+        if ((n/num_json) % 0.1) == 0:
+            log.info("Harvested {}% of the JSON files".format((n/num_json)*100))
+
```
mdlpstsci, line 221: This progress information is very nice to have!

stsci-hack: This minimizes the log messages while still providing progress on the processing.  Any time stamps on the log messages themselves then act as timing for the processing.  It skips some values, since it relies on exact match to 0.1, but it helps.

PR #743, opened 2020-07-23
Use EPOCH for GAIADR2 catalog positions

### LINKED ISSUE #740: Update GAIA catalog query to include date
_Issue [HLA-396](https://jira.stsci.edu/browse/HLA-396) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The standard pipeline astrometry processing relies on a web-service provided by ASB to return the GAIA catalog suitable for a given set of observations or a given pointing.  This service, though, does not return catalog positions appropriate for the observations date of observation, rather all coordinates are simply 2015.5 positions.  ASB (B. McLean) has plans to implement an update to the web-service to take the observations date (decimal year) and uses that to correct the GAIA catalog positions based on the GAIA catalog proper motions.  

Once this update has been made to the GAIA catalog web service, the astrometry code will need to be revised to compute and pass the observation's date (DATE-OBS) as a decimal year to the web service in 'create_astrometric_catalog'.  This should result in more accurate fits to GAIA, and perhaps even allow additional fields to match to GAIA that currently do not due to the proper motions affecting the cross-matching.

stsci-hack:
The functions for obtaining the astrometric catalog source positions from the web-service were updated to use the EPOCH parameter to obtain positions of the sources at the time of the observations.  This set of changes was implemented strictly for GAIADR2 catalog searches and any subsequent GAIA catalogs as they are the only ones with proper motions.  

In the process, the interpretation of the web-service results was changed from using the Python base 'csv' package to using
astropy.tables to directly interpret the results, vastly simplifying the code in the process.  The primary interface, create_astrometric_catalog, was also updated to allow the user to get ALL the columns from the web-service upon request, with the default being only to return the subset of columns needed for pipeline alignment (ra, dec, mag, objID).

This resolves #740 as well as [Jira Ticket HLA-396](https://jira.stsci.edu/browse/HLA-396).

### DIFF
--- a/drizzlepac/haputils/astrometric_utils.py
+++ b/drizzlepac/haputils/astrometric_utils.py
@@ -16,7 +16,6 @@ reference catalog. ::
 import os
 import pdb
 from io import BytesIO
-import csv
 import requests
 import inspect
 import sys
@@ -43,6 +42,7 @@ from astropy.nddata.bitmask import bitfield_to_boolean_mask
 from astropy.visualization import SqrtStretch
 from astropy.visualization.mpl_normalize import ImageNormalize
 from astropy.modeling.fitting import LevMarLSQFitter
+from astropy.time import Time
 
 import photutils  # needed to check version
 from photutils import detect_sources, source_properties, deblend_sources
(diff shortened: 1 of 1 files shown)


stsci-hack:
Changes were tested against dataset icw402010 which generates a successful fit to GAIADR1 and GAIADR2, with the DR2 fit being the final.

PR #753, opened 2020-07-29
Only try to generate QA JSONs for data with sources

stsci-hack:
These changes address issues identified during DSB regression testing when using the QA code with the astrometry/pipeline processing.  This change only tries to generate QA results (JSON files) for input exposures that actually had identifiable sources useful for alignment, since the code now relies on the pipeline processing results for the inputs data for the QA results.


stsci-hack:
In addition to the pipeline/astrometry QA bug fix, the SVM QA code also triggered Exceptions when trying to generate a comparison to GAIA but there are no GAIA sources for the field-of-view.  These Exceptions were seen during the '1000 dataset' SVM testing.

PR #771, opened 2020-08-19
Fix logic for SVM with multiple detectors and possiblity of detection of no sources

mdlpstsci:
- Ensure the phot_mode variable is reset to its input setting for every total product
  being processed.  The phot_mode can be modified during processing if one of the
  source detection algorithms does not find any sources.
- Bubble up the stack the "no sources detected" warning for the segmentation algorithm
  and return to the invoking algorithm.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -974,6 +974,13 @@ class HAPSegmentCatalog(HAPCatalogBase):
                                                               source_box=self._size_source_box,
                                                               mask=mask)
 
+            # Check if custom_segm_image is None indicating there are no detectable sources in the 
+            # total detection image.  If value is None, a warning has already been issued.  Issue
+            # a final message and return.
+            if custom_segm_img is None:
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -974,6 +974,13 @@ def identify_sources(self, **pars):
                                                               source_box=self._size_source_box,
                                                               mask=mask)
 
+            # Check if custom_segm_image is None indicating there are no detectable sources in the 
+            # total detection image.  If value is None, a warning has already been issued.  Issue
+            # a final message and return.
+            if custom_segm_img is None:
+                log.warning("End processing for the Segmentation Catalog due to no sources detected with Custom or Gaussian kernel.")
+                return None
```
stsci-hack, line 982: It is not clear that this needs to return any value at all, since this method was not designed to return a value upon success.  A simple empty return statement should be sufficient here.  Same for Line 1027 as well.

mdlpstsci: Remove unnecessary return value of None.

PR #773, opened 2020-08-19
Improved JSON data collection for "number of sources" statistics

mdlpstsci:
Improved comparison of sources found based on the white light image by protecting
against the cases where no sources are found and no catalog is actually written.
This allows the collection of data for the JSON files to continue as only the comparison
is skipped versus the entire evaluation function.

Name of the output number of sources JSON file was updated to match the convention
used by the majority of the other tests.

### DIFF
--- a/drizzlepac/haputils/svm_quality_analysis.py
+++ b/drizzlepac/haputils/svm_quality_analysis.py
@@ -210,17 +210,12 @@ def characterize_gaia_distribution(hap_obj, json_timestamp=None, json_time_since
 
 # ----------------------------------------------------------------------------------------------------------------------
 
-def compare_num_sources(catalog_list, drizzle_list, json_timestamp=None, json_time_since_epoch=None,
+def compare_num_sources(drizzle_list, json_timestamp=None, json_time_since_epoch=None,
                         log_level=logutil.logging.NOTSET):
     """Determine the number of viable sources actually listed in SVM output catalogs.
 
     Parameters
     ----------
-    catalog_list: list of strings
-        Set of files on which to actually perform comparison.  Catalogs, Point and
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -248,18 +243,25 @@ def compare_num_sources(catalog_list, drizzle_list, json_timestamp=None, json_ti
...
     seg_suffix = '_segment-cat.ecsv'
 
+    # Get a list of the catalog files (*.ecsv) in the current working directory
+    catalog_files = []
+    basepath = os.getcwd()
+    for entry in os.listdir(basepath):
+        if os.path.isfile(os.path.join(basepath, entry)):
+            if entry.endswith("ecsv"):
+                catalog_files.append(entry)
+
```
stsci-hack, line 253: For future reference, all these lines could replaced with:

`catalog_files = glob.glob('*.ecsv')`

drizzlepac/haputils/svm_quality_analysis.py
```diff
@@ -270,12 +272,11 @@ def compare_num_sources(catalog_list, drizzle_list, json_timestamp=None, json_ti
...
+            if catalog in catalog_files:
 
-            # if the catalog exists, open it and find the number of sources string
-            num_sources = -1
-            cat_type = ""
-            if does_exist:
+                # if the catalog exists, open it and find the number of sources string
+                num_sources = -1
+                cat_type = ""
                 file = open(catalog, 'r')
```
stsci-hack, line 280: For future reference, the best practice for working with file I/O is to use 'with open(catalog, 'r') as file'  and indent all the code related to 'file'.  This will automatically close the file handle upon completion of the code indented code.

PR #778, opened 2020-08-20
Fix dependencies

### LINKED ISSUE #777: Dependencies listed are wrong
You don't need `tables` and `nose`.

pllim:
Remove unused dependencies. Fix #777 

cc @jhunkeler and @rendinam

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,6 @@ setup(
         'astropy>=4.0.0',
         'fitsblender',
         'nictools',
-        'nose',
         'numpy>=1.18',
         'scipy',
         'matplotlib',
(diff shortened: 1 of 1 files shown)


stsci-hack:
The tables package was used to support generation of HDF5 files, which we now use as part of our pipeline processing testing code, code which (unfortunately) is not adequately included in the Azure tests run with each PR.  We need to insure that removing this dependency does not impact our ability to perform pipeline processing or the QA generation of pipeline products before this can be merged.

pllim:
But I don't see `tables` used anywhere within *this* package. Where is the code for "generation of HDF5 files" that actually uses `tables`?

pllim:
See #780 and #781

PR #780, opened 2020-08-20
MNT: Remove nose and move scikit-image dependencies

pllim:
A less controversial version of #778 by removing the things that you definitely are not using. Also see #777 

cc @jhunkeler and @rendinam

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,6 @@ setup(
         'astropy>=4.0.0',
         'fitsblender',
         'nictools',
-        'nose',
         'numpy>=1.18',
         'scipy',
         'matplotlib',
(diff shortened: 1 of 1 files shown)


pllim:
I don't think this failure is related?

`astroquery/mast/observations.py:687: NoResultsWarning: No products to download.`

stsci-hack:
The 'testing' code which was referred to in #778 was NOT part of the package tests or Azure tests, but rather an integrated system for generating quality & analysis (Q&A) results during integration testing under the DSB TEST environment (or equivalent).  It may be that the changes proposed in #778 would not have caused any problems, but it **REQUIRES MANUAL testing of the code** in order to verify properly, which this PR and #780 need to allow before any dependency related PR can be merged.

PR #781, opened 2020-08-20
MNT: Move tables to test dependencies

### LINKED ISSUE #777: Dependencies listed are wrong

pllim:
Second attempt of #778. Complements #780. cc @jhunkeler and @rendinam 

>  we now use as part of our pipeline processing testing code

Since it is used in testing code, then perhaps it should be optional dependency for testing only?

Fix #777

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,7 +56,8 @@ TESTS_REQUIRE = [
     'ci_watson',
     'crds',
     'pytest',
-    'pytest-remotedata'
+    'pytest-remotedata',
+    'tables',
 ]
 
 setup(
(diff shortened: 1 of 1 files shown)


pllim:
I don't think this failure is related?

`/astroquery/mast/observations.py:687: NoResultsWarning: No products to download.`

stsci-hack:
The 'testing' code which was referred to in #778 was NOT part of the package tests or Azure tests, but rather an integrated system for generating quality & analysis (Q&A) results during integration testing under the DSB TEST environment (or equivalent).  It may be that the changes proposed in #778 would not have caused any problems, but it **REQUIRES MANUAL testing of the code** in order to verify properly, which this PR and #780 need to allow before any dependency related PR can be merged.

PR #785, opened 2020-08-27
Use WHT array to refine point catalog

stsci-hack:
The weight array gets used in this set of changes to modify how the point source catalog gets computed.  It uses it to define which regions have nearly the same S/N (WHT) within a factor of 1.5.  The relative weight across the WHT array (relative to the max WHT value) gets computed and used to scale each region before computing the RMS.  This accounts for the lower S/N in lower WHT regions that causes background noise fluctuations to be scaled up in DRZ products.  This allows the algorithm to account for regions where fewer inputs contribute to the final output image.  

**This needs to be discussed and reviewed thoroughly before being merged.**  Initial results processing IBWB01 indicate that it significantly improves the actual detection rate by minimizing the number of faint, scaled background noise peaks identified as real sources.

### DIFF
--- a/drizzlepac/haputils/align_utils.py
+++ b/drizzlepac/haputils/align_utils.py
@@ -11,7 +11,7 @@ import numpy as np
 from scipy import ndimage
 
 from astropy.io import fits
-from astropy.table import Table, vstack
+from astropy.table import Table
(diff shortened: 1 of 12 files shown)


stsci-hack:
Added parameters to catalog json config files to allow control over the region generation algorithm.

PR #789, opened 2020-08-29
Add HAPLEVEL as attribute to HAPProduct class

### LINKED ISSUE #786: MVM: Fix implementation of HAPLEVEL keyword in MVM products
_Issue [HLA-412](https://jira.stsci.edu/browse/HLA-412) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The MVM products should contain the HAPLEVEL keyword in their headers, however, the first products generated did not have them despite being implemented in parts of the processing code.  This keyword is required in order to archive ingest to succeed and will need to be correctly added to all MVM product headers with appropriate values.

stsci-hack:
This change adds the HAPLEVEL keyword value as an attribute of the HAPProduct class.  A new method, find_member, was also added to the Filter, Exposure, and SkyCell product classes as well to support use from within the 'refine_product_headers' function.  These changes allow 'refine_product_headers' to add the correct value of the HAPLEVEL keyword to each of the drizzle products which gets created.  

In addition, the 'runmultihap' interface was fixed to actually call the 'hapmultisequencer' module for correctly processing MVM data.  

This will close #786.

### DIFF
--- a/drizzlepac/haputils/product.py
+++ b/drizzlepac/haputils/product.py
@@ -2,6 +2,14 @@
 
     Classes which define the total ("white light" image), filter, and exposure
     drizzle products.
+
+    These products represent different levels of processing with the levels noted in the
+    'HAPLEVEL' keyword.  The 'HAPLEVEL' values are:
+
+      * 1 : calibrated (FLT/FLC) input images and exposure level drizzle products with improved astrometry
+      * 2 : filter and total products combined using the improved astrometry,
+            consistent pixel scale, and oriented to North.
+      * 3 : (future) multi-visit mosaics aligned to common tangent plane
 """
 import logging
 import sys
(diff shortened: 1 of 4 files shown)


drizzlepac/haputils/product.py
```diff
@@ -62,6 +62,9 @@ def __init__(self, prop_id, obset_id, instrument, detector, filename, filetype,
         # this attribute is updated in the hapsequncer.py module (run_hla_processing()).
         self.configobj_pars = None
 
+        # Define HAPLEVEL value for this product
+        self.haplevel = 1
+
```
mdlpstsci, line 67: Please put some documentation into this module (at top?) which explains what haplevel means, in terms of product and associated numeric value.

mdlpstsci (changes requested):
Just a bit of documentation needed and this PR is ready to go.

drizzlepac/haputils/product.py
stsci-hack: Addressed in bc2ac06.

mdlpstsci (approved):

PR #794, opened 2020-09-02
Update runastrodriz to work under Windows

stsci-hack:
These fairly simple, and blunt, changes address problems that prevented `runastrodriz` from successfully executing to completion under Windows (the OS).  The changes are:
* insure that `num_cores` parameter value as set by the user gets passed to `astrodrizzle` when generating the a posteriori drizzle products 
* added `try...except` block around the deletion of the astrodrizzle log file in `run_driz`
* A note was added to the `runastrodriz` sphinx docs to suggest that users set `num_cores` to 1 when running under Windows due to problems with threaded logging (or, in fact, lack of any working logging when using multiprocessing as we have it implemented now in `astrodrizzle`). 

In addition, a few (PEP8) Flake8 syntax errors were also corrected.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -336,9 +336,9 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
 
             _cal_prodname = _infile
             _calfiles = [_infile]
-                
+
             _inlist = [_infile]
-                
+
             print("_calfiles initialized as: {}".format(_calfiles))
             if len(_calfiles) == 1 and "_raw" in _calfiles[0]:
                 _verify = False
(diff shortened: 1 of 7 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -1065,19 +1066,19 @@ def verify_alignment(inlist, calfiles, calfiles_flc, trlfile,
...
 def apply_headerlet(filename, headerlet_file, flcfile=None):
-    
-    # Use headerlet module to apply headerlet as PRIMARY WCS 
+
+    # Use headerlet module to apply headerlet as PRIMARY WCS
     headerlet.apply_headerlet_as_primary(filename, headerlet_file,
                                         attach=True, archive=True)
     # Verify that all keywords from headerlet got applied
-    hlet = headerlet.Headerlet.fromfile(headerlet_file)
+    # hlet = headerlet.Headerlet.fromfile(headerlet_file)
```
mdlpstsci, line 1074: Is this meant to be commented out?

mdlpstsci (changes requested):
One question which needs to be answered and otherwise this PR is ready for merging.

drizzlepac/runastrodriz.py
stsci-hack: Fixed under 11fdea6.

stsci-hack:
The changes in this PR now enforce single-CPU operations when running under Windows for both AstroDrizzle (single-drizzle step) and runsinglehap for the upcoming SVM processing.  This will need to be revised once the logging can be updated by resolving #799.

PR #798, opened 2020-09-08
Fixed display of flag values in catalog headers

mdlpstsci:
Fixed typo which caused documentation of flag values in catalog headers to be displayed improperly (e.g., not displaying some values).  Added documentation for flag 128.


stsci-hack (dismissed):
Simple enough fix for a simple enough mistake. 👍

stsci-hack (approved):
Still OK.

PR #810, opened 2020-09-21
Implement an alternative way to compute the background in support of Segmentation

mdlpstsci:
If the background subtracted science minus Background2D image is found to have too many negative values, an alternative method of determining the background is employed.  The sigma-clipped-stats() method in Astropy is used to determine the mean, median, and std for an input science image.  The median is then used to generate a two-dimensional image with the single value of median.  The std is used to generate a corresponding two-dimensional rms image. This was done to mitigate the "zones of avoidance" - regions where the segmentation algorithm could not find any sources.  

The catalog configuration files were updated to contain additional configuration values used to determine when the alternative algorithm should be used, and how the sigma-clipped background is determined.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -879,6 +879,14 @@ class HAPSegmentCatalog(HAPCatalogBase):
     """
     catalog_suffix = "_segment-cat.ecsv"
 
+    # Class variable which indicates to the Filter object the Total object had to determine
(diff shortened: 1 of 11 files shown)


stsci-hack (dismissed):
Looks good to me.

dulude (approved):

dulude (approved):
Looks good!

stsci-hack (approved):

PR #811, opened 2020-09-22
Utilize the weight image in the catalog creation algorithms

dulude:
- See JIRA ticket [HLA-411](https://jira.stsci.edu/browse/HLA-411)

**High-level summary**
- The 2-d source detection threshold array used for identification of sources for the segment catalog is now weighted by the relative fractional input image contribution (i. e. \<number of input images\>/\<max number of input images\> for any given pixel). This minimizes  background fluctuations in regions of the drizzle-combined image where there is low S/N due to there being relatively few contributing input images.
- This PR applies the functionality of  PR #785 to segment catalog source identification.

**Changelog**
- catalog_utils.py:
     - Moved creation of weight masks from HAPPointCatalog.identify_sources() (as local variable) to HAPCatalogBase.\_\_init\_\_() (as object attribute)
     - Segment catalog source detection threshold array now scaled by the relative weight masks computed by make_wht_masks().

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1,8 +1,9 @@
 """This script contains code to support creation of photometric sourcelists using two techniques:
 aperture photometry and segmentation-map based photometry."""
-import sys
-import pickle  # FIX Remove
+
 import copy
+import pickle  # FIX Remove
+import sys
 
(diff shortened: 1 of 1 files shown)


mdlpstsci (approved):
The algorithm has already been tested in the Point catalog processing.  All looks fine.

PR #815, opened 2020-09-25
Get single-image CR algorithm working

### LINKED ISSUE #795: Be judicious regarding the exposures which are used in the total detection image
_Issue [HLA-420](https://jira.stsci.edu/browse/HLA-420) was created on JIRA by [Michele De La Pena](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mdelapena):_

This issue is discussed in HLA-419 with regard to the cosmic ray rejection of the catalogs. The intent is to limit the exposures used in total detection image to avoid adding images with cosmic rays into the mix.  A snippet from HLA-419:

Recent versions of the HLA pipeline avoid creating detection images with a mix of N=1 and N>1 filters.  If all filters have N=1 exposures, then all exposures are included in the detection image.  If some filters have N>1 exposures, then *only the N>1 filters get included in the detection image._  Filters in such visits that have N=1 do _not* get included in the detection image.  But they still use the source lists created from the detection image for photometry on the N=1 filters.

stsci-hack:
This PR implements the following changes to the SVM processing code:

- apply single-image cosmic-ray(CR) rejection by updating DQ arrays for sole exposure that defines a given FilterProduct 
    - The algorithm is the same one used for alignment
- update single-image CR rejection algorithm to also identify most CR streaks (not delta CRs)
- update default parameters for astrodrizzle to NOT reset CR bits for single exposures or for the total detection image
- improve the memory use for the SkyFootprint algorithm
- fix a few more Flake8 syntax warnings/errors 

These changes were demonstrated to improve the total detection image, and the single image filter products using the data from the visit 'ib6w62'.

This will close #795 .

### DIFF
--- a/drizzlepac/haputils/align_utils.py
+++ b/drizzlepac/haputils/align_utils.py
@@ -113,7 +113,6 @@ class AlignmentTable:
         default_fwhm_set = False
 
         try:
-
             for img in self.process_list:
                 log.info("Adding {} to HAPImage list".format(img))
                 catimg = HAPImage(img)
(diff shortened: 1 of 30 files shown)


dulude (dismissed):
Looks good

PR #820, opened 2020-09-29
Refine segmentation algorithm

mdlpstsci:
Modified segmentation algorithm to use exclusion mask defined inHAPCatalogBase to remedy false source detections.  Combine some diagnostic output generation and some clean up.  Reset the negative threshold value to 0.0.


stsci-hack (approved):
You have to like it anytime code gets simpler and yet works better.  LGTM.

Issue #729, opened 2020-07-17
MVM: Set parameters based on input files


stscijgbot-hstdp:
_Issue [HLA-389](https://jira.stsci.edu/browse/HLA-389) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

MVM processing will involve creating output frames where some or many of the input images only minimally overlap if at all or where many, many frames overlap.  The parameters needed to produce the best possible combined sky cell mosaic depends critically on how many images overlap across the sky cell.  

The 'cell_utils' module provides the means to fairly easily generate the footprint of all the input exposures onto the sky cell pixels, including determining how many images contribute to each pixel in the sky cell.  Logic should be implemented to create this map and use statistics from this map like 'mean number of images/pixel' to determine what parameters to use for combining the images using 'AstroDrizzle'.  

 

Completion Criteria:
 * MVM processing can successfully select the "most appropriate" set of drizzle parameters for test sky cells which cover a couple of different coverage scenarios. 
 ** This does not require full MVM processing to be complete, although using the selected parameters to generate the output drizzle product would allow for verification that the parameters selected were sufficiently appropriate.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-389?focusedCommentId=476832#comment-476832):_

The `generate_footprint_mask()` method of the `HAPProduct` class in `haputils/product.py` already computes the basic parameters necessary for the initial implementation of this functionality for MVM processing.

PR #822, opened 2020-09-30
MVM: Set parameters based on input files

dulude:
**Relevant tickets and issues**
- See JIRA ticket[ HLA-389](https://jira.stsci.edu/browse/HLA-389)
- See Git Issue #729 

**Summary**
- product.HAPProduct.generate_footprint_mask() is utilized to generate a map of the number of contributing exposures for each pixel in each SkyCellProduct object in total_obj_list in hapmultisequencer.py. 
- The mean number of contributing exposures from non-zero value footprint pixels (rounded up or down to the nearest integer value) is then used to select the best set of AstroDrizzle parameters in config_utils.HapConfig._determine_conditions().


stsci-hack (approved):
Elegant set of changes that look fine to me.

PR #827, opened 2020-10-01
Re-organize/streamline background computation algorithms

mdlpstsci:
Re-organize and streamline the various background computation algorithms to provide a general resource for all source detection algorithms.  Move the majority of the existing code into the compute_background method of CatalogImage class, and clean up the residue in the HAP[Point | Segment]Catalog classes.  Updated the parameters files as necessary and appropriate.


stsci-hack (approved):
LGTM

PR #828, opened 2020-10-02
Fix a bug introduced when cleaning up two-dimensional images

mdlpstsci:
Changed or added where "del" is used in the compute_background image method for cleanup.  One of the del commands caused a bug when used in the mode of "simple_bkg=True".  Deleted some redundant code.


stsci-hack (dismissed):
Changes make sense.  LGTM.

stsci-hack (approved):
Subtle change in logic can certainly have a big effect in outcome.  These changes, though, appear to me to be fine.

PR #829, opened 2020-10-06
Try another fit to GAIA if necessary

stsci-hack:
Some datasets end up causing 'tweakwcs' to throw an Exception when attempting to perform a fit to GAIA after doing a relative fit amongst the input images.  This update adds logic to trap that Exception from `tweakwcs.align__wcs`, resets the AlignmentTable instance by calling the `configure_fit` method, then tries another fit using the `default` to the reference catalog.  

This is still not guaranteed to work for all datasets, but for visit `ibjc44`, it results in a successful fit finally being applied.

In addition, a very minor fix was also included to update the GAIA catalog DS9 `.reg` file generated in debug mode by making the column header row a comment.


mdlpstsci (approved):
Changes look fine and even better that the test dataset got a successful fit.  At the least, we are now catching an Exception!

PR #832, opened 2020-10-12
Refine auto kernel

stsci-hack:
These changes enable more control over how the code determines the FWHM and kernel from the PSFs actually found in the image being processed.  This includes:

- adds 'good_fwhm' range to the catalog generation config files as a controllable parameter
- increases the number of attempts to find a valid PSF by a factor of 10 

This was implemented to address problems with WFC3/IR data returning measured FWHM based on nearly saturated sources which apparently have a skewed PSF with a FWHM much greater than the actual PSF found for high S/N but non-saturated sources.   

This was tested using data from `icnw34` to return a FWHM closer to 1.1 instead of 3.45.

### DIFF
--- a/drizzlepac/haputils/astrometric_utils.py
+++ b/drizzlepac/haputils/astrometric_utils.py
@@ -498,7 +498,7 @@ def compute_2d_background(imgarr, box_size, win_size,
     return bkg_background, bkg_median, bkg_rms, bkg_rms_median
 
 def build_auto_kernel(imgarr, whtarr, fwhm=3.0, threshold=None, source_box=7,
-                      good_fwhm=[1.0, 4.0], num_fwhm=3,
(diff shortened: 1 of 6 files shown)


mdlpstsci (changes requested):
The good_fwhm values are now configurable via the configuration files which increases flexibility.  The values "in play" should be printed in the log for informational purposes.  (We all need to remember to do this!) Upon this simple change, this PR can be approved.

stsci-hack:
Review comments addressed in 7c262d7.  In addition, documentation was updated to provide a reference for the algorithm used for the verification code.  Finally, the threshold computation for the point catalog was updated to be consistent with the use of the WHT array regions as used by the segmentation catalog.

mdlpstsci (approved):

PR #834, opened 2020-10-14
Set the gain to a value of 1.0 when processing ACS/SBC data

mdlpstsci:
Ensure the photometry routine has a value of 1.0 for the ACS/SBC gain to avoid values being set to nan and causing
all of the sources to be trimmed from the output catalogs.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -496,6 +496,10 @@ class HAPCatalogBase:
         gain_values = [g for g in gain_keys if g > 0.0]
         self.gain = self.image.keyword_dict['exptime'] * np.mean(gain_values)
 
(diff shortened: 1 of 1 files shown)


stsci-hack:
Would it be simpler to add the if...else statement to the computation of `self.gain` in the init of the base class (Line 501?) for use in both places as once?  
May I also request that you also set this gain value to 1.0 for 'IR' as well as 'SBC'?  That would resolve problems with the photometry errors and swarm detection for WFC3/IR as well.

mdlpstsci:
Indeed, it would.  I guess I misinterpreted (or overinterpreted) our discussion in Slack as I thought you wanted it this way.  In any case, easily addressed.  Will also set the gain for for IR as well.

mdlpstsci:
Review addressed in commit 9666240.

stsci-hack (approved):
This looks quite efficient.  LGTM.

PR #835, opened 2020-10-15
Scale parameters relative to HLA plate scale

stsci-hack:
These changes automatically scale the pixel-based source flagging parameters based on the ratio of the HAP plate scale and the HLA plate scale used to derive the original set of parameters.   The parameter 'eppsky_limit' was also corrected to match the actual value used by the HLA.  

This, along with improvements to the FWHM computation and lower thresholds, allows the swarm flagging to work as designed for WFC3/IR data as demonstrated with data from visit `icnw34`.


mdlpstsci (approved):
Changes look fine.

PR #836, opened 2020-10-15
Write the Segment "regions" file (*.reg) for use with ds9 after the catalog sources have been trimmed

mdlpstsci:
Add code to write a "regions" file (*.reg) for use with ds9 in the segment write_catalog method to capture the trimmed data for overplot. Retain the previous "regions" file which is created just after all "sources" have been identified.  This file is now
*_all.reg.


stsci-hack (approved):
This looks perfectly fine to me.

PR #837, opened 2020-10-16
Add computation for the minimum allowable threshold 

mdlpstsci:
Added the computation for the minimum allowable threshold based upon the readnoise, gain, number of images, and total exposure time.  This minimum only applies for the sigma_clipped_stats algorithm where a comparison is done between the rms computed by the sigma_clipped_stats and the minimum allowable threshold. Clean up with respect to images written in diagnostic mode.  Fixed a bug where a residual computation of the threshold had not been deleted.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -159,8 +159,8 @@ class CatalogImage:
         maxiters : float, optional
             Parameter for the sigma_clipped_stats algorithm - number of sigma-clipping iterations to perform
 
-        Returns
-        -------
+        Attributes
+        ----------
         self.bkg_background_ra : 2D ndarray
             Background array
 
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -136,16 +136,43 @@ def compute_background(self, box_size, win_size,
...
+            Parameter for the sigma_clipped_stats algorithm - number of standard deviations to use for both
+            the lower and upper clipping limit.
+
+        maxiters : float, optional
+            Parameter for the sigma_clipped_stats algorithm - number of sigma-clipping iterations to perform
+
         Returns
         -------
-        bkg_background_ra : 2D ndarray
+        self.bkg_background_ra : 2D ndarray
```
stsci-hack, line 164: This doesn't make sense to me.  This implies that this method provides these updated values by explicitly passing that value back through the use of the 'return' command, yet, they are already attributes of the class and the code does not use the 'return' command.   It appears that this is meant more for documentation as to what attributes get updated with new values, but the 'Returns' section implies something very different.  Something different should be done here to avoid this confusion.

stsci-hack (changes requested):
Some docstrings need to be clarified, then this should be ready to merge.

stsci-hack (approved):

Issue #793, opened 2020-09-02
SVM: Part 2 - Rejection of catalogs due to cosmic ray contamination


stscijgbot-hstdp:
_Issue [HLA-419](https://jira.stsci.edu/browse/HLA-419) was created on JIRA by [Michele De La Pena](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mdelapena):_

The final step in evaluating the quality of the HAP catalogs for public availability is the cosmic ray contamination test. Details regarding this algorithm can be found here
<https://innerspace.stsci.edu/x/_htlAg> (updated discussion)
after the "Construct visit table" and noted as "step 2".

Information from Rick White's email:

 
The CR test is not applied to the WFC3/IR images or the ACS/SBC images.  Those images effectively do not have cosmic rays (at least, they are rare enough to ignore).  In ACS/SBC, a CR is just a single count (not a bright star), and in WFC3/IR nearly all CRs get removed by processing the multiple readouts.
 
The CR test has three parameters:
 
(1) the number of sources in the catalogs with flags<2 (good point and extended sources) **nPoint** and **nSegment**,
(2) the total exposure time of the detection image **ExposureTime**, and
(3) the total exposure time of N=1 (single-exposure) filters that contribute to the detection image, **n1ExposureTime**.
 
That last one can use a little more explanation.  If a visit consists entirely of filters with only 1 exposure each, **n1ExposureTime** is the same as **ExposureTime**.  If a detection image includes some single-exposure filters and some multiple-exposure (CR-split) images, then **n1ExposureTime** is just the sum of the exposure times for the single-exposure filters.  If all filters have multiple exposures, then **n1ExposureTime** = 0.
 
Recent versions of the HLA pipeline avoid creating detection images with a mix of N=1 and N>1 filters.  If all filters have N=1 exposures, then all exposures are included in the detection image.  If some filters have N>1 exposures, then *only the N>1 filters get included in the detection image._  Filters in such visits that have N=1 do _not* get included in the detection image.  But they still use the source lists created from the detection image for photometry on the N=1 filters.
 
The result of this strategy is that either **n1ExposureTime** = 0 (meaning there are no CRs in the detection image) or **n1ExposureTime** = **ExposureTime** (all contributing images have CRs).  If you assume this is true, then the algorithm becomes a little simpler.  We kept the more complicated version of the algorithm with both **n1ExposureTime** and **ExposureTime** because we still had some old catalogs lying around in the database.
 
[...]

Issue #804, opened 2020-09-14
SVM: Investigate scope of HLA-419 and HLA-420


stscijgbot-hstdp:
_Issue [HLA-427](https://jira.stsci.edu/browse/HLA-427) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

Investigate what effort will be necessary to implement HLA-419 and HLA-420 in order to be able to size the ticket effectively and with some level of confidence.  This will provide us with the information necessary to plan what sprints these tickets can be worked and when we can complete this work.

stsci-hack:
Resolved with #839.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-427?focusedCommentId=606127#comment-606127):_

Resolved with [Drizzlepac PR #&#x2060;839](https://github.com/spacetelescope/drizzlepac/pull/839) on 14-Sep-2020.

PR #839, opened 2020-10-20
Implement catalog rejection based on expected CR levels

stsci-hack:
These changes implement the logic used by the HLA to reject catalogs when they have been contaminated by cosmic-rays which have not been removed from the total detection image due to the presence of single-image filter products.  The basic logic has been described in [HLA #419](https://jira.stsci.edu/browse/HLA-419).  A concession, however, has been made for the use of single-image cosmic-ray identification implemented as #815 by scaling the exposure time of contributing single image filter products to only 5% of the actual exposure time.  This represents the residual signature of cosmic-rays from each N=1 product after applying #815.  Otherwise, this logic is identical to that used by the HLA and will result in NO catalogs being written out if the rejection threshold is met.


stsci-hack:
Will close #793 and #804 .

mdlpstsci (approved):

PR #844, opened 2020-10-27
Removed debug code from catalog_utils

stsci-hack:
A number of statements to write out intermediate products (for debugging and testing) needed to be removed.  In addition, logic needed to be added to accommodate the case when there are no sources after trimming to join with the total detection table.  

These changes fully address the errors encountered during testing on 2020-10-21 that exhibited the following Traceback messages:

`ib1f0b/HAP_ib1f0b.out-    sources.write(self.image.imgname.replace(drc, 'raw_sources.ecsv'), overwrite=True, format='ascii.ecsv')`
`ib1f0b/HAP_ib1f0b.out-AttributeError: 'NoneType' object has no attribute 'write'`

`j8ca07/HAP_j8ca07.out-    raise OSError(f"File {self.name!r} already exists.")`
`j8ca07/HAP_j8ca07.out-OSError: File 'hst_9029_07_acs_hrc_total_j8ca07_region1.fits' already exists.`

`j6m903/HAP_j6m903.out-    raise ValueError('input tables for join must both have at least one row')`
`j6m903/HAP_j6m903.out-ValueError: input tables for join must both have at least one row`

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -835,7 +835,7 @@ class HAPPointCatalog(HAPCatalogBase):
             for mask in self.tp_masks:
                 # apply mask for each separate range of WHT values
                 region = image * mask['mask']
-                fits.PrimaryHDU(data=region).writeto(self.image.imgname.replace(drc, 'region1.fits'))
+
                 # Compute separate threshold for each 'region'
                 reg_rms = self.image.bkg_rms_ra * np.sqrt(mask['mask'] / mask['rel_weight'].max())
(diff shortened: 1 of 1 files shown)


mdlpstsci (changes requested):
Just a question.  Does the verify_crthresh() return appropriately or throw an unhandled exception if the catalogs could not be combined in the previous step?

stsci-hack:
The logic in verify_crthresh accounts for the fact that the FLAG columns may not be present by returning an empty list of column names.  This allows the code to continue running with the default behavior of NOT rejecting the use of the catalogs.

PR #845, opened 2020-10-27
Implement env variables for catalog generation

### LINKED ISSUE #843: SVM: Add environment variables to control catalog generation
_Issue [HLA-444](https://jira.stsci.edu/browse/HLA-444) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

A set of environment variables needs to be added to the SVM processing code to provide control over whether or not the source catalogs are generated.  There should be a separate variable to control the catalog generation for each detector (instrument/detector combination).  

If this variable is NOT present or set to 'on', then the catalog will be created for that detector.  

The set of values used by this variable should be the same as the set used by the astrometry pipeline environment variables.  

Suggested environment variable names could follow the convention:

    SVM_CATALOG_<DETECTOR> = 'on'/'off'/'yes'/'no'

For example:  SVM_CATALOG_IR = 'off'  would turn off WFC3/IR catalog generation.

stsci-hack:
Implements environment variables for SVM catalog generation.  If no environment variables are found, it will default to generating the catalogs for that detector.  

This addresses [HLA-444](https://jira.stsci.edu/browse/HLA-444) and will close #843 .

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -15,6 +15,19 @@
     level chosen. The logger is acting as a gate on the messages which are allowed to be
     passed to the handlers.
 
+    Creation of source catalogs can be controlled through the use of environment variables:
+
+      - SVM_CATALOG_HRC
+      - SVM_CATALOG_SBC
+      - SVM_CATALOG_WFC
+      - SVM_CATALOG_UVIS
+      - SVM_CATALOG_IR
+
+    These variables can be defined using values of:
+
+      - 'on', 'true', 'yes' : Create catalogs
+      - 'off', 'false', 'no' : Turn off generation of catalogs
+
     The output products can be evaluated to determine the quality of the alignment and
     output data through the use of the environment variable:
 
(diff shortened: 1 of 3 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -92,6 +100,9 @@ def create_catalog_products(total_obj_list, log_level, diagnostic_mode=False, ph
         Which algorithm should be used to generate the sourcelists? 'aperture' for aperture photometry;
         'segment' for segment map photometry; 'both' for both 'segment' and 'aperture'. Default value is 'both'.
 
+    catalog_switches : dict, optional
+        Specify which, if any, catalogs should be generated at all, based on detector.
+
```
mdlpstsci, line 105: Please clarify the possible value of the switches and their corresponding meaning.

drizzlepac/hapsequencer.py
```diff
@@ -730,7 +751,7 @@ def run_sourcelist_flagging(filter_product_obj, filter_product_catalogs, log_lev
     return filter_product_catalogs
 
 
-def _get_envvar_switch(envvar_name):
+def _get_envvar_switch(envvar_name, default=None):
     """
     This private routine interprets the environment variable, SVM_QUALITY_TESTING,
     if specified.  NOTE: This is a copy of the routine in runastrodriz.py.  This
```
mdlpstsci, line 757: Update comment now that multiple env vars are being read.

mdlpstsci (changes requested):
Just clarify a few comment statements and looks good to go.

stsci-hack:
Docstrings have been updated in ce88be8 to address review comments.

mdlpstsci (approved):

drizzlepac/hapsequencer.py
stsci-hack: Updated in 65312f8.

drizzlepac/hapsequencer.py
stsci-hack: See changes in 65312f8.

PR #853, opened 2020-11-04
Initial draft of catalog rejection description

stsci-hack:
An initial draft of text describing the basic algorithm implemented to determine whether to reject the catalogs based on contributions from cosmic-rays.  This description includes the description of the single-image cosmic-ray rejection algorithm that is applied to the single image filter product exposures to partially mitigate the effects of cosmic-ray contamination in the total detection image.  

In addition, the API documentation for the runsinglehap interface was updated to include pointers to the current set of classes and functions in the primary SVM modules.  

Finally, a few small Flake8 (PEP8) corrections were included for some of the code being documented in the runsinglehap API docs.

### DIFF
--- a/doc/source/catalog_generation.rst
+++ b/doc/source/catalog_generation.rst
@@ -411,6 +411,9 @@ Also note that we reject both the point and segment catalogs if either one fails
 behind that is that since the catalogs are based on the same image, it is unlikely that one catalog will be
 good and the other contaminated.
 
+Should the catalogs fail this test, neither type of catalogs will be written out to disk for this visit.
+
+
 3.3.2 Single-image CR Rejection
 """"""""""""""""""""""""""""""""
 An algorithm has been implemented to identify and ignore cosmic-rays in single exposures.  This algorithm has
--- a/doc/source/catalog_generation.rst
+++ b/doc/source/catalog_generation.rst
@@ -359,6 +359,81 @@ ordering in the .ecsv file as well):
     * CI: Concentration index (MagAp1 – MagAp2), in AB magnitude
     * Flags: See Section 2.3.2 for flag value definitions
 
+3.3 Rejection of Cosmic-Ray Dominated Catalogs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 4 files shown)


doc/source/catalog_generation.rst
```diff
@@ -359,6 +359,84 @@ ordering in the .ecsv file as well):
...
+
+Also note that we reject both the point and segment catalogs if either one fails this test.  The reasoning
+behind that is that since the catalogs are based on the same image, it is unlikely that one catalog will be
+good and the other contaminated.
+
+Should the catalogs fail this test, neither type of catalogs will be written out to disk for this visit.
+
+
+3.3.2 Single-image CR Rejection
+""""""""""""""""""""""""""""""""
```
mdlpstsci, line 418: I would add the word Algorithm to the title of this subsection.

doc/source/catalog_generation.rst
```diff
@@ -359,6 +359,84 @@ ordering in the .ecsv file as well):
...
+effort to only publish catalogs which provide the highest science value, criteria developed by the Hubble Legacy
+Archive (HLA) has been implemented to recognize those catalogs dominated by cosmic-rays and not provided as an
+output product.
+
+.. note ::
+  This rejection criteria is NOT applied to WFC3/IR or ACS/SBC data since they are not affected by cosmic-rays
+  in the same way as the other detectors.
+
+3.3.1 Rejection Criteria
+"""""""""""""""""""""""""
```
mdlpstsci, line 377: I am having trouble deciding if the subsections should be in this order (3.3.1. vs 3.3.2) or in the reverse order.  I wonder if describing the algorithm first, then followed by the nitty-gritty details would be better.

Let us discuss.

mdlpstsci (changes requested):
Let us discuss this update in conjunction with the larger documentation discussion.

doc/source/catalog_generation.rst
stsci-hack: Done in 4bd2200.

doc/source/catalog_generation.rst
stsci-hack: Swapping the order of the 2 sections does make a bit more sense.  Done in 4bd2200.

mdlpstsci (approved):

PR #858, opened 2020-11-10
Add FINE and COARSE IR SkyCell layers

### LINKED ISSUE #819: MVM: Verify IR sky cells get defined with the correct plate scale
_Issue [HLA-434](https://jira.stsci.edu/browse/HLA-434) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The plate scale of WFC3/IR sky cells needs to be verified in order to make sure that the correct agreed upon plate scale will be used for the IR data; namely, 0.12arcsec/pixel.

stsci-hack:
These changes are intended to define both the 'fine' and 'coarse' layers for WFC3/IR sky cells.  

Closes #819 and [HLA-434](https://jira.stsci.edu/browse/HLA-434)

### DIFF
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -546,7 +546,7 @@ class ProjectionCell(object):
 
 class SkyCell(object):
 
-    def __init__(self, name=None, projection_cell=None, x=None, y=None, scale=None):
+    def __init__(self, name=None, projection_cell=None, x=None, y=None, scale="fine"):
         """Define sky cell at position x,y within projection cell.
 
         Parameters
@@ -596,7 +596,7 @@ class SkyCell(object):
 
         self.x_index = int(scell_id[6:8])
         self.y_index = int(scell_id[9:11])
-        self.projection_cell = ProjectionCell(index=pcell_id, scale=self.scale)
+        self.projection_cell = ProjectionCell(index=pcell_id)
         self.sky_cell_id = name
 
     def __repr__(self):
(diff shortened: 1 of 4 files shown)


stsci-hack:
Code still needs to be debugged to insure that it creates the correct layer definitions with the correct plate scales.  No review necessary at this time...

stsci-hack:
This is now ready for review...

drizzlepac/haputils/cell_utils.py
```diff
@@ -562,7 +588,7 @@ def __init__(self, name=None, projection_cell=None, x=None, y=None):
 
         self._build_wcs()
 
-    def _from_name(self, name):
+    def _from_name(self, name, scale=None):
```
mdlpstsci, line 591: I am not sure why scale is a parameter here.

stsci-hack: Residual from a previous attempt to account for different plate scales for the cells.  However, this turned out to be the wrong place to include the plate scales, and this vestigial parameter was left behind. It will be removed in the next commit to this PR.

drizzlepac/haputils/cell_utils.py
```diff
@@ -544,10 +546,34 @@ def plot(self, output=None, color='b'):
 
 class SkyCell(object):
 
-    def __init__(self, name=None, projection_cell=None, x=None, y=None):
+    def __init__(self, name=None, projection_cell=None, x=None, y=None, scale="fine"):
         """Define sky cell at position x,y within projection cell.
```
mdlpstsci, line 550: This constructor will work as long as either (1) name or (2) (projection_cell + x + y) are specified.  Items (1) and (2) are mutually exclusive.  This looks like it wants an overloaded constructor (which does not exist in Python).  I remember encountering this situation previously, but I cannot find my notes just now.  However, I did find this

https://medium.com/@shamir.stav_83310/overloading-constructors-in-python-17e42b7430b6

which discusses this same topic.  This solution eliminates conflict.

drizzlepac/haputils/cell_utils.py
stsci-hack: Removed in 0377f9a.

drizzlepac/haputils/cell_utils.py
stsci-hack: This will be addressed in a separate PR to address issue #868 (linked to [HLA-453](https://jira.stsci.edu/browse/HLA-453).

PR #864, opened 2020-11-16
MVM: output from poller utility should only contain specified sky cells when input from a "pipeline" poller file

mdlpstsci:
The "pipeline" MVM poller file specifies the skycell for each entry in the file.  In actuality, every entry has the same skycell id in the file.  The poller utility supports input not only from a poller file, but also a list of filenames where the skycell ids are computed.  The output table from the poller utility when MVM and input comes from a poller file, needs to be trimmed to contain only the specified sky cells.


stsci-hack (approved):
LGTM

PR #866, opened 2020-11-16
Populate HAP keywords in copied HAP input files

stsci-hack:
A number of keywords need to be updated/added to the HAP input files when they are created as copies of the original calibrated pipeline (FLT/FLC) files.  These changes include:

- updating 'FILENAME' with the new HAP filename to replace the original 'ipppssoot'-based filename
- adding the 'IPPPSSOO' keyword to the primary header 
- updating the 'HAPEXPNAME' column in the HAP DRZ product HDRTAB table with the basename, not full filename.

These changes address the issues identified in try to define the correct release date for the HAP products.

### DIFF
--- a/drizzlepac/haputils/product.py
+++ b/drizzlepac/haputils/product.py
@@ -522,7 +522,7 @@ class ExposureProduct(HAPProduct):
 
         # Add HAP keywords as required by pipeline processing
         with fits.open(edp_filename, mode='update') as edp_hdu:
-            edp_hdu[0].header['HAPLEVEL'] = (0, 'Classificaion level of this product')
+            edp_hdu[0].header['HAPLEVEL'] = (0, 'Classification level of this product')
             edp_hdu[0].header['IPPPSSOO'] = edp_hdu[0].header['ROOTNAME']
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/product.py
```diff
@@ -520,8 +520,11 @@ def copy_exposure(self, filename):
         except PermissionError:
             pass
 
-        # Add HAPLEVEL keyword as required by pipeline processing
-        fits.setval(edp_filename, 'HAPLEVEL', value=0, comment='Classificaion level of this product')
+        # Add HAP keywords as required by pipeline processing
+        with fits.open(edp_filename, mode='update') as edp_hdu:
+            edp_hdu[0].header['HAPLEVEL'] = (0, 'Classificaion level of this product')
```
mdlpstsci, line 525: Typo on Classification.

mdlpstsci (changes requested):
Just need to fix a typo.

drizzlepac/haputils/product.py
stsci-hack: Fixed in eb22c42

mdlpstsci (approved):

PR #867, opened 2020-11-18
WIP: Too many resources are used even when and MVM poller file is used as input

mdlpstsci:
Replaced cell_utils.get_sky_cells call which does time-consuming and
memory-consuming discovery with more rudimentary functions given that for
poller file input the output sky cell is already defined.


stsci-hack (dismissed):
That looks right to me.

stsci-hack (approved):

PR #872, opened 2020-11-19
Remove hard-coded values for the "good" FWHM search range 

mdlpstsci:
The hard-coded values for the FWHM search range in the build_kernel method of the CatalogImage class have been removed, and the values which are read from the detector-specific configuration files are used instead.  These values, as well as the default FWHM value used when no good candidate PSFs are found in the input image, are reported as log.info just before the call to build_auto_kernel.


stsci-hack (approved):
Simple enough.  I assume that this change did not affect the ability of the code to process successfully.

mdlpstsci:
Indeed, processing proceeded just fine for the two datasets tested.  Will do more tests to get results that can be used for other evaluations.

PR #873, opened 2020-11-30
Revise MVM manifest name to be more generic

stsci-hack:
This simple change makes the MVM manifest name more generic, removing the layer label, so that one manifest file will be more appropriately named for the entire input MVM poller file.  

This will partially address [HLA-455](https://jira.stsci.edu/browse/HLA-455).


mdlpstsci (approved):
This change looks fine.  The more specific string/file names are retained for use where needed.

Issue #728, opened 2020-07-17
MVM: Manage input data


stscijgbot-hstdp:
_Issue [HLA-388](https://jira.stsci.edu/browse/HLA-388) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

MVM processing will need to insure that all input files are available for processing.  This may require any or all of the following steps (listed in random order):
 * getting list of input files from 'poller' input file 
 * copying input files from cache directory
 * renaming input files using MVM-specific naming convention, as done with SVM processing
 * analyzing input files to see what can be aligned; ignoring moving target, grism, ... exposures.

This should follow much of the same logic done for SVM processing and use SVM developed code as much as practical.

Completion Criteria:
 * MVM processing can take a poller file for a sky cell and successfully process the input files **to allow** for successful MVM processing (full MVM processing not required for this ticket).

stscijgbot-hstdp:
_Comment by [Robert Swaters](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rswaters) on [JIRA](https://jira.stsci.edu/browse/HLA-388?focusedCommentId=489897#comment-489897):_

See HSTSDP-1040 for an example poller file. If changes are desired to the poller file, please let me know.

stscijgbot-hstdp:
_Comment by [Robert Swaters](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rswaters) on [JIRA](https://jira.stsci.edu/browse/HLA-388?focusedCommentId=493532#comment-493532):_

Lisa and I discussed a ticket she is about to start working, <https://jira.stsci.edu/browse/HSTSDP-1203.> In the process, we discovered we could reuse existing code and safe some work if we changed the poller file slightly. In HSTSDP-1040, the example says the format will be:
icnk20a8q_flc.fits,13773,CNK,20,515.0,F657N,UVIS,/path/to,skycell-p2381x05y09,NEW
If we changed the order of the last three columns to:
icnk20a8q_flc.fits,13773,CNK,20,515.0,F657N,UVIS,skycell-p2381x05y09,NEW,/path/to
If we did that, I'm hoping this would be a simple change on the HLA side?

stscijgbot-hstdp:
[...]

PR #874, opened 2020-12-01
Manage MVM input data

dulude:
Relevant JIRA Tickets/Git issues:
- JIRA ticket [HLA-388](https://jira.stsci.edu/browse/HLA-388)
- Git issue #728 

Summary of changes:
- poller_utils.py:
     - added code to build_poller_table() so that it can process MVM poller files with the new format.
     -Added logic to build_poller_table() to check that each file listed in the poller file exists in the current working directory. If a file is missing, it's copied over from the path specified in the poller file. Failing that, an exception is raised.

### DIFF
--- a/drizzlepac/haputils/poller_utils.py
+++ b/drizzlepac/haputils/poller_utils.py
@@ -5,10 +5,12 @@ poller, and produces a tree listing of the output products.  The function,
 parse_obset_tree, converts the tree into product catagories.
 
 """
+from collections import OrderedDict
 import os
 import pdb
+import shutil
 import sys
-from collections import OrderedDict
+
 import numpy as np
 
 from sklearn.cluster import KMeans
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/poller_utils.py
```diff
@@ -768,9 +771,30 @@ def build_poller_table(input, log_level, poller_type='svm'):
             # This code interprets that as False since it is NOT new, so the code
             # inverts the meaning from the pipeline poller file.
             if poller_type == 'mvm':
+                # Translate new format back to old format ("NEW" -> 0 and "OLD" -> 1)
+                poller_table_mapping = {"NEW": 0, "OLD": 1}
+                for tbl_ctr in range(0, len(input_table)):
+                    if input_table[tbl_ctr]['skycell_new'] in ["OLD", "NEW"]:
```
stsci-hack, line 777: There should be logic to trap the case where a 'corrupted' (hand-generated?) file was provided with mixed-case values or perhaps different strings other than 'OLD' and 'NEW' (all upper-case)?  Mixed case inputs can be handled by simply using '.upper()'.  Other problems, though, should immediately raise an Exception so that the user can edit the file to conform to supported values which get reported in the Exception ('We only support values of "OLD" and "NEW"...).

drizzlepac/haputils/poller_utils.py
```diff
@@ -768,9 +771,30 @@ def build_poller_table(input, log_level, poller_type='svm'):
...
+            for table_line in input_table:
+                if os.path.exists(table_line['filename']):
+                    log.info("Input image {} found in current working directory.".format(table_line['filename']))
+                elif os.path.exists(table_line['pathname']):
+                    log.info("Input image {} not found in current working directory. However, it was found in the path specified in the poller file.".format(table_line['filename']))
+                    shutil.copy(table_line['pathname'], os.getcwd())
+                    log.info("Input image {} copied to current working directory.".format(table_line['pathname']))
+                else:
+                    log.error("Input image {} not found in current working directory.".format(table_line['filename']))
+                    log.error("Input image {} not found.".format(table_line['pathname']))
```
stsci-hack, line 794: This log message should be a little more specific, as opposed to being confusingly similar to the previous line.  Something like 'Archived input image {} not found' might be a little clearer.

stsci-hack (changes requested):
Just a couple of error conditions which should be handled a little more robustly, and this will be ready to merge.

drizzlepac/haputils/poller_utils.py
dulude: Issue resolved in commit [66de19a](https://github.com/spacetelescope/drizzlepac/pull/874/commits/66de19adca258a7c5da14aec0846878d3d8c147e)

drizzlepac/haputils/poller_utils.py
dulude: Issue resolved in commit [79ca373](https://github.com/spacetelescope/drizzlepac/pull/874/commits/79ca373b5297551ecebfff51866b0554c551808a)

dulude:
> Just a couple of error conditions which should be handled a little more robustly, and this will be ready to merge.

Should be good to go.

stsci-hack (approved):
LGTM

PR #876, opened 2020-12-01
Rename static mask

stsci-hack:
These changes implement a new naming convention for the common static mask file that gets used by `astrodrizzle.AstroDrizzle()`.  The new name is based on the original input provided to astrodrizzle, and if it is ASN file, the filename defined in the ASN file as the 'output' product.  This results in a dataset unique name for the static mask based on the original input file which will not be confused with the static mask for any other dataset being processed in the same directory.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,9 @@ of the list).
 3.1.9 (unreleased)
 ==================
 
+- revise naming convention for the StaticMask file so that it has a
+  dataset-specific name instead of a generic common name. [#876]
+
 - Update ``runastrodriz`` to work under Windows while adding documentation
   to tell the user to run with ``num_cores`` set to 1.  [#794]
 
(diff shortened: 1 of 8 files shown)


mdlpstsci (approved):
Changes are straightforward and look to be complete.

mcara:
This, in general, looks good. One thing though: in my experience static mask files are usually left over on the disk (i.e., not "cleaned"). Am I wrong? If not, wouldn't this PR result in a large(r) number of files being left over on disk?

stsci-hack:
These static mask files are removed if the 'clean' parameter is set to 'yes'.  They do get left behind when processing does not complete as well.  However, these changes should at least associate these intermediate (temp) files with the dataset being processed so that they only get get cleaned up for the data which has been successfully processed.  I will try to see whether more can be done to clean these files up by identifying when they do get left behind, but that should be handled in a separate PR.

mcara (approved):
LGTM

PR #877, opened 2020-12-02
Rename static mask (#876)

stsci-hack:
* Rename common static mask file

* Insure static mask name is lower-cased

* update change log


stsci-hack:
Adding #876 to the release branch.

PR #878, opened 2020-12-02
Revise SkyCell API

stsci-hack:
This change to the SkyCell API removes the overloaded set of input parameters in favor of using a decorator classmethod.


mdlpstsci (approved):
Code looks fine.  Sorry, it took a bit, but I had to look up and understand the @classmethod decorator.

Issue #880, opened 2020-12-03
SVM: Address Segmentation algorithm for low number of sources identified


stscijgbot-hstdp:
_Issue [HLA-456](https://jira.stsci.edu/browse/HLA-456) was created on JIRA by [Michele De La Pena](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mdelapena):_

The Segmentation algorithm statistically (see Rick's analysis) finds far fewer sources than the Point algorithm.  This issue needs to be investigated and resolved.  The background level and/or associated threshold could be culprits, but it should be remembered both the Point and Segment algorithms use the same background determination.

stscijgbot-hstdp:
_Comment by [Robert Swaters](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rswaters) on [JIRA](https://jira.stsci.edu/browse/HLA-456?focusedCommentId=499219#comment-499219):_

1) We have a good set of test cases, ask Rick
2) Explore parameter space, in config files - does this improve the number of sources found? (Normalizing background already improved results, not enough)
3) May require code changes if config changes do not address the issue, but this would be under a new ticket
4) Testing for this ticket will be done with a smaller set, but we need to run the 1000 dataset test as well, possibly together with other tickets (e.g., HLA-457, HLA-458, HLA-443)

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-456?focusedCommentId=499404#comment-499404):_

Testing with dataset 'ia0m04' to determine that the background computation is flagged as returning over-subtracted data.  The tests were resulting only less than 300 sources being identified, while the HLA catalog returned over 1400 sources, and overlaying the catalogs on the image in DS9 confirmed that the difference comes from HAP not identifying the faintest of sources while the HLA catalog includes sources in the diffraction spikes and Airy rings as well.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-456?focusedCommentId=499458#comment-499458):_

This should be addressed with [Drizzlepac PR #&#x2060;895](https://github.com/spacetelescope/drizzlepac/pull/895).  

[...]

PR #895, opened 2020-12-08
Fix segment background determination and use

stsci-hack:
These changes revise the computation of the background and how it gets used for determining the sources in the segment catalog.  The changes include:

- using the -1 * RMS of the background as the negative threshold
- removing the 'negative_threshold' parameter from the config files
- updated computation of sigma-clipped background to be less influenced by bright sources 
- revising the Background2D parameters for the WFC3/IR format (smaller box size)
- reduced the source box size for the segment catalog sources for all detectors to pick up smaller/fainter sources
- eliminate background subtraction prior to calling photutils segmentation code (avoids applying the background twice)

These changes were tested using dataset 'ia0m04' (hst_11099_04, WFC3/IR).  Original HAP segment catalog had <300 sources compared to HLA catalog of >1400 sources.  Revised HAP segment catalog now has >1000 sources.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -252,24 +252,11 @@ class CatalogImage:
         if not is_zero_background_defined:
             log.info("")
             log.info("Computing the background using sigma-clipped statistics algorithm.")
-            bkg_mean_full, bkg_median_full, bkg_rms_full = sigma_clipped_stats(imgdata,
+            bkg_mean, bkg_median, bkg_rms = sigma_clipped_stats(imgdata,
                                                                 self.inv_footprint_mask,
                                                                 sigma=nsigma_clip,
-                                                                cenfunc='median',
-                                                                maxiters=maxiters)
+                                                                cenfunc='median')
 
-            log.info("FULL Sigma-clipped Statistics - Background mean: {}  median: {}  rms: {}".format(bkg_mean_full, bkg_median_full, bkg_rms_full))
-            # Refine background to better compute the median value
-            imgnz = imgdata.copy()
-            imgnz[self.inv_footprint_mask] = 0.0  # apply mask ahead of time
-            imgnz = imgnz[imgnz > 0.0]  # only want non-negative values
-            imgvals = imgnz[imgnz < (bkg_median_full + (bkg_rms_full*0.5))]
-            bkg_mean, bkg_median, bkg_rms = sigma_clipped_stats(imgvals,
-                                                                None,
-                                                                sigma=nsigma_clip,
-                                                                cenfunc='median',
-                                                                maxiters=maxiters)
-            del imgnz, imgvals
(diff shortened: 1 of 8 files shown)


stsci-hack:
Commit ff16070 fixes problems with the threshold computation for use with the RickerWavelet segmentation.  

This was initially tested using 'ib2t06' (hst_11603_06).  The catalog now actually comes out with a few more real sources than the HLA catalog.

mdlpstsci (changes requested):
The changes look fine and consistent.  The only issue is if the code is no longer using "imgarr_bkgsub" as it is really "imgarr" (the input data is no longer background subtracted), then the variable name and associated comments should be cleaned up to avoid confusion. I would agree to this being done now, or writing a new ticket and the clean up being done after this sprint in order to keep the current changes to the code to a minimum.

stsci-hack:
> The changes look fine and consistent. The only issue is if the code is no longer using "imgarr_bkgsub" as it is really "imgarr" (the input data is no longer background subtracted), then the variable name and associated comments should be cleaned up to avoid confusion. I would agree to this being done now, or writing a new ticket and the clean up being done after this sprint in order to keep the current changes to the code to a minimum.

This has been addressed in 8f44785.

mdlpstsci (changes requested):
As long as the variables and comments are being updated, the detect_and_deblend_sources() method also needs to be updated with regard to imgarr vs imgarr_bkgsub.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -299,15 +307,17 @@ def compute_background(self, box_size, win_size,
...
         # sigma-clipped algorithm, regardless of any other criterion. If simple_bkg == True,
         # the compute_background() is done, otherwise try to use Background2D to compute the background.
         if not simple_bkg and not is_zero_background_defined:
 
-            # If the sigma-clipped background image skew is less than the threshold,
-            # compute a two-dimensional background fit.
+            # If the sigma-clipped background image skew is greater than the threshold,
+            # compute a two-dimensional background fit.  A larger skew implies
+            # more sources in the field, which requires a more complex background.
             if bkg_skew < bkg_skew_threshold:
```
mdlpstsci, line 321: I believe the comment, and currently the bkg_skew_threshold = 0.5.  The code matches my notes from our discussion long ago, but the code seems backwards for when to compute the more complex 2D background fit. It seems the code should be 

if bkg_skew > bkg_skew_threshold:

mdlpstsci (changes requested):
Previous comment to update variable name in detect_and_deblend_sources as the parameter is now imgarr and not imgarr_bkgsub.

mdlpstsci (approved):

drizzlepac/haputils/catalog_utils.py
stsci-hack: Reversed in 9e1e7ed.  The logic for applying the skew in determining what background model to use was also inverted, so larger skews end up using 2D background and lower skews use sigma-clipped backgrounds.

stsci-hack:
Closes #880.

PR #898, opened 2020-12-09
Fixed two bugs during SVM testing: attribute not part of a class and error in joining tables

mdlpstsci:
The "big island" message which can be generated for the output catalog documentation section only applies for the segmentation algorithm.  The "if" statements were re-arranged.

Modified the join of specified filter columns to be added to the total catalog.  It is now a "left"
join (versus "inner") which means, in this case, the in-memory "total table" is always populated at this stage in the code.  If this table is empty, the algorithm has exited previously.   This change triggered unexpected behavior when generating the total Point catalog in that the RA and Dec columns were not necessarily totally populated as they were previously constructed from the filter in-memory tables.  Modified construction of Point total catalog with respect to the RA and Dec columns to ensure these columns are always present and totally populated.


stsci-hack (approved):
The logic seems right to me.

PR #902, opened 2020-12-22
Update Drizzlepac to work with non-IDC-updated WCSs

stsci-hack:
These changes allow AstroDrizzle to process data which has not been updated using the IDCTAB and avoids the Exception with 'wcs.idcscale / 3600.'.  The distorted WCS plate scale gets assigned as the 'model' plate scale 'wcs.idcscale' with these changes.   These changes allowed data from `j8eg060t0` to process without being updated by the IDCTAB reference file.  

The consequence of this change will be that data which has not been updated, and was expected to be updated, will not be using distortion model specified in the reference files.  This makes the WCSNAME keyword more important, and may impact re-processing done on the data either by the user or the automated pipeline.

### DIFF
--- a/drizzlepac/wcs_functions.py
+++ b/drizzlepac/wcs_functions.py
@@ -280,7 +280,13 @@ def make_outputwcs(imageObjectList, output, configObj=None, perfect=False):
                 cw.cpdis2 = None
                 cw.det2im = None
             undistort = False
+        # Insure that information is available for output_wcs to always work
+        for chip in chip_wcs:
+            if chip.idcscale is None:
+                chip.idcscale = chip.pscale
+
         hstwcs_list += chip_wcs
+        
     if not undistort and len(hstwcs_list) == 1:
         default_wcs = hstwcs_list[0].deepcopy()
     else:
--- a/drizzlepac/wcs_functions.py
+++ b/drizzlepac/wcs_functions.py
@@ -284,9 +284,11 @@ def make_outputwcs(imageObjectList, output, configObj=None, perfect=False):
         for chip in chip_wcs:
             if chip.idcscale is None:
                 chip.idcscale = chip.pscale
+                warn_str = "{}: Data from {} was NOT updated with the distortion model from {}!"
+                print(warn_str.format("WARNING", chip.filename, chip.idctab))
 
         hstwcs_list += chip_wcs
-        
+
     if not undistort and len(hstwcs_list) == 1:
         default_wcs = hstwcs_list[0].deepcopy()
     else:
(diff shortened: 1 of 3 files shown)


drizzlepac/wcs_functions.py
```diff
@@ -280,7 +280,15 @@ def make_outputwcs(imageObjectList, output, configObj=None, perfect=False):
                 cw.cpdis2 = None
                 cw.det2im = None
             undistort = False
+        # Insure that information is available for output_wcs to always work
+        for chip in chip_wcs:
+            if chip.idcscale is None:
+                chip.idcscale = chip.pscale
+                warn_str = "{}: Data from {} was NOT updated with the distortion model from {}!"
+                print(warn_str.format("WARNING", chip.filename, chip.idctab))
```
mcara, line 288:
```suggestion
                print(f"WARNING: Data from {chip.filename} was NOT updated with the distortion model from {chip.idctab}!")
```

mcara (approved):
LGTM

mcara:
I have approved this PR on technical merits. However, I have concerns about this approach. In particular, this PR makes it difficult to detect when an input dataset was not run through `updatewcs` and even advanced users may get results that are not what they expect. While this PR would log a warning message, I am afraid it would be easily missed in often very busy log files.

I would propose that instead of (mostly) quietly assigning `idcscale = pscale`, raise some custom exception such as:
```python
class UncalibratedDataError(ValueError):
    pass

..................

if chip.idcscale is None:  # or blot_wcs.idcscale
    raise UncalibratedDataError('AstroDrizzle expects data with full distortion models')
```
and then, in the pipeline, catch these exceptions like this:
```python
try:
    run_astrodrizzle(...)
except UncalibratedDataError:
    pass  # skip this data set
```

I discussed this offline with @stsci-hack and, if I understood correctly, another approach would be to detect when `IDCSCALE` is not set and do not process that data set.

In my opinion both approaches can and should be used: in the pipeline - detect this early and skip that data set, and when called by users - raise a meaningful exception for the user to read. Currently code often crashes with puzzling tracebacks.

mcara (changes requested):
While I have approved this PR before, I would like to ask you to reconsider this approach as I feel it would allow unnoticeable mistakes for users.

If you think this is the best approach I will approve this PR, but for now I will disapprove so that it cannot be merged accidentally.

stsci-hack:
This PR will NOT be merged, in favor of more appropriate solutions such as not processing images with DRIZCORR='OMIT' and selecting more up-to-date headerlets with a complete distortion model.

Issue #893, opened 2020-12-07
SVM: Address incomplete catalog due to bright sources (smiley faces)


stscijgbot-hstdp:
_Issue [HLA-467](https://jira.stsci.edu/browse/HLA-467) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

HLA flagging around bright sources often leads to not adding sources to the catalog.

New code has already been implemented for a new TinyTim PSF based source-finding algorithm to replace photutil's DAOFindStar/IRAFFindStar algorithms.  This new algorithm avoids (mostly?) identifying peaks in Airy rings and diffraction spikes as sources that were being flagged by the HLA flagging code as 'swarm' sources.   As a result, the HLA swarm flagging can be turned off (ideally).  This will resolve the problem with point catalogs resulting in rings of avoidance around bright sources (as seen in [the 'smiley face' example provided by INS on Slack ](https://stsci.slack.com/archives/GH3GNJ6QZ/p1605034180272600) for dataset `ib4103`).

This ticket's focus would be on calling the new source finding algorithm for point catalogs that use the new TinyTim source finding routine and verifying that it resolves the 'smiley face' catalog problem. If the problem persists in some datasets, then we will file a new ticket.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-467?focusedCommentId=503737#comment-503737):_

[Drizzlepac PR #&#x2060;903](https://github.com/spacetelescope/drizzlepac/pull/903) has been submitted to use the new PSF-based version of DAOFind to generate the point-source catalog by default.  This new algorithm is only another option which is selected by default for pipeline use, as opposed to replacing the previous algorithms.  The new code was originally developed using visit `ib4103` as the test dataset, and will need to be tested on additional visits to confirm its affect on the point-source catalog creation on all types of images from all detectors.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-467?focusedCommentId=507101#comment-507101):_

[...]

PR #903, opened 2020-12-23
Implement deconvolution-based point-catalog creation

stsci-hack:
This PR includes the first full implementation of an FFT-based deconvolution method for identifying point-sources.  It defines a new sub-class from `photutils` to replace DAOStarFinder with a new class that accepts a user-provided list of integer pixel positions as input to measure in the image using the DAOFind algorithms.  The user-provided positions are generated by converting the theoretical PSFs into undistorted PSFs which match the scale and orientation of the drizzle images.  This deconvolution is more of a refinement of what DAOStarFinder does by using a full PSF rather than simply relying on a model, although in this implementation, a model is still used to provide the final measured properties (including position).  

The extensions of photutils implemented here can serve as the basis for a photutils PR as well in the near future, something which the photutils developer has already agreed would be worthwhile.

The advantage of this method is that it reduces (almost completely eliminates, but not quite) SWARM detections.  Furthermore, the PSFs do not need to match the exact PSF found in the image, only be a close approximation, which is why the theoretical PSFs in the library created using TinyTim work well enough for identifying the sources to the nearest pixel.  

This code has been tested using single visit datasets `ia0m04`, 'ib4103' (smiley face point-catalog originally reported), `j92c01', and others.


stsci-hack:
One feature of this new source identification algorithm that came out as a result of the testing is that larger PSFs are needed for more heavily saturated image sources.  These larger PSFs need to be 2 as large in radius compared to the current set of PSFs in order to eliminate 'swarm'-sources farther from the core of the star.

stsci-hack:
Merging in order to enable more extensive testing.

stsci-hack:
Closes #893

PR #906, opened 2021-01-05
Add DRIZCORR as rejection criteria

stsci-hack:
This change will check to see whether the DRIZCORR header keyword has been set to "OMIT" or "SKIPPED" and reject SVM processing for those images.  This will avoid problems with incomplete or problematic IDCTAB use with the images which would cause an Exception to the thrown.


mdlpstsci (approved):
This is a straightforward update.

PR #907, opened 2021-01-07
Insure all processing results in updated WCSs

stsci-hack:
These changes address a couple of problems found during processing of HST data; specifically, j94vc1010 and jdgq05020.  

The changes include:

- adding WCSNAME=OPUS to the header if no IDCTAB WCS was created by `updatewcs` (due to NGOODPIX=0 or similar reasons identified by `updatewcs` itself)
- turning off checking of input files by `updatewcs` when performing the initial update to try to create IDC WCS for all input files insuring WCSNAME and IDCSCALE are defined even if data is no good for default pipeline solution, not just when checking the _a priori_ solutions.
- Update logic to insure the latest IDCTAB _a priori_ solution gets applied when checking _a priori_ solutions, not just the first _a priori_ WCS appended as a new extension

In addition, a minor change was also included to not try to delete the log files when running under Windows.  

This addresses the issues raised in HSTDP-1306.


drizzlepac/runastrodriz.py
```diff
@@ -473,7 +479,7 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
                 updatewcs.updatewcs(_calfiles, checkfiles=False)
                 if _calfiles_flc:
                     updatewcs.updatewcs(_calfiles_flc, checkfiles=False)
-
+            """
```
mdlpstsci, line 482: Was this meant to be removed?  If not, please add a comment as to why the code is commented out at this time vs being removed.

mdlpstsci (changes requested):
Just add some commentary as to why some code is commented out or remove the code.

drizzlepac/runastrodriz.py
stsci-hack: I still needed to insure that the changes to the code (using 'checkfiles=False' when updatewcs is first called) will remove the need for this code.  This code will be removed if testing with an appropriate dataset indicates there are no problems.  Problems with the testing will then come back to this code which may need to be uncommented again.

stsci-hack: This commented out code has been removed in ea92282.

mdlpstsci (approved):

Issue #900, opened 2020-12-10
Update pipeline drizzle products with S_REGION keyword


stscijgbot-hstdp:
_Issue [HLA-470](https://jira.stsci.edu/browse/HLA-470) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The standard pipeline processing of HST data with AstroDrizzle through `runastrodriz` should result in the addition of the 'S_REGION' keyword.  This will allow CAOM to populate the footprints for standard drizzle products from this keyword.  

This can be done by applying the `processing_utils.compute_sregion()` function to each final  Drizzle product before writing it out to disk.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-470?focusedCommentId=502944#comment-502944):_

[Drizzlepac PR #&#x2060;910](https://github.com/spacetelescope/drizzlepac/pull/910) has been submitted to implement this change.  The new keyword gets added to all calibrated FLT and FLC files before checking whether or not DRIZCORR was set to PERFORM.  This insures that all observations get this new keyword regardless of whether the data gets drizzled or not.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-470?focusedCommentId=502982#comment-502982):_

The question was raised about how to test this.  That can be handled best by it's own ticket.

PR #910, opened 2021-01-07
Add S_REGION to FLC files

stsci-hack:
This simple change adds the S_REGION keyword to all the SCI headers for all the calibrated input files (FLT and FLC files, as present).  The keyword will get added before checking whether or not the data has DRIZCORR turned on so that ALL exposures get this new keyword.  

The S_REGION keyword will then get picked up by CAOM when archiving these observations in order to define the footprint on the sky as presented through the archive interface. 

The changes were tested using the ASN 'j94vc1010'.


mdlpstsci (approved):

stsci-hack:
Changes were tested using both the ASN as input as well as using a single exposure filename as input (from the same ASN).

stsci-hack:
Closes #900.

mdlpstsci (approved):

Issue #817, opened 2020-09-29
MVM: Only generate sky cell layers being updated by input exposures


stscijgbot-hstdp:
_Issue [HLA-432](https://jira.stsci.edu/browse/HLA-432) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

MVM processing takes 2 lists of inputs:
* observations which need to be aligned to the projection cells
* observations which have ALREADY been aligned to the projection cell

Current MVM processing code will process all exposures, recreating all layers even if no new data was provided.  New logic needs to be added to recognize what layers are affected by the 'new' input exposures and ONLY align the data to the data in those layers and only run drizzle to generate those updated or new layers.

stscijgbot-hstdp:
_Comment by [Robert Swaters](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rswaters) on [JIRA](https://jira.stsci.edu/browse/HLA-432?focusedCommentId=488511#comment-488511):_

Here's an example (with fake content) of what the poller file will look like. The skycell will also be in the poller file name.
ib3p01r9q_flc.fits,11635,B3P,01,1492.0,F469N,UVIS,/ifs/archive/dev/hst/info/rswaters/test/test_rswaters_20200506_05/regr_08120/pipe/science/hap/ib3p01r9q_flc.fits,<skycell>,old
ib3p01raq_flc.fits,11635,B3P,01,1492.0,F469N,UVIS,/ifs/archive/dev/hst/info/rswaters/test/test_rswaters_20200506_05/regr_08120/pipe/science/hap/ib3p01raq_flc.fits,<skycell>,old
ib3p01rbq_flc.fits,11635,B3P,01,1561.0,F469N,UVIS,/ifs/archive/dev/hst/info/rswaters/test/test_rswaters_20200506_05/regr_08120/pipe/science/hap/ib3p01rbq_flc.fits,<skycell>,new
ib3p01rdq_flc.fits,11635,B3P,01,1561.0,F469N,UVIS,/ifs/archive/dev/hst/info/rswaters/test/test_rswaters_20200506_05/regr_08120/pipe/science/hap/ib3p01rdq_flc.fits,<skycell>,new

stsci-hack:
Resolved with #911.

PR #911, opened 2021-01-08
MVM: Only generate sky cell layers being updated by input exposures

dulude:
**Relevant Tickets and Issues**
- JIRA Ticket [HLA-432](https://jira.stsci.edu/browse/HLA-432)
-  Git Issue #817 

**Summary of Changes**
- poller_utils.py: added logic to build_poller_table() to remove images listed as "OLD" in the input poller file from input_table so that they will omitted from MVM processing.
- cell_utils.py: updated return value definition of Skycell.from_name() to use skycell cell object, rather than int for input arg 'projection_cell'.


drizzlepac/haputils/poller_utils.py
```diff
@@ -7,6 +7,7 @@
 """
 from collections import OrderedDict
 import os
+import pdb
```
stsci-hack, line 10: This appears to be a leftover from debugging and can be removed.

stsci-hack (dismissed):
Aside from the extra debug statement, it looks good.

drizzlepac/haputils/poller_utils.py
dulude: issue resolved by commit [aca89a7](https://github.com/spacetelescope/drizzlepac/pull/911/commits/aca89a7d2ae66605db0e3e372742388e74991a3d)

stsci-hack (approved):

PR #915, opened 2021-01-15
update version of python to >=3.7

nden:
The next release of astropy will not support Python 3.6. This PR changes the min Python version to
` >= 3.7`. 
Changing the version will help also with running the regression tests with dev dependencies.
Is there a reason not to update?
This is in support of AL-453.


nden:
This PR revealed that drizzlepac does not run with python 3.9 because a dependency (tables 3.6.1) failed to build with python 3.9. The latest tables release is from Oct, 2019. Is it still supported? This is a separate issue which needs to be addressed in the future.

nden:
@stsci-hack This is ready for review

stsci-hack (approved):

PR #916, opened 2021-01-18
fix align test, update Jenkins job title

nden:
This PR allows the tests in `test_align.py` to run so comparison with `truth` results can be done.
With this the Jenkins tests are now running and failing with trivial differences. The truth files need to be updated.


stsci-hack (approved):
Nice simple corrections that look good to me.

Issue #909, opened 2021-01-07
Support updated IDCTAB WCSs


stscijgbot-hstdp:
_Issue [HLA-474](https://jira.stsci.edu/browse/HLA-474) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The logic in standard pipeline processing of HST imaging data turns out to not correctly use the latest distortion model information when the IDCTAB reference file specified in the header is newer than the IDCTABs used for WCSs retrieved from the astrometry database.  

This logic needs to be revised to be based on the latest distortion model specified for an exposure/file when computing the 'best' alignment available, especially when performing a posteriori fitting. 

INS input will be required to define this new logic to meet their expectations for pipeline processing.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-474?focusedCommentId=502888#comment-502888):_

A discussion with INS (Mack, Bajaj, and Avila) came to the following decisions:
 * the pipeline-default should always (and currently does) get defined based on the IDCTAB specified in the header (the latest, newest)
 * the a priori solution should be defined based on the newest IDCTAB
 ** For data taken after Oct-2017, the pipeline-default solution is a valid a priori solution based on GSC240, so the pipeline-default can be used as-is (assuming it passes its own verification of relative alignment)
 ** For data taken before Oct-2017, the offset applied to create the GSC240 solution based on a previous IDCTAB should be extracted and used to create a new GSC240 solution and run through verification of relative alignment. 
 ** Failure to verify alignment would cause pipeline-default to be used instead.
 * The a posteriori fit should be performed based on the same IDCTAB as the pipeline-default WCS.  
 ** Failure to align with an a posteriori fit would default to verified a priori or pipeline-default based on newest IDCTAB.

The discussion also came to the decision to request a change to the astrometry database to only return the solutions based on the most recent distortion model by default, and a separate switch/parameter would be defined to request all WCSs including those based on older IDCTABs.  It is not clear whether this is viable, but we will work with ASB to see what is possible.  

The updated logic will need to be implemented in 'runastrodriz' for basic pipeline use.  Data taken before the new IDCTAB was started to be used operationally will most likely need to be reprocessed with these changes in order to insure that the latest IDCTAB gets used to define the WCS for the 'older' data.

[...]

Issue #170, opened 2018-08-11
DOC: Fix failing RTD build


pllim:
Looks like RTD has been failing for over a month -- https://readthedocs.org/projects/drizzlepac/builds/

pllim:
Might be same problem as https://github.com/spacetelescope/jwst/issues/2323#issuecomment-422057746

xref astroconda/astroconda-contrib#404

PR #922, opened 2021-01-22
Create new apriori WCS for new IDCTAB

stsci-hack:
This addresses #909 by recognizing when no solution returned by the astrometric WCS database was based on the IDCTAB specified in the image header.  In that situation, these changes compute a new 'apriori' WCS based on the offsets stored in the guide star database and applies that as the new Primary WCS.  

This newly created apriori WCS also gets saved as a stand-alone headerlet so that the astrometric database can ingest the new solution for subsequent reprocessing efforts using the same distortion model.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -1209,7 +1209,11 @@ def restore_pipeline_default(files):
 
 
 def apply_apriori_wcs(calfiles, calfiles_flc, trlmsg, gsc_catalog='GSC240'):
-    """ Apply apriori WCS's from astrometry database based on currently defined IDCTAB"""
+    """ Apply apriori WCS's from astrometry database based on currently defined IDCTAB """
+    #
+    # TODO:  Have this code only update the headers with the WCSs based on the IDCTAB specified
+    #        in the image header (the 'latest' model).
+    #
     # Determine what IDCTAB reference file has been specified in the header
     idctab = fits.getval(calfiles[0], 'idctab')
     # Define:  refval - env var for dir with IDCTAB
@@ -1223,6 +1227,7 @@ def apply_apriori_wcs(calfiles, calfiles_flc, trlmsg, gsc_catalog='GSC240'):
         idcroot = idctab.split('_')[0]
         refval = None
 
+    del refval
     rootname = fits.getval(calfiles[0], 'rootname')
 
     # Query the astrometry database to determine what distortion models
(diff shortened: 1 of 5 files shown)


stsci-hack:
After discussion with Nadia, it was determined that it would be simpler to put this computation directly in 'stwcs.updatewcs' instead.  Doing so would co-locate more of the logic for determining the WCS.  So, this PR will need to be revised to move a lot of the code to 'stwcs'.

stsci-hack:
The logic in runastrodriz was revised in the following ways:

- It uses the updated version of `stwcs.updatewcs.updatewcs()` which enforces the use of the IDCTAB specified in the header and computes a new a priori WCS if none matching the IDCTAB can be found in the astrometry database.  This means that `runastrodriz()` no longer needed the code `verify_gaia_wcsnames()` since `updatewcs()` now can be trusted. 
- If an a posteriori WCS was returned by `updatewcs()`, then it is assumed to be accurate already and it gets used without any further verification.  

This version also REQUIRES STWCS > 1.6.2a in order to run, since it relies on the functionality from [STWCS PR #170](https://github.com/spacetelescope/stwcs/pull/170).

drizzlepac/runastrodriz.py
```diff
@@ -453,29 +452,36 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
...
+
+        # run updatewcs with use_db=True to insure all products have
+        # have a priori solutions as extensions
+        # This is NOT done in the apriori sub-directory in order to insure that all
+        # relevant solutions from the astrometry database get appended to the input FLT/FLC files.
+        #
+        updatewcs.updatewcs(_calfiles, use_db=True, all_wcs=False)
+        if _calfiles_flc:
+            _trlmsg += "Adding apriori WCS solutions to {}\n".format(_calfiles_flc)
+            updatewcs.updatewcs(_calfiles_flc, use_db=True, all_wcs=False)
```
mdlpstsci, line 468: Should there be a message the for the 'a priori' solutions being added to the _calfiles in addition to the the message for the _calfiles_flc just so the user does not have to assume it is being done to the flt by default?

drizzlepac/runastrodriz.py
```diff
@@ -453,29 +452,36 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
...
+        if align_dicts is not None:
+            find_crs = not align_dicts[0]['alignment_verified']
+        else:
+            find_crs = False
+
+        # run updatewcs with use_db=True to insure all products have
+        # have a priori solutions as extensions
+        # This is NOT done in the apriori sub-directory in order to insure that all
+        # relevant solutions from the astrometry database get appended to the input FLT/FLC files.
+        #
```
mdlpstsci, line 464: Since I was momentarily confused regarding the comment beginning at Line 473, perhaps it would be good to add a comment before the use of updatewcs reminding folks that all of the "returned" WCS solutions are based upon the "newest" distortion model by default (all_wcs=False).

mdlpstsci (changes requested):
Really only one request to add a comment as to the functionality of updatewcs.  My other remarks are comments.

stsci-hack:
The code for creating new apriori WCS solutions for new IDCTABs was moved to STWCS.  These changes should NOT be merged in.

Issue #925, opened 2021-01-26
Update HDRTAB for EXP DRC HAP products for CAOM


stscijgbot-hstdp:
_Issue [HLA-483](https://jira.stsci.edu/browse/HLA-483) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The HAPEXPNAME list needs to provide the rootname (drop the _flc.fits) for the exposure level input files that were used to create the exposure level drizzle products.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-483?focusedCommentId=507011#comment-507011):_

This issue has been resolved by merging [Drizzlepac PR #&#x2060;926](https://github.com/spacetelescope/drizzlepac/pull/926).

PR #926, opened 2021-01-26
Fix HAPEXPNAME in HDRTAB for EXP HAP Products

stsci-hack:
This makes the logic consistent for all HAP Drizzle products when updating the HDRTAB extension so that the HAPEXPNAME column contains the product_basename, not full filename, as the value. 

These changes were tested by processing visit 'hst_11603_06' (ib2t06).  

This addresses HLA-483 and the related issue #925 .


mcara:
`HAPEXPNAME` sounds like some pharmaceuticals or out of organic chemistry book.

PR #929, opened 2021-02-01
Set a common active WCS for direct as well as corresponding Grism/Prism images

mdlpstsci:
Accommodate processing Grism/Prism images in the form of the SVM hst*[flt/flc].fits) at least to the stage where the Grism/Prism SVM image and the direct SVM images have a common active WCS.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -54,9 +54,11 @@ import sys
 import traceback
 
 import numpy as np
+from astropy.io import fits
 from astropy.table import Table
 
 import drizzlepac
+from drizzlepac import util
 from drizzlepac.haputils import config_utils
 from drizzlepac.haputils import diagnostic_utils
 from drizzlepac.haputils import hla_flag_filter
@@ -67,8 +69,9 @@ from drizzlepac.haputils import svm_quality_analysis as svm_qa
 from drizzlepac.haputils.catalog_utils import HAPCatalogs
 
 from stsci.tools import logutil
+from stwcs import updatewcs
 from stwcs import wcsutil
-
+from stwcs.wcsutil import headerlet
 
 __taskname__ = 'hapsequencer'
 MSG_DATEFMT = '%Y%j%H%M%S'
@@ -525,7 +528,6 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
     cat_switches = {sw: _get_envvar_switch(sw, default=envvar_cat_svm[sw]) for sw in envvar_cat_svm}
 
     total_obj_list = []
-    product_list = []
     try:
         # Parse the poller file and generate the the obs_info_dict, as well as the total detection
         # product lists which contain the ExposureProduct, FilterProduct, and TotalProduct objects
@@ -573,6 +575,14 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
             reference_catalog = run_align_to_gaia(total_item, log_level=log_level, diagnostic_mode=diagnostic_mode)
             if reference_catalog:
                 product_list += reference_catalog
+
+        # If there are Grism/Prism images present in this visit, as well as corresponding direct images
+        # for the same detector, update the primary WCS in the direct and/or Grism/Prism images as
+        # appropriate to be an 'a priori' or the pipeline default (fallback) solution
+        for total_item in total_obj_list:
+            if total_item.grism_edp_list and total_item.edp_list:
+                grism_flt_list = update_wcs_grism_in_visit(total_item)
+                product_list += grism_flt_list
 
         # Run AstroDrizzle to produce drizzle-combined products
         log.info("\n{}: Create drizzled imagery products.".format(str(datetime.datetime.now())))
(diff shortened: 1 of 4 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -812,3 +825,222 @@ def _get_envvar_switch(envvar_name, default=None):
...
+    # Loop over all the direct images for this detector in the visit
+    for edp in tdp.edp_list:
+        filename = edp.full_filename
+        hdu = fits.open(filename)
+
+        # Make sure the WCS is up-to-date - doing the update for the
+        # direct images here so there is no need to modify a
+        # pre-existing class which could cause an issue
+        drizcorr = hdu[0].header['DRIZCORR']
+        if drizcorr == "OMIT":
```
stsci-hack, line 942: It is unclear why this logic is included.  Why limit the WCSs to pipeline-only WCSs when you expect no other WCSs to have been added to the file already? If DRIZCORR was OMIT, we are almost guaranteed that 'updatewcs' has not been run on the data and no a priori or pipeline-default WCSs have been added to the file or SCI header.  The only way an a priori WCS solution will be appended to a direct image is using 'updatewcs(filename, use_db=True)'.  When 'use_db=False', you are restricting the WCSs being included in the SCI header to only the pipeline-default, unless the file has already been processed previously with 'updatewcs(filename, use_db=True)', something which is not guaranteed.  Even with 'use_db=True', the pipeline-default WCS will be added to the SCI header, even if it is just as an alternate WCS.

drizzlepac/hapsequencer.py
```diff
@@ -812,3 +825,222 @@ def _get_envvar_switch(envvar_name, default=None):
         switch_val = envvar_bool_dict[default] if default else None
 
     return switch_val
+
+# ------------------------------------------------------------------------------
+
+
+def update_wcs_grism_in_visit(tdp):
+    """Examine entries in the total data produce for Grism/Prism data
```
stsci-hack, line 833: Typo?
```suggestion
    """Examine entries in the total data product for Grism/Prism data
```

drizzlepac/hapsequencer.py
```diff
@@ -812,3 +825,222 @@ def _get_envvar_switch(envvar_name, default=None):
...
+            # from the current image
+            else:
+                direct_wcs_set &= set(dict_names.values())
+
+            # Oops...no common wcsnames
+            if not direct_wcs_set:
+                log.error("There are no common WCS solutions with this direct image {} and previously processed images".format(filename))
+                log.error("    There is a problem with this direct image/visit.")
+                sys.exit(1)
+        # If there are no WCS solutions, the image could be bad (e.g., EXPTIME=0 or EXPFLAG="TDF-DOWN...")
```
stsci-hack, line 966: ...or the input image was a _raw.fits file.  This is the most unlikely prospect, but never underestimate users. ;-)

drizzlepac/hapsequencer.py
```diff
@@ -812,3 +825,222 @@ def _get_envvar_switch(envvar_name, default=None):
...
+
+    Parameters
+    ----------
+    tdp : total data product (TotalProduct) object
+        Object comprised of FilterProduct, ExposureProduct, and GrismExposureProduct
+
+    Returns
+    -------
+    grism_product_list : list
+        List of all the SVM Grism/Prism FLT/FLC files generated
```
stsci-hack, line 852:
```suggestion
        List of all the SVM Grism/Prism FLT/FLC files updated with the common WCS
```

drizzlepac/hapsequencer.py
```diff
@@ -812,3 +825,222 @@ def _get_envvar_switch(envvar_name, default=None):
...
+    filename : str
+        Output SVM FLT/FLC filename
+
+    wcsname : str
+        Name of the desired WCS active/primary solution to be set for the filename
+    """
+    # For exposures with multiple science extensions (multiple chips),
+    # generate a combined WCS
+    num_sci_ext, extname = util.count_sci_extensions(filename)
+    extname_list = []
```
stsci-hack, line 1040: Another way to create extname_list (lines 1040-1042) would be to use list comprehension with:

extname_list = [(extname, x+1) for x in range(num_sci_ext)]

drizzlepac/hapsequencer.py
```diff
@@ -812,3 +825,222 @@ def _get_envvar_switch(envvar_name, default=None):
...
+    for x in range(num_sci_ext):
+        extname_list.append((extname, x+1))
+
+    hdu = fits.open(filename, mode="update")
+    wcsutil.altwcs.restoreWCS(filename, ext=extname_list, wcsname=wcsname)
```
stsci-hack, line 1045: If you really wanted to go dense with the code, you could replace 'ext=extname_list' with the list-comprehension form of  extname_list.  ;-)  This is NOT something you need to do, but merely pointing out a way to reduce the amount of code that needs to be debugged and maintained.

drizzlepac/haputils/analyze.py
```diff
@@ -287,12 +287,19 @@ def analyze_data(input_file_list, log_level=logutil.logging.NOTSET):
...
-        split_sfilter = sfilter.split('_')
+        #
+        # Grism/Prism images are also IMAGING=SPECTROSCOPIC, suppress the "no processing"
+        # indicators to allow the Grism/Prism images to be minimally processed for
+        # keyword updates.  This was done as a retrofit to allow Grism/Prism images
+        # to have the same WCS solution as the direct images in the visit (same detector).
+        split_sfilter = sfilter.upper().split('_')
         for item in split_sfilter:
-            if item.startswith(('G', 'PR', 'BLOCK')):
+            if item.startswith(('G', 'PR')):
```
stsci-hack, line 299: It is not clear to me what purpose this block of code for 'G', 'PR' options serves.  Any reason it can not simply be removed?

drizzlepac/haputils/product.py
```diff
@@ -529,6 +538,81 @@ def copy_exposure(self, filename):
...
+        self.filters = filters
+        self.full_filename = self.copy_exposure(filename)
+
+        # Open the input FITS file to mine some header information.
+        # and make sure the WCS is up-to-date
+        hdu_list = fits.open(filename)
+        self.mjdutc = hdu_list[0].header['EXPSTART']
+        self.exptime = hdu_list[0].header['EXPTIME']
+
+        drizcorr = hdu_list[0].header['DRIZCORR']
```
stsci-hack, line 559: See previous notes on my questions about the use of this logic in hapsequencer.

stsci-hack (changes requested):
There were a few comments, and a couple of typos, that should be addressed before it can be merged.  Overall, clean simple logic that is easy to follow and maintain.

drizzlepac/hapsequencer.py
mdlpstsci: Fixed.

drizzlepac/hapsequencer.py
mdlpstsci: Fixed.

drizzlepac/hapsequencer.py
mdlpstsci: Cool - done.

drizzlepac/hapsequencer.py
mdlpstsci: Too much density makes my head explode.  Pass on this.

drizzlepac/haputils/analyze.py
mdlpstsci: This code is needed here in order to process the Grism/Prism images.  The "filter" check is done last and is not part of the extended if/elif block.  The Grism/Prism images are also SPECTROSCOPIC, so even if we do not exclude the images based upon filter, they would be excluded based upon imaging mode.  To process the Grism/Prism data, I make sure the "do not process" sentinel is cleared.

drizzlepac/hapsequencer.py
mdlpstsci: Ack.  I did not consider this.  I added a message regarding use of *_raw.fits files.

drizzlepac/hapsequencer.py
mdlpstsci: Changed the if/else to just an if: if drizcorr is equal to OMIT, then updatewcs with use_db=True.

drizzlepac/haputils/product.py
mdlpstsci: Same as above
Changed the if/else to just an if: if drizcorr is equal to OMIT, then updatewcs with use_db=True.

mdlpstsci:
Updates to accommodate a majority of the reviewer's comments are included in the 53fdf3d commit.  This ticket also does some "special" processing for Quad and Ramp filters.  These particular images are only processed as far as the "align to Gaia" stage.  There are no drizzle products for these images thought the SVM version of the FLT/FLC images have been updated as compared to the pipeline FLT/FLC products.

drizzlepac/haputils/poller_utils.py
```diff
@@ -634,35 +664,61 @@ def parse_obset_tree(det_tree, log_level):
...
+            for item in tdp.grism_edp_list:
+                try:
+                    os.remove(item.full_filename)
+                except OSError:
+                    pass
+            index_to_delete.append(index)
+            tdp.grism_edp_list.clear()
+            del tdp.grism_edp_list[:]
+    for item in index_to_delete:
+        del tdp_list[item]
```
stsci-hack, line 714: Whenever you are deleting multiple items from a list, it should be done in reverse order (last items deleted first).  Think about `a = ['x', 'y', 'z', '0', '1', '2']` and you want to delete 'z' and '1' at indices [2, 4].  If you do `for item in [2,4]: del a[item]', you will end up with `a = ['x','y','0','1']`, but if you do 'for item in [4,2]: del a[item]', you will get what you want.  I am afraid the deletion in 713 and 714 will run into problems with this problem.  Please check whether you need to add `index_to_delete.reverse()` before line 713.

stsci-hack (commented):
Just one last item to check...

drizzlepac/haputils/poller_utils.py
mdlpstsci: Indeed.  I was trying to be careful about this very issue, but worried more about deletion of the Grism list attribute in the object than the Total object from Total product list!  Worked correctly in testing as the list only had two items.  Fixed in git commit 1fce265.

stsci-hack (approved):

PR #941, opened 2021-02-16
Clean up manifest file and disk as necessary

mdlpstsci:
Fix to ensure the literal 'None' and the name of the 'metawcs all ref catalog' strings are not in the manifest.  In addition the
'metawcs fall ref catalog' file should be deleted from disk.  The disk clean up may have already happened, but this is to ensure it is done and is encapsulated in a try/except.


stsci-hack (approved):
Simple enough.  I wish I could remember why the refname was originally included in the output, but I don't see any problems with this change. Approved!

PR #943, opened 2021-02-18
Update alignment functions docstrings

stsci-hack:
This PR adds API documentation for the functions in ``drizzlepac.align`` to the RTD docs, and updates the docstrings for the functions in the RTD docs in that module to be more complete.


drizzlepac/align.py
```diff
@@ -202,6 +218,24 @@ def perform_align(input_list, archive=False, clobber=False, debug=False, update_
...
+    catalog_list : list
+        Set of astrometric catalogs which should be used as references for fitting the input images.  A
+        separate fit will be performed for each catalog specified. The catalog name will also be used
+        as part of the output `WCSNAME` value for the fit determined from that catalog.
+
+    fit_label : str
+        String to use as a unique tag for the WCS solutions generated by these fits.  This tag will be
+        inserted between the `-FIT` and the catalog name in the `WCSNAME` keyword value; for example,
+        `fit_label="User"` will result in fits to GAIADR2 with names ending in `-FIT-User-GAIADR2`.  
+
```
mdlpstsci, line 230: -FIT-User-GAIADR2  should be -FIT_User_GAIADR2

mdlpstsci (changes requested):
One typo to address.

drizzlepac/align.py
stsci-hack: Fixed in 05e19ef.

mdlpstsci (approved):

PR #944, opened 2021-02-18
Store FITS keyword WCS solutions as a dictionary for use in multiple routines

mdlpstsci:
Keep track of the WCS solutions which are defined in FITS keywords, not headerlets, in a dictionary to associate the solutions property with its image.  Since a Grism/Prism image my be acquired with another filter (ACS only) yielding a "filter" name in the processing code as g800l_f814w, for example, make sure to check the entire filter string for 'g' or 'pr' for Grism/Prism images.


stsci-hack (approved):
Pretty straight-forward set of changes.

PR #945, opened 2021-02-22
Corrected manner of checking for a return type of None

mdlpstsci:
Simple fix to check for a return type having no values.  Replace the check for a length of zero which causes an exception to "if not variable_name".


stsci-hack:
This PR is no longer needed given the changes to this section of code in #950.

PR #946, opened 2021-02-22
Handle situation of no viable Total Data Products in the visit

mdlpstsci:
If a visit only contains Grism or Prism data and no direct images, for example, there is no reason to process the visit.  When Grism/Prism and direct exposures are present, it is desired for all of the images in the visit from the same detector to have the same WCS.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -544,6 +544,14 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
         log.info("Parse the poller and determine what exposures need to be combined into separate products.\n")
         obs_info_dict, total_obj_list = poller_utils.interpret_obset_input(input_filename, log_level)
 
+        # It is possible the total_obj_list output from the poller_utils is an empty list.  An example
+        # of this would be a visit with only Grism/Prism and no direct images.  If this is the case,
+        # there is actually nothing to be done for the visit.
+        if not total_obj_list:
+            log.warning("")
+            log.warning("There are no viable Total Data Products in this visit so processing is exiting.")
+            sys.exit(0)
+
         # Generate the name for the manifest file which is for the entire visit.  It is fine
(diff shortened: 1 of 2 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -542,6 +542,14 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
         log.info("Parse the poller and determine what exposures need to be combined into separate products.\n")
         obs_info_dict, total_obj_list = poller_utils.interpret_obset_input(input_filename, log_level)
 
+        # It is possible the total_obj_list output from the poller_utils is an empty list.  An example
+        # of this would be a visit with only Grism/Prism and no direct images.  If this is the case,
+        # there is actually nothing to be done for the visit.
+        if not total_obj_list:
```
stsci-hack, line 548: This logic occurs inside a `try...except...finally` block which this change circumvents.  The `finally` block manages the (not inconsequential) task of closing down the logging as well as returning the status variable to the calling routine.  It would be better to see whether these operations can be handled even in the instance that no inputs are suitable for processing.

stsci-hack (commented):
This change is fine as far as it goes, however, an answer to my comment would insure the code quits cleanly.

drizzlepac/hapsequencer.py
mdlpstsci: Understood and have to think about this.  The logging is all shutdown properly with the sys.exit(0) or trapping the exception and re-raising the exception with an informative message.  However, no log (trl) file is written as there is no viable filename for the log to use.  As it stands, the temporary log filename (poller filename with the suffix of log) has all the processing information in it up to this point.  Will think of a non-hacky way to handle this.

mdlpstsci: Addressed in changes for commit 646528d.

stsci-hack (approved):
LGTM.

PR #950, opened 2021-02-25
Fix PSF point-catalog memory issues, plus more

stsci-hack:
Testing on 1082 datasets uncovered a number of problems with the new PSF-based point-catalog generation code.  These changes include:
  - logic to insure that the computed background for catalogs has a median > 0
  - computes the skew based on the statistics of the entire field-of-view, not just the 'background' pixels
  - adds logic to weight the WHT arrays in `make_wht_arrays()` to be based on number of input exposures
  - implements a fail-over for the PSF-based catalog generation to use the original DAOFindStar in case of problems with matching the PSFs to the field (typically due to extreme CTE that was not corrected in subarray observations)
  - simplified the threshold use in the PSF deconvolution
  - implemented a block-by-block (default: 1024x1024) FFT of the drizzled exposure to reduce the memory usage
  - Flake8 corrections in some places
  - updated the library of PSFs to not only avoid artifacts which can be induced by TinyTim (tiny3, specifically), but also doubled the radius of the WFC3/IR PSFs to avoid more swarm features from being detected as sources in the catalog 
  - logic for selecting PSFs was also revised to avoid problem with exposures only have "CLEAR" filters in place.
  - deletes PSFs used during processing.
These changes should correct the errors reported during the 19-Feb 1k test of this package.


drizzlepac/haputils/catalog_utils.py
```diff
@@ -909,34 +926,54 @@ def identify_sources(self, **pars):
...
+                                                                 box_size=self.param_dict['region_size'],
+                                                                 mask=self.image.footprint_mask,
+                                                                 block_size=self.param_dict['block_size'],
+                                                                 diagnostic_mode=self.diagnostic_mode)
+                    except Exception:
+                        # In case we run into out-of-memory error, or any other exception with
+                        # PSF use (like CTE or horribly mismatched PSFs), fail-over to using
+                        # DAOFind mode instead
+                        log.warning("Exception thrown when trying to use PSFs to find sources with UserStarFinder.")
                         user_peaks = None
```
mdlpstsci, line 942: Should source_fwhm have a default value as it gets used in Line 955?

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -66,10 +68,6 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
         This value should result in selecting a frequency
         one to two orders of magnitude below the largest spectral component of the point-spread function.
```
mdlpstsci, line 70: Add documentation for the block_size parameter.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -83,44 +81,59 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
...
-        centered_psf = psf * 0.0
-        centered_psf[psf_center[0] - psf_len: psf_center[0] + psf_len,
-                     psf_center[1] - psf_len: psf_center[1] + psf_len] = psf[psf_y - psf_len: psf_y + psf_len,
-                                                                             psf_x - psf_len: psf_x + psf_len]
-        psf = centered_psf
-
-    # Insure input image also has no NaN values
-    img = np.nan_to_num(img, 0.0)
+    # Insure input image also has no NaN values, only if necessary
+    # This should be less memory-intensive that np.isnan()
```
mdlpstsci, line 94: Typo (I think): that ==> than

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -131,127 +144,252 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
...
+    zero_mask = (img_block > 0).astype(np.int16)
+    m_map *= zero_mask
+
     return m_map
 
-# ======================================================================================================================
 
-# Functions to manage PSF library for deconvolution
+def _center_psf(psf, img_block_shape):
+    """Create centered PSF image with same shape as img_block"""
```
mdlpstsci, line 154: If this is really a public method, please add description of arguments.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -131,127 +144,252 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
...
+        block = (block, block)
+    new_shape = (arr.shape[0] + (block[0] - (arr.shape[0] % block[0])),
+                 arr.shape[1] + (block[1] - (arr.shape[1] % block[1])))
+    new_arr = np.zeros(new_shape, dtype=arr.dtype)
+    new_arr[:arr.shape[0], :arr.shape[1]] = arr
+    return new_arr
+
+
+def create_blocks(arr, block_size=(1024, 1024)):
+    if arr.shape[0] % block_size[0] != 0 or arr.shape[1] % block_size[1] != 0:
```
mdlpstsci, line 189: Same comment...if public method...

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -639,44 +748,98 @@ def find_stars(self, data, mask=None):
 # Main user interface
 #
 # -----------------------------------------------------------------------------
-def find_point_sources(drzname, data=None, mask=None, nsigma=5.0, box_size=11, sigma=2.0):
+def find_point_sources(drzname, data=None, mask=None,
+                       box_size=11, block_size=(1024, 1024),
```
mdlpstsci, line 752: Same as above...if public method...

mdlpstsci (changes requested):
Some comments to clarify.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -131,127 +144,252 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
...
+    psf_len = max(max(psf_max[0][0] // 2, psf_size[0]), max(psf_max[1][0] // 2, psf_size[1]))
+
+    centered_psf[psf_center[0] - psf_len: psf_center[0] + psf_len,
+                 psf_center[1] - psf_len: psf_center[1] + psf_len] = psf[psf_y - psf_len: psf_y + psf_len,
+                                                                         psf_x - psf_len: psf_x + psf_len]
+    return centered_psf
+
+
+def pad_arr(arr, block=1024):
+    """ Zero-pad an input array up to an integer number of blocks in each dimension"""
```
mdlpstsci, line 178: If this is really a public method, please add description of arguments.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -131,127 +144,252 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
...
+            slices.append([a, b, slice(block_size[0] * a, block_size[0] * (a + 1)),
+                                 slice(block_size[1] * b, block_size[1] * (b + 1))])
+
+    del new_arr
+    return {'slices': slices, 'new_shape': new_shape, 'blocks': blocks}
+
+
+def rebuild_arr(block_arr, slices, new_shape, output_shape):
+    """Convert convolved blocks into full array that matches original input array"""
+    out_arr = np.zeros(new_shape, dtype=block_arr.dtype)
```
mdlpstsci, line 212: If this is really a public method, please add description of arguments.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -131,127 +144,252 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
...
+        for i in indx:
+            del filter_list[i]
+
+    # Now select the widest-band filter used for this observation
+    # This accounts for cross-filter usage.
+    found_filter = False
+    if len(filter_list[2:]) >= 1:
+        # Loop over types of bandpass based on filter names
+        # going from widest band to narrowest bands
+        bandpass = ['lp', 'w', 'm', 'n']
```
mdlpstsci, line 329: Maybe I am lost in the weeds at this point, but is this code meant to handle SBC too?  If so, SBC has a filter "BLOCK" which is for internal calibration.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -131,127 +144,252 @@ def fft_deconv_img(img, psf, freq_limit=0.95, sigma=None):
...
+    # Drizzle PSF FLT file to match orientation and plate scale of drizzled science (total detection) image
+    astrodrizzle.AstroDrizzle(input=psf_flt_names,
+                              output=psf_drz_name,
+                              **drizzle_pars)
 
-    log.info("Using Library PSF: {}".format(os.path.basename(psf_name)))
+    if clean_psfs:
+        # clean up intermediate files
+        for psf_name in psf_flt_names:
+            os.remove(psf_name)
```
mdlpstsci, line 383: Cool. This was called out in HLA-493, but I see it getting cleaned now.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -264,7 +402,6 @@ def convert_library_psf(calimg, drzimg, psf,
     # This will be the name of the new file containing the library PSF that will be drizzled to
     # match the input iamge `drzimg`
     psf_flt_name = psf_root.replace('.fits', '_psf_flt.fits')
```
mdlpstsci, line 404: Typo: iamge ==> image

mdlpstsci (changes requested):
May take another look, but only some requests for additional commentary.

drizzlepac/haputils/catalog_utils.py
stsci-hack: This part of the code (specifically, Lines 929-941) only overrides the default already provided at the start of the method.

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Fixed in d7c635f

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Primary description at top edited to add a note about this change.

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Fixed in d7c635f

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Fixed in d7c635f

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Added in d7c635f

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Added in d7c635f

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Added in d7c635f

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Added in d7c635f

mdlpstsci (approved):
Documentation improved so ready for the 1k test suite.

PR #951, opened 2021-02-26
Redo S_REGION computation

stsci-hack:
This PR revises how the S_REGION gets computed for images.  Instead of relying on the `HSTWCS.calc_footprint()` method which only returns the corners of the array, these changes look for the corners of the non-zero data in the image and reports the RA/Dec of those pixels instead.


rlwastro:
I think this algorithm of finding the max and min corners of the individual exposures will not be accurate enough for the footprints.  As an example, here is an image of the footprint for hst_13826_11_acs_wfc_f775w_jcol11, which has 5 exposures in an "X" shape.  The plot shows the HLA footprint (inner red) and the current HAP footprint (outer green square).  There is no simple rectangle (at any rotation) that will capture the actual shape of the footprint.  The actual HLA footprint in this case has 24 points.  This gets much more complicated with mosaic image footprints, which can have multiple separate islands.

hst_13826_11_acs_wfc_f775w_jcol11

stsci-hack:
This latest version (1a4d896) produces a set of positions which match (qualitatively, at least) the HLA outline, as demonstrated for hst_13826_11_acs_wfc_f775w_jcol11 as well as data from visit ICOL05 (hst_13826_05_wfc3_ir_total_icol05).  The yellow represents the actual exposed pixels from all the inputs, and the red outline and points are the computed region pixels.  

This update also includes a small utility function for interpreting the S_REGION keyword as an array of RA/Dec positions.    

**This code still needs to be debugged to work against images where there is a gap between the chips in the final product.**

SREGION_icol05_mask_vs_coords
SREGION_icol05_mask_vs_coords_closeup1
SREGION_jcol11_mask_vs_coords
SREGION_jcol11_mask_vs_coords_closeup1

stsci-hack:
The changes in a5794c3 support detecting regions in multiple un-connected chips in the mosaic image, without affecting the results for the previously reported computations.  The changes for this commit were tested using a single ACS/WFC exposure as seen in these plots.  
SREGION_jcol11itq_mask_vs_coords
SREGION_jcol11itq_mask_vs_coords_closeup1

rlwastro (dismissed):
This looks great to me!  Your tests look good, and the approach appears robust.

drizzlepac/haputils/cell_utils.py
```diff
@@ -648,6 +812,375 @@ def build_mask(self):
...
+            end_corner = start_edge['sky_edge'][1]
+
+        # At this point, both the starting edge and end corner are defined.
+        # Now start building up region polygon from all exterior segments
+        # starting from this edge
+        remaining_segments -= 1
+        region.append(start_edge)
+        region_end = None
+
+        while region_end != end_corner and remaining_segments > 0:
```
mdlpstsci, line 929: I guess I am sleepy, but what is the purpose of this while loop?

drizzlepac/haputils/cell_utils.py
```diff
@@ -648,6 +812,375 @@ def build_mask(self):
...
+    phi = np.arctan2(y, x)
+    deg = 90. - np.rad2deg(phi) if clockwise else np.rad2deg(phi) - 90.
+    deg[deg < 0.] += 360.
+    return(rho, phi, deg)
+
+def pol2cart(rho, phi):
+    x = rho * np.cos(phi)
+    y = rho * np.sin(phi)
+    return(x, y)
+
```
mdlpstsci, line 945: Just a comment.  I am surprised about the lack of library functions perform these conversions.

mdlpstsci (changes requested):
The methods in the SkyFootprint class could use a bit more explanation of their purpose.

drizzlepac/haputils/cell_utils.py
stsci-hack: Astropy has functions to convert from cartesian to spherical, but not simple polar coordinates like I needed here.  Numpy also did not have anything, since they figured people can work out the 2 lines of code needed.

drizzlepac/haputils/cell_utils.py
stsci-hack: This whole class ended up being un-used and so it was deleted in 07d2e887.

drizzlepac/haputils/cell_utils.py
```diff
@@ -240,9 +261,14 @@ def find_footprint(self, member='total'):
...
-            mask = self.exp_masks[member]
+            mask = self.exp_masks[member]['mask']
         self.footprint = np.clip(mask, 0, 1)
-        # self.footprint = np.clip(self.total_mask, 0, 1)
+
+        # Shrink footprint by 1 pixel, then invert.
+        # This results in a mask where pixels inside the footprint (but not
+        # the very edge) will be False.
+        # This will allow for easy identification of any pixel interior to the footprint
+        self.inside = np.invert(morphology.binary_erosion(self.footprint))
```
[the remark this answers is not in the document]
stsci-hack: All references to 'morphology' were converted to simply 'ndimage' in 07d2e887.

drizzlepac/haputils/cell_utils.py
```diff
@@ -648,6 +812,375 @@ def build_mask(self):
 
         self.mask = mask
 
+
+class SkyCorners(object):
+    def __init__(self):
```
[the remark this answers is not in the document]
stsci-hack: Docstrings were added or updated to all methods in 07d2e887.

mdlpstsci (approved):

Issue #862, opened 2020-11-13
Drizzled images contain WCS distortion keywords in their headers


mcara:
Below is a fragment of a `'SCI',1` header from a drizzled image (created with `build=True` option). This header contains distortion terms that should not be present in a drizzled image header:

```
XTENSION= 'IMAGE   '           / Image extension                                
BITPIX  =                  -32 / array data type                                
NAXIS   =                    2 / number of array dimensions                     
NAXIS1  =                 4224                                                  
NAXIS2  =                 4245                                                  
PCOUNT  =                    0 / number of parameters                           
GCOUNT  =                    1 / number of groups                               
INHERIT =                    T / inherit the primary header                     
EXTNAME = 'SCI     '           / extension name                                 
EXTVER  =                    1                                                  
BUNIT   = 'ELECTRONS/S'        / Units of science product                       
                                                                                
CD1_1   = -4.2987619123789E-06                                                  
CD1_2   = -1.3206887619290E-05                                                  
CD2_1   = -1.3206887619290E-05                                                  
CD2_2   = 4.29876191237893E-06                                                  
CRVAL1  =    114.6886400110381                                                  
CRVAL2  =    65.49437411415843                                                  
CRPIX1  =               2112.0                                                  
CRPIX2  =               2122.5                                                  
CTYPE1  = 'RA---TAN'                                                            
CTYPE2  = 'DEC--TAN'                                                            
ORIENTAT=    -71.9702450148845                                                  
VAFACTOR=                  1.0                                                  

RADESYS = 'ICRS    '           / Equatorial coordinate system                   
D2IM1   = 'EXTVER: 1' / Version number of WCSDVARR extension                    
D2IM1   = 'NAXES: 2' / Number of independent variables in D2IM function         
D2IM1   = 'AXIS.1: 1' / Axis number of the 1st variable in a D2IM function      
D2IM1   = 'AXIS.2: 2' / Axis number of the 2nd variable in a D2IM function      
D2IM2   = 'EXTVER: 2' / Version number of WCSDVARR extension                    
D2IM2   = 'NAXES: 2' / Number of independent variables in D2IM function         
D2IM2   = 'AXIS.1: 1' / Axis number of the 1st variable in a D2IM function      
D2IM2   = 'AXIS.2: 2' / Axis number of the 2nd variable in a D2IM function      
MJDREFI =                  0.0 / [d] MJD of fiducial time, integer part         
[...]

PR #541, opened 2019-12-10
Add automated HLA classic sourcelist comparison

dulude:
See issue #542.
This script adds an additional step in hapsequencer which attempts to locate the equivalent HLA classic filter product image and sourcelist for each HAP filter drizzled product image/source list pair, and then if the HLA classic files are found,  executes drizzlepac/devutils/comparison_tools/compare_sourcelists.py.

Highlights:
- Added subroutine _run_sourcelist_comparision_ to hapsequencer.py. This subroutine takes care of locating the relevant HLA classic image/sourcelist and the actual execution of compare_sourcelists.py
- Brought logging in compare_sourcelists.py and starmatch_hist.py up to HAP standards
- Sanitized all mentions of absolute paths in log messages in compare_sourcelists.py and starmatch_hist.py
- The filenames of all .pdf plot files generated by compare_sorucelists.py are now prepended with a text string containing filt_obj.product_basename and hap catalog type ("point" or "segment") information. This information, as well as the date/time of plot generation is also written to the lower left corner of every plot to eliminate any potential ambiguity concerning which catalog comparison the plot files go to.

Issue #490, opened 2019-11-11
Single visit processing needs to control num_cores parameter


stsci-hack:
The pipeline interface for the single-visit processing, runsinglehap, needs to allow the user or pipeline to control how many CPUs get used by astrodrizzle during processing for each process.  This is a feature provided to HSTDP processing for standard pipeline processing to generate the default drizzle products, and would allow them to optimize single-visit processing based on the available resources in the pipeline. 

The default value for all single-visit processing has been set to use only 1 CPU to minimize resource usage, especially for very large visits.  This would allow the pipeline to specify how many more CPUs to use to process a visit so that larger visits can be processed or reprocessed in a timely manner.

PR #954, opened 2021-02-26
Fix EXTNAME/EXTVER in simple FITS files

stsci-hack:
These changes insure that all simple FITS files written out by astrodrizzle no longer have the `EXTNAME` and `EXTVER` keywords in the simple FITS drizzle product headers.  Instead, the `FILETYPE` keyword has been added to document the type of data in the file (SCI, WHT, CTX).  

This avoids the problems reported in #590 .  

In addition, 
  - all Flake8 errors in the 'outputimage.py' module have been resolved.
  - leftover WCS distortion keywords are finally cleaned up properly

Fixes https://github.com/spacetelescope/drizzlepac/issues/862, #590

### DIFF
--- a/drizzlepac/outputimage.py
+++ b/drizzlepac/outputimage.py
@@ -7,7 +7,7 @@ This module manages the creation of the output image FITS file.
 
 """
 from astropy.io import fits
-from stsci.tools import fileutil, readgeis, logutil
+from stsci.tools import fileutil, logutil
 
 from . import wcs_functions
 from . import version
@@ -18,7 +18,7 @@ from fitsblender import blendheaders
 yes = True
 no = False
 
-RESERVED_KEYS = ['NAXIS', 'BITPIX', 'DATE', 'IRAF-TLM', 'XTENSION', 'EXTNAME',' EXTVER']
+RESERVED_KEYS = ['NAXIS', 'BITPIX', 'DATE', 'IRAF-TLM', 'XTENSION', 'EXTNAME', ' EXTVER']
 
 EXTLIST = ('SCI', 'WHT', 'CTX')
 
@@ -32,21 +32,21 @@ PYFITS_COMPRESSION = hasattr(fits, 'CompImageHDU')
 # Set up dictionary of default keywords to be written out to the header
 # of the output drizzle image using writeDrizKeywords()
 DRIZ_KEYWORDS = {
-                'VER':{'value':"",'comment':'Drizzle, task version'},
-                'GEOM':{'value':"wcs",'comment':'Drizzle, source of geometric information'},
-                'DATA':{'value':"",'comment':'Drizzle, input data image'},
-                'DEXP':{'value':"",'comment':'Drizzle, input image exposure time (s)'},
-                'OUDA':{'value':"",'comment':'Drizzle, output data image'},
-                'OUWE':{'value':"",'comment':'Drizzle, output weighting image'},
-                'OUCO':{'value':"",'comment':'Drizzle, output context image'},
-                'MASK':{'value':"",'comment':'Drizzle, input weighting image'},
-                'WTSC':{'value':"",'comment':'Drizzle, weighting factor for input image'},
-                'KERN':{'value':"",'comment':'Drizzle, form of weight distribution kernel'},
-                'PIXF':{'value':"1.0",'comment':'Drizzle, linear size of drop'},
-                'COEF':{'value':"SIP",'comment':'Drizzle, source of coefficients'},
-                'OUUN':{'value':"cps",'comment':'Drizzle, units of output image - counts or cps'},
-                'FVAL':{'value':"INDEF",'comment':'Drizzle, fill value for zero weight output pix'},
-                'WKEY':{'value':"",'comment':'Input image WCS Version used'}
+                'VER': {'value': "", 'comment': 'Drizzle, task version'},
+                'GEOM': {'value': "wcs", 'comment': 'Drizzle, source of geometric information'},
+                'DATA': {'value': "", 'comment': 'Drizzle, input data image'},
+                'DEXP': {'value': "", 'comment': 'Drizzle, input image exposure time (s)'},
+                'OUDA': {'value': "", 'comment': 'Drizzle, output data image'},
+                'OUWE': {'value': "", 'comment': 'Drizzle, output weighting image'},
+                'OUCO': {'value': "", 'comment': 'Drizzle, output context image'},
+                'MASK': {'value': "", 'comment': 'Drizzle, input weighting image'},
+                'WTSC': {'value': "", 'comment': 'Drizzle, weighting factor for input image'},
+                'KERN': {'value': "", 'comment': 'Drizzle, form of weight distribution kernel'},
+                'PIXF': {'value': "1.0", 'comment': 'Drizzle, linear size of drop'},
+                'COEF': {'value': "SIP", 'comment': 'Drizzle, source of coefficients'},
+                'OUUN': {'value': "cps", 'comment': 'Drizzle, units of output image - counts or cps'},
+                'FVAL': {'value': "INDEF", 'comment': 'Drizzle, fill value for zero weight output pix'},
+                'WKEY': {'value': "", 'comment': 'Input image WCS Version used'}
                 }
 
 log = logutil.create_logger(__name__, level=logutil.logging.NOTSET)
@@ -105,7 +105,7 @@ class OutputImage:
         self.blot = blot
 
         if PYFITS_COMPRESSION and 'compress' in input_pars:
-            self.compress = input_pars['compress'] # Control creation of compressed FITS files
+            self.compress = input_pars['compress']  # Control creation of compressed FITS files
         else:
             self.compress = False
 
@@ -115,10 +115,10 @@ class OutputImage:
 
         if not blot:
             self.output = plist[0]['output']
-            self.shape = (plist[0]['outny'],plist[0]['outnx'])
+            self.shape = (plist[0]['outny'], plist[0]['outnx'])
         else:
             self.output = plist[0]['blotImage']
-            self.shape = (plist[0]['blotny'],plist[0]['blotnx'])
+            self.shape = (plist[0]['blotny'], plist[0]['blotnx'])
 
         # Keep track of desired output WCS computed by PyDrizzle
         self.wcs = wcs
@@ -145,7 +145,7 @@ class OutputImage:
             # Report values appropriate for entire combined product
             self.texptime = plist[0]['texptime']
             self.expstart = plist[0]['texpstart']
-            self.expend = plist[_nimgs-1]['texpend']
+            self.expend = plist[_nimgs - 1]['texpend']
 
 
         if blot:
@@ -159,13 +159,13 @@ class OutputImage:
         self.outweight = _outweight
         self.outcontext = _outcontext
 
-    def set_bunit(self,bunit):
+    def set_bunit(self, bunit):
         """
         Method used to update the value of the bunit attribute.
         """
         self.bunit = bunit
 
-    def set_units(self,units):
+    def set_units(self, units):
         """
         Method used to record what units were specified by the user for the output product.
         """
@@ -234,19 +234,18 @@ class OutputImage:
         # If only writing out single drizzle product, blending needs to be
         # forced off as there is only 1 input to report, no blending needed
         if self.single:
-            blend=False
+            blend = False
 
         # If input data is not in MEF FITS format, it will return 'None'
         # and those headers will have to be generated from drizzle output
         # file FITS headers.
         # NOTE: These are HEADER objects, not HDUs
-        #prihdr,scihdr,errhdr,dqhdr = getTemplates(template)
         self.fullhdrs, intab = getTemplates(template, blend=False,
                                             rules_file=rules_file)
 
-        newhdrs, newtab = getTemplates(template,blend=blend,
+        newhdrs, newtab = getTemplates(template, blend=blend,
                                         rules_file=rules_file)
-        if newtab is not None: nextend += 1 # account for new table extn
+        if newtab is not None: nextend += 1  # account for new table extn
 
         prihdr = newhdrs[0]
         scihdr = newhdrs[1]
@@ -281,7 +280,7 @@ class OutputImage:
             prihdu.header['EXPSTART'] = self.expstart
             prihdu.header['EXPEND'] = self.expend
 
-        #Update ASN_MTYPE to reflect the fact that this is a product
+        # Update ASN_MTYPE to reflect the fact that this is a product
         # Currently hard-wired to always output 'PROD-DTH' as MTYPE
         prihdu.header['ASN_MTYP'] = 'PROD-DTH'
 
@@ -292,14 +291,14 @@ class OutputImage:
         if 'DITHCORR' in prihdu.header:
             prihdu.header['DITHCORR'] = 'COMPLETE'
 
-        prihdu.header['NDRIZIM'] =(len(self.parlist),
+        prihdu.header['NDRIZIM'] = (len(self.parlist),
                                    'Drizzle, No. images drizzled onto output')
 
         # Only a subset of these keywords makes sense for the new WCS based
         # transformations. They need to be reviewed to decide what to keep
         # and what to leave out.
         if not self.blot:
-            self.addDrizKeywords(prihdu.header,versions)
+            self.addDrizKeywords(prihdu.header, versions)
 
         if scihdr:
             try:
@@ -312,7 +311,7 @@ class OutputImage:
                 scihdr['NCOMBINE'] = self.parlist[0]['nimages']
 
             # If BUNIT keyword was found and reset, then
-            bunit_last_kw = self.find_kwupdate_location(scihdr,'bunit')
+            bunit_last_kw = self.find_kwupdate_location(scihdr, 'bunit')
             if self.bunit is not None:
                 comment_str = "Units of science product"
                 if self.bunit.lower()[:5] == 'count':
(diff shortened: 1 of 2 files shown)


mcara (changes requested):
This PR would result in creating a non-standard FITS file: `'EXNAME'` and `'EXTVER'` do not belong in the primary header, see https://archive.stsci.edu/fits/fits_standard/node40.html#SECTION00942600000000000000

> 5.4.2.6 Extension Keywords
> 
> These keywords are used to describe an  extension **and should not appear in the primary header**.

drizzlepac/outputimage.py
```diff
@@ -491,25 +494,30 @@ def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
...
-                    pre_wcs_kw = self.find_kwupdate_location(hdu.header,'CD1_1')
+                    pre_wcs_kw = self.find_kwupdate_location(hdu.header, 'CD1_1')
                     # Update WCS Keywords based on PyDrizzle product's value
                     # since 'drizzle' itself doesn't update that keyword.
-                    addWCSKeywords(self.wcs,hdu.header, blot=self.blot,
+                    addWCSKeywords(self.wcs, hdu.header, blot=self.blot,
                                    single=self.single, after=pre_wcs_kw)
 
+                # explicitly set EXTNAME, EXTVER to (sci,1)
+                hdu.header.set('EXTNAME', value='WHT', before='TELESCOP')
```
mdlpstsci, line 506: Comment should be updated.

drizzlepac/outputimage.py
```diff
@@ -530,32 +538,37 @@ def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
...
-                    pre_wcs_kw = self.find_kwupdate_location(hdu.header,'CD1_1')
+                    pre_wcs_kw = self.find_kwupdate_location(hdu.header, 'CD1_1')
                     # Update WCS Keywords based on PyDrizzle product's value
                     # since 'drizzle' itself doesn't update that keyword.
-                    addWCSKeywords(self.wcs,hdu.header, blot=self.blot,
+                    addWCSKeywords(self.wcs, hdu.header, blot=self.blot,
                                    single=self.single, after=pre_wcs_kw)
 
+                # explicitly set EXTNAME, EXTVER to (sci,1)
+                hdu.header.set('EXTNAME', value='CTX', before='TELESCOP')
```
mdlpstsci, line 554: Comment should be updated.

mdlpstsci (changes requested):
Just small change to two comments and this ticket is ready to go.

mcara:
> Just small change to two comments and this ticket is ready to go.

Is adding keywords that are supposed to appear in the extensions to the primary header OK?

mdlpstsci:
Actually, as strange as it may seem, I do not think this violates the standard - though it is unusual.  See
https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf

Section 4.4.2.6 and Appendix C: Summary of keywords (Table C.2).

mdlpstsci:
HI,
I added a comment to the PR.
“Actually, as strange as it may seem, I do not think this violates the standard - though it is unusual. See
https://fits.gsfc.nasa.gov/standard40/fits_standard40aa-le.pdf
Section 4.4.2.6 and Appendix C: Summary of keywords (Table C.2).”
Having said the above, I also consulted with a member of one of the FITS Technical panels and Dedicated Task Forces, Richard Shaw. He indicated that use of EXTNAME/EXTVER in the Primary HDU (or only HDU in the case of a simple FITS file) may not technically be a violation of the standard, but it should not be done as it would be confusing to users.
Best,
Michele

Michele D. De La Pena
mdelapena@stsci.edu<mailto:mdelapena@stsci.edu>
Science Calibration Software Branch
Space Telescope Science Institute
Mobile: 520.419.8825
Office: 667.218.6501

stsci-hack:
Given that the FITS standard states that neither EXTNAME or EXTVER can be in the Primary header, and to close all further discussions, this PR has been modified to instead remove those keywords and instead add the FILETYPE keyword which is not a required FITS keyword and is designed to be used in this manner.  Sorry about the confusion...

drizzlepac/outputimage.py
stsci-hack: Although this code has been replaced, the new keyword being added does include a comment as implemented in 89fa478.

drizzlepac/outputimage.py
stsci-hack: Although this code has been replaced, the new keyword being added does include a comment as implemented in 89fa478.

drizzlepac/outputimage.py
```diff
@@ -744,20 +766,25 @@ def deleteDistortionKeywords(hdr):
...
+    dist_kws = ['D2IMERR1', 'D2IMERR2', 'D2IMDIS1', 'D2IMDIS2',
+                'D2IMEXT', 'DP1', 'DP2', ]
+    dist_multi_kws = ['D2IM1.*', 'D2IM2.*', 'CPDIS?', 'CPERR?']
     for kw in dist_kws:
         if kw in hdr:
             del hdr[kw]
+    for multi_kw in dist_multi_kws:
+        for kw in hdr[multi_kw]:
+            if kw in hdr:
+                del hdr[kw]
```
mcara, line 778:
```suggestion
    for multi_kw in dist_multi_kws:
        del hdr[multi_kw]
```
This works just fine.

drizzlepac/outputimage.py
```diff
@@ -744,20 +766,25 @@ def deleteDistortionKeywords(hdr):
     """ Delete distortion related keywords from output drizzle science header
         since the drizzled image should have no remaining distortion.
     """
-    dist_kws = ['D2IMERR1','D2IMERR2','D2IMDIS1','D2IMDIS2','D2IM1.*','D2IM2.*',
-                'D2IMEXT', 'DP1', 'DP2', 'CPDIS?', 'CPERR?']
+    dist_kws = ['D2IMERR1', 'D2IMERR2', 'D2IMDIS1', 'D2IMDIS2',
+                'D2IMEXT', 'DP1', 'DP2', ]
+    dist_multi_kws = ['D2IM1.*', 'D2IM2.*', 'CPDIS?', 'CPERR?']
```
mcara, line 771: Would it make sense to make this a global constant, especially if `addWCSKeywords()` is called often?

drizzlepac/outputimage.py
```diff
@@ -446,31 +446,38 @@ def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
 
             # explicitly set EXTEND to FALSE for simple FITS files.
             dim = len(sciarr.shape)
-            hdu.header.set('extend', value=False, after='NAXIS%s'%dim)
+            hdu.header.set('extend', value=False, after='NAXIS%s' % dim)
+
+            # explicitly remove EXTNAME, EXTVER from header - they are not allowed
+            for kw in RESERVED_KEYS[-2:]:
+                if kw in hdu.header:
+                    del hdu.header[kw]
```
mcara, line 454: I do not think this is necessary at all because `'EXTNAME'` and `'EXTVER'` should not be present in `hdu` for the following reasons:
1. On line `#423` `hdu_header` is defined as a copy of a primary HDU header (`hdu_header = prihdu.header.copy()`) so it cannot (or should not) have `'EXTNAME'` and `'EXTVER'` in it.
2. On lines `#441-444`, `hdu` is created as, i.e., `hdu = fits.ImageHDU(data=sciarr, header=hdu_header)` without `name` or `ver` specified and hence, even though it inherits from `hdu_header`, it also will not contain `'EXTNAME'` and `'EXTVER'`.

So, this code seems unnecessary. But if you do want to keep it, I would suggest the following modification:

```suggestion
            for kw in ['EXTNAME', 'EXTVER']:
                hdu.header.pop(kw, None)
```
Explicitly specifying keywords to be deleted is safer if anyone later would like to update `RESERVED_KEYS` constant and add other words after the last one in the list.

mcara: Moreover, line `#444` (`hdu = fits.ImageHDU(data=sciarr, header=hdu_header)`) could be replaced directly with 

```python
hdu = fits.PrimaryHDU(data=sciarr, header=hdu_header)
```
which I highly recommend.

drizzlepac/outputimage.py
```diff
@@ -491,25 +498,33 @@ def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
...
                     # Update WCS Keywords based on PyDrizzle product's value
                     # since 'drizzle' itself doesn't update that keyword.
-                    addWCSKeywords(self.wcs,hdu.header, blot=self.blot,
+                    addWCSKeywords(self.wcs, hdu.header, blot=self.blot,
                                    single=self.single, after=pre_wcs_kw)
 
+                # explicitly remove EXTNAME, EXTVER from header - they are not allowed
+                for kw in RESERVED_KEYS[-2:]:
+                    if kw in hdu.header:
+                        del hdu.header[kw]
```
mcara, line 512: Same reasoning as for `sciarr` above, this is not necessary: code on lines `#490-493` insures that `hdu` is not going to contain `'EXTNAME'` and `'EXTVER'`. Also, replace line `#493` with:
```python
hdu = fits.PrimaryHDU(data=whtarr, header=prihdu.header)
```

drizzlepac/outputimage.py
```diff
@@ -530,32 +545,40 @@ def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
...
```
mcara, line 563: Same reasoning as for `sciarr` above, this is not necessary: code on lines `#541-543` insures that `hdu` is not going to contain `'EXTNAME'` and `'EXTVER'`. Also, replace line `#543` with:
```python
hdu = fits.PrimaryHDU(data= _ctxarr, header=prihdu.header)
```

drizzlepac/outputimage.py
```diff
@@ -530,32 +545,40 @@ def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
                 # header to Primary header...
                 if dqhdr:
                     for _card in dqhdr.cards:
-                        if ( (_card.keyword not in RESERVED_KEYS) and
+                        if ((_card.keyword not in RESERVED_KEYS) and
```
mcara, line 548: This adds to the argument that `hdu` cannot contain `'EXTNAME'` and `'EXTVER'` because they are not copied over.

mcara (changes requested):
Adding `FILETYPE` is fine. The code that explicitly deletes `'EXTNAME'` and `'EXTVER'` is unnecessary unless one can convince me otherwise.

Also, when creating a "simple FITS" file (here this means "no extensions"), I would suggest using directly `PrimaryHDU` instead of `ImageHDU` on lines `#444, 485, 528` (current `master` branch).

mcara:
Also, while at fixing thing, please delete this completely useless (in the sense that it is not doing anything other than slow down execution) code:
https://github.com/spacetelescope/drizzlepac/blob/7bf45d6d4fd1296e32b4cb1a195341c0156571a6/drizzlepac/outputimage.py#L433-L437

mcara:
Looking at git blame (see https://github.com/spacetelescope/drizzlepac/commit/8013e79929b5ee348564e2754237246693efcacf), cc: @nden , it seems that the code is necessary but it was a bug:
```python
del hdu.header['PCOUNT']
del hdu.header['GCOUNT']
```
should be replaced with simply:
```python
hdu.header.pop('PCOUNT', None)
hdu.header.pop('GCOUNT', None)
```

mcara:
Also, while at fixing thing, please replace this code:
https://github.com/spacetelescope/drizzlepac/blob/7bf45d6d4fd1296e32b4cb1a195341c0156571a6/drizzlepac/outputimage.py#L433-L437 that does not do anything with the fix below.

Looking at git blame (see https://github.com/spacetelescope/drizzlepac/commit/8013e79929b5ee348564e2754237246693efcacf), cc: @nden , it seems that the code is necessary but it was a bug:

**Original/Old Code:**
```python
del hdu.header['PCOUNT']
del hdu.header['GCOUNT']
```
should be replaced with simply this
**Fix:**
```python
hdu.header.pop('PCOUNT', None)
hdu.header.pop('GCOUNT', None)
```
(which also may not be necessary if, when appropriate, `PrimaryHDU` is used instead of the `ImageHDU`).

mcara:
I missed the discussion about FITS standard. Indeed, in the version 4.0 (see PDF file linked by @mdlpstsci, section H.2.3) `EXTEVER` and `EXTNAME` are now allowed in primary headers compared to version 3.0 were they were not allowed. A part of confusion is also due to both standards being listed, see, i.e., https://heasarc.gsfc.nasa.gov/docs/fcg/standard_dict.html which says they are not allowed (likely standard version 3.0).

I agree that it would be better not to put them in the primary header, especially if adding `FILETYPE` (which seems to be used in other parts of the `drizzlepac` code anyway) solves the issue.

mcara:
I do believe that since this a bug fix ("avoids the problems reported in #590" and deleting all distortions) to a released code, this PR should add a change log entry.

mcara:
@stsci-hack So, does this fix https://github.com/spacetelescope/drizzlepac/issues/862 _completely_? Shouldn't `TDD_*` be removed as well?

drizzlepac/outputimage.py
stsci-hack: Unfortunately, if the keyword does not exist, the line 'del hdr[multi_kw]' will actually cause an Exception to be thrown, which is what my version explicitly avoids.

drizzlepac/outputimage.py
stsci-hack: This does make sense.  I will see about making these global constants...

drizzlepac/outputimage.py
mcara: Have you tried this? Here is my test:
```python
In [2]: from astropy.io import fits
   ...: import numpy as np
   ...: fo = fits.HDUList()
   ...: hdr = fits.Header()
   ...: hdr['TELESCOP'] = 'HST'

In [3]: hdr
Out[3]: TELESCOP= 'HST     '

In [4]: del hdr['test*']

In [5]:
```

stsci-hack: Actually, this is what I get:

```
In [409]: hdu = fits.open('test_sregion_drc.fits')

In [410]: scihdr = hdu[1].header.copy()
In [414]: 'sregion' in scihdr
Out[414]: False

In [415]: del scihdr['sregion']
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-415-64a3a51e610a> in <module>
----> 1 del scihdr['sregion']

c:\users\warren\miniconda3\envs\rthstdev\lib\site-packages\astropy\io\fits\header.py in __delitem__(self, key)
    228                 # To delete keyword without caring if they were present,
    229                 # Header.remove(Keyword) can be used with optional argument ignore_missing as True
--> 230                 raise KeyError(f"Keyword '{key}' not found.")
    231
    232             for idx in reversed(indices[key]):

KeyError: "Keyword 'SREGION' not found."

In [416]: del scihdr['sregion*']

```
Your test example worked because 'del' has different rules for deleting multiple keywords at once as opposed to deleting individual keywords.  This forces the use of '.pop' or checking for the existence of the individual specific keywords before trying to work with them.

drizzlepac/outputimage.py
stsci-hack: Included as part of 4b565a7

drizzlepac/outputimage.py
stsci-hack: Leaving this in only insures against any unexpected problems with the inputs, and does not hurt the final code or take any measurable time to process.  So, this line was not modified in commit 4b565a7.

drizzlepac/outputimage.py
stsci-hack: Change was made in 4b565a7.

drizzlepac/outputimage.py
stsci-hack: All these concerns were addressed in 4b565a7, including using PrimaryHDU instead.  Unfortunately, this code needs to remain in place (as noted in the new comments) since the HDU that gets created MAY be a CompImageHDU instead of a PrimaryHDU depending on whether 'driz_sep_compress' is turned on or not.

drizzlepac/outputimage.py
stsci-hack: Done in 4b565a7

drizzlepac/outputimage.py
stsci-hack: This has been cleaned up in 4b565a7 by using 'del' and '.pop' appropriately avoiding the need to check for the existence of the keywords ahead of time.

stsci-hack:
This can now be said to completely address #862 as well with the addition of the 'TDD*' keywords to the list of distortion keywords removed from the output image headers (implemented in 4b565a7).   The change log has also been updated now that we know the full extent of the changes in this PR.

drizzlepac/outputimage.py
```diff
@@ -18,35 +18,41 @@
 yes = True
 no = False
 
-RESERVED_KEYS = ['NAXIS', 'BITPIX', 'DATE', 'IRAF-TLM', 'XTENSION', 'EXTNAME',' EXTVER']
+RESERVED_KEYS = ['NAXIS', 'NAXIS1', 'NAXIS2', 'EXTEND',
+                 'BITPIX', 'DATE', 'IRAF-TLM',
+                 'XTENSION', 'EXTNAME', ' EXTVER']
```
mcara, line 23:
```suggestion
                 'XTENSION', 'EXTNAME', 'EXTVER']
```

mcara (dismissed):
LGTM.

mcara:
Could you, please fix the extra white space at the beginning of `' EXTVER'` and maybe squash all these commits? Thanks you!

drizzlepac/outputimage.py
stsci-hack: Fixed in 6a252d0.

PR #956, opened 2021-03-03
Improved the way WCS solutions are manipulated for visits with Grism/Prism data

mdlpstsci:
In particular, when the active WCS solution is being replaced, the current active solution (and any alternative WCS solutions from the same distortion model) are saved as headerlet extensions (if they are not duplicates according to their
HDRNAME).  Duplicates are deleted.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -560,17 +560,18 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
 
         # It is possible the total_obj_list output from the poller_utils contains only Grism/Prism
         # data and no direct images, so no further processing should be done.  If this is the case,
-        # there is actually nothing to be done for the visit, except write out a manifest and 
+        # there is actually nothing to be done for the visit, except write out a manifest and
         # a log file.  Check every item in the total data product list.
         found_data = False
         for total_item in total_obj_list:
             if total_item.edp_list and not found_data:
-               found_data = True
+                found_data = True
             elif total_item.grism_edp_list:
-               no_data_trl = total_item.trl_filename
+                no_data_trl = total_item.trl_filename
         if not found_data:
             log.warning("")
-            log.warning("There are no viable direct images in any Total Data Product. No processing can be done.")
+            log.warning("There are no viable direct images in any Total Data Product for this visit. No processing can be done.")
+            log.warning("No SVM processing is done for the Grism/Prism data - no SVM output products are generated.") 
             product_list += [no_data_trl]
             sys.exit(0)
 
@@ -609,7 +610,7 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
             _ = delete_ramp_exposures(total_item.fdp_list, 'FILTER')
             ramp_list = delete_ramp_exposures(total_item.edp_list, 'EXPOSURE')
             product_list += ramp_list
-  
+
             # If there are Grism/Prism images present in this visit, as well as corresponding direct images
             # for the same detector, update the primary WCS in the direct and/or Grism/Prism images as
             # appropriate to be an 'a priori' or the pipeline default (fallback) solution.  Note: there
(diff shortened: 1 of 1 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -1122,49 +1146,76 @@ def update_active_wcs(filename, wcsname, keyword_wcs_names_dict, image_dict):
...
+        current_active_hdrname = hdu['SCI', 1].header['HDRNAME']
+        current_active_wcsname = hdu['SCI', 1].header['WCSNAME']
+        wcsutil.headerlet.archive_as_headerlet(filename, current_active_hdrname, wcskey='PRIMARY')
+        log.info("Archiving previous active WCS solution as necessary: {}".format(current_active_wcsname))
+
+        # Get the distortion model identification of the desired active WCS solution
+        index = wcsname.upper().find('IDC_')
+        idc_new_string = ''
+        if index > 0:
+            idc_new_string = wcsname[0:13]
```
stsci-hack, line 1178: A more reliable, non-hardcoded, way to get this string would be to split on '-'.  Something like this might help:

```
wcsname = wcsname.split('-')[0]
index = wcsname.upper.find('IDC_')
idc_new_string = ''
if index > 0:
    idc_new_string = wcsname[index:]
```
If there is no '-' in the WCSNAME, this will still work as desired.

drizzlepac/hapsequencer.py
```diff
@@ -1122,49 +1146,76 @@ def update_active_wcs(filename, wcsname, keyword_wcs_names_dict, image_dict):
...
+            index = wname.upper().find(idc_new_string.upper())
+
+            # No match so solution will be copied to a headerlet automatically when the new primary is set
+            if index == -1 and wkey.upper() != 'O':
+                log.info("Archiving alternate WCS solution as a headerlet as necessary: {}".format(wname))
+
+                # Now check if the HDRNAME between this solution and a headerlet already exists
+                # Get the headerlet HDRNAMES fresh every time
+                hdr_keyword = hdu[1].header['HDRNAME' + wkey.upper()]
+                headerlet_hdrnames = wcsutil.headerlet.get_headerlet_kw_names(filename, kw="HDRNAME")
```
stsci-hack, line 1193: The set of headerlet_hdrnames could be determined prior to this loop and, I think, eliminate the need to do this again in line 1204, if I understand what is going on correctly...

drizzlepac/hapsequencer.py
```diff
@@ -1122,49 +1146,76 @@ def update_active_wcs(filename, wcsname, keyword_wcs_names_dict, image_dict):
...
+
+                # Solution already exists as a headerlet extension, so just delete it
+                if hdr_keyword in headerlet_names:
+                    wcsutil.altwcs.deleteWCS(filename, extname_list, wcskey=wkey)
```
stsci-hack, line 1197: This line will delete the alternate WCS from the SCI header, but it is not clear to me what code actually saves any alternate WCS as a headerlet.

stsci-hack (changes requested):
I had a couple of questions and suggestions that should be resolved before this gets merged.

drizzlepac/hapsequencer.py
mdlpstsci: Modified in the suggested manner.

drizzlepac/hapsequencer.py
mdlpstsci: This portion of the code is just looking at alternate WCS solutions which have a different distortion model than that of the soon to be installed "new" active solution.  If this particular alternate WCS solution already exists as a headerlet, then it should be deleted so it does not create a duplicate headerlet extension.  In testing I found the alternate WCS solutions are saved automatically as headerlet extensions when the solutions have conflicting distortion models.  However, I will double-check this is true for **_BOTH_** cases: when the new active solution's origin is an alternate WCS or a headerlet extension.

stsci-hack: You are right that they should already be saved as headerlet extensions.  I just did not remember that detail when looking at just this code by itself.  It may be helpful to just add a comment here to make that more obvious for anyone coming into this code later.

drizzlepac/hapsequencer.py
mdlpstsci: True. This is cruft.

drizzlepac/hapsequencer.py
mdlpstsci: Performed more testing and made more improvements.

stsci-hack (approved):
LGTM.

PR #958, opened 2021-03-09
Ensure diagnostic files do not get created during standard processing

mdlpstsci:
Surrounded the creation of some output files used for diagnosticpurposes only with "if self.diagnostic_mode" so they do not get created during standard processing.


stsci-hack (approved):
LGTM

PR #962, opened 2021-03-12
New SREGION computation based on WCS footprint

stsci-hack:
The SREGION computation from #951 was susceptible to a number of error conditions.  These changes completely replace that algorithm with one based on wcs.calc_footprint() (and related edge pixel) positions to avoid those previous error conditions.  This new version not only improved on the vertices identified in the previous PR, but uses an algorithm which is not susceptible to as many error conditions.  

The vertices are based on edges defined by the corners (and mid-points of each outer edge) of each input chip and where they intersect the total footprint.   

This code was tested against all the datasets listed in #951 as well as data from visit ia0m04 which triggered the exceptions in testing on 1082 datasets.

### DIFF
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -4,6 +4,7 @@ from itertools import chain, combinations
 
 from matplotlib import path
 from matplotlib import pyplot as plt
+from matplotlib.path import Path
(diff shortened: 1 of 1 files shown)


mdlpstsci (approved):
More comments are always grand for those who might need help following the algorithm.  Having said this, the code looks fine.

PR #963, opened 2021-03-12
Debugging issues with archiving and restoring WCS solutions from headerlets

mdlpstsci:
Code currently appears to be working fine now with traps for exceptions.  I just need to understand why some of the exceptions are being raised in the first place.  Examining the headerlet code in stwcs.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -74,8 +74,6 @@ from stwcs import updatewcs
 from stwcs import wcsutil
 from stwcs.wcsutil import headerlet
 
-import pdb
-
 __taskname__ = 'hapsequencer'
 MSG_DATEFMT = '%Y%j%H%M%S'
 SPLUNK_MSG_FORMAT = '%(asctime)s %(levelname)s src=%(name)s- %(message)s'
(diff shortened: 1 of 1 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -74,6 +74,8 @@
 from stwcs import wcsutil
 from stwcs.wcsutil import headerlet
 
+import pdb
```
stsci-hack, line 77: This should not remain in operational code.

drizzlepac/hapsequencer.py
```diff
@@ -1035,6 +1037,8 @@ def update_wcs_in_visit(tdp):
         log.info("The active WCS solution for each Grism/Prism and direct image is not changed.")
 
     log.info("Grism product list: {}".format(grism_product_list))
+    # MDD
+    sys.exit(0)
```
stsci-hack, line 1041: Did you mean to exit before returning the list?

drizzlepac/hapsequencer.py
mdlpstsci: Sorry, code was still "In progress".  This line of code is deleted.

drizzlepac/hapsequencer.py
mdlpstsci: Sorry, code was still "In Progress" and being debugged.  The lines 1040/1041 have been removed.

stsci-hack (approved):
LGTM

PR #964, opened 2021-03-15
Make MVM test datasets

dulude:
**Relevant JIRA Ticket**
- [HLA-488](https://jira.stsci.edu/browse/HLA-488)

**Summary**
- Created search tool /devutils/search_skyframes.py that allows user to query list of HST observations based on some combination of skycell name, instrument/detector name, filter name, exposure name and/or observation date range.
- Created script /haputils/make_poller_files.py that generates SVM or MVM poller file if given a list of image rootnames, and skycell name (for MVM)
- Using these tools, Found that the skycell that contains ACS and WFC3 observations of NGC 45 is an ideal MVM test dataset. The skycell (p0797x13y05) contains both overlapping and non-overlapping groups of images. In addition, the 37 images are a mix of ACS/WFC and WFC3/UVIS observations that span 11 years. The oldest image was taken early June 2004, and the youngest image was taken mid-July, 2015. In addition, at only 37 images,


drizzlepac/devutils/search_skyframes.py
```diff
@@ -0,0 +1,307 @@
...
+
+def make_search_string(arg_dict):
+    """Create search string that will be used to query observation tables
+
+    Parameters
+    ----------
+    arg_dict: dict
+        dictionary with the search arguments. They are as follows: 'exposure': Image name,
+        'skycell': Skycell name, 'config': Instrument/Detector name (Uppercase, separated by a "/"
+        (e.g. WFC3/IR)), 'spec': Filter name. lowercase.
```
stsci-hack, line 63: It may be helpful for future users to know how to provide multi-filter use-cases, since ACS can use 2 filters at once.

dulude: issue resolved by commit [04379d8].(https://github.com/spacetelescope/drizzlepac/pull/964/commits/04379d8ac5eefe2d50335d3da7cc9c73824050b1)

drizzlepac/devutils/search_skyframes.py
```diff
@@ -0,0 +1,316 @@
...
+        add_an_s = "s"
+    else:
+        add_an_s = ""
+    print("Building footprint {} image{}. Please be patient. This may take some time...".format(n_unique_skycells, add_an_s))
+    for skycell_name in unique_skycell_list:
+        skycell = cell_utils.SkyCell.from_name("skycell-{}".format(skycell_name))
+        footprint = cell_utils.SkyFootprint(meta_wcs=skycell.wcs)
+        img_list = results.query('skycell == "{}"'.format(skycell_name)).filename.values.tolist()
+        footprint.build(img_list)
+        footprint_imgname = "skycell-{}-footprint.fits".format(skycell_name)
```
stsci-hack, line 189: It appears that footprints of different instrument/detector combinations for the same SkyCell, let alone filter combinations, would end up overwriting each other with a name that only includes the SkyCell ID.  If the `skycell_name` includes these other aspects as part of the name, that would be fine, but some comments would go a long way towards making that clear.  Otherwise, this name will need to be expanded based on the search criteria.

drizzlepac/devutils/search_skyframes.py
```diff
@@ -0,0 +1,316 @@
+#!/usr/bin/env python
+
+""" This script allows the user to query a master observations .csv file by any combination of exposure name,
+    skycell name, instrument/detector, and/or filter name. The results are printed the screen and optionally
+    written to a user-specified output .csv file. The results will be sorted first by skycell name, then by
+    instrument/detector, filter name and finally by image name."""
```
stsci-hack, line 6: Please include at least one simple example of how to use this code in the module level docstring here.   The example should start with how to point to the master file, how to define a search, how to perform the search, and how to visualize the results.  A clear as the code is to follow, it hides a lot of details in the calls it uses.  This will also help anyone trying to either maintain or just simply use this code in the future to know how this code was designed and intended to be used.

drizzlepac/haputils/make_poller_files.py
```diff
@@ -0,0 +1,154 @@
...
+    ----------
+    rootname : str
+        rootname to process
+
+    Returns
+    -------
+    fullfilepath : str
+        full path + image name of specified rootname.
+    """
+    if in_args.master_observations_file is "":
```
stsci-hack, line 109: Something is not right here, as `in_args.master_observations_file` gets defined in `search_skyframes.py` not this file, so this check will never succeed.  Perhaps, just checking that 'DATA_PATH' is defined and exists is what was intended.

drizzlepac/devutils/search_skyframes.py
```diff
@@ -0,0 +1,323 @@
...
+        # flip order of dates if first is later than the second.
+        if date_range_dt[1] < date_range_dt[0]:
+            arg_dict["date_range"].reverse()
+
+    # Exit if in_args.master_observations_file == in_args.output_results_file so the master observations table doesn't get overwritten
+    if in_args.master_observations_file == in_args.output_results_file:
+        sys.exit("ERROR: The output results file cannot have the same name is the input master observations table")
+
+    # Exit if in_args.master_observations_file is not found.
+    if in_args.master_observations_file is "":
```
stsci-hack, line 290: It may also be more direct, and less prone to misinterpretation, to simply check for the existence of `ALL_EXP_FILE` in `os.environ` and check that it exists.

drizzlepac/haputils/make_poller_files.py
dulude: Issue resolved by commit [d584864](https://github.com/spacetelescope/drizzlepac/pull/964/commits/d584864cbce664123ea42889659c6cf656bfba3a)

drizzlepac/devutils/search_skyframes.py
dulude: Issue resolved by commit [44205e0](https://github.com/spacetelescope/drizzlepac/pull/964/commits/44205e0991622893e87586fea26e1381a5ee9d24)

drizzlepac/devutils/search_skyframes.py
dulude: Issue resolved by commit [44205e0](https://github.com/spacetelescope/drizzlepac/pull/964/commits/44205e0991622893e87586fea26e1381a5ee9d24)

drizzlepac/devutils/search_skyframes.py
dulude: issue resolved by commit [d96c85b5](https://github.com/spacetelescope/drizzlepac/pull/964/commits/d96c85b53ac5931851995bfcfd89ede8e8e51e90)

stsci-hack (approved):
The documentation really makes this a lot easier to understand and use!

PR #965, opened 2021-03-15
Add more error cases for FWHM use

stsci-hack:
The use of the FWHM determination by the library PSF use for point-catalog generation now includes proper catches for invalid values returned by photutils.  In addition, other error cases where the code did not correctly catch empty lists for point-catalogs based on the library PSF were also addressed by the changes in the PR.  

The additional check for CR map used during alignment also needed to be added for cases where no updates are returned when alignment sources are identified.  Should be tested against dataset j8dnd5.

These changes were tested against datasets ib1319, ibbso1, ia1s70 and jcol11.

### DIFF
--- a/drizzlepac/haputils/deconvolve_utils.py
+++ b/drizzlepac/haputils/deconvolve_utils.py
@@ -926,7 +926,7 @@ def find_point_sources(drzname, data=None, mask=None,
     # try to measure just the core of the PSF
     # This will be a lot less likely to result in invalid/impossible FWHM values
     yc, xc = np.where(drzpsf == drzpsf.max())[0]
-    psf_core = drzpsf[yc - box_size: yc + box_size, xc - box_size: yc - box_size]
+    psf_core = drzpsf[yc - box_size: yc + box_size, xc - box_size: xc + box_size]
     psf_fwhm = amutils.find_fwhm(psf_core, def_fwhm)
 
     # check value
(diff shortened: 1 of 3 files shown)


mdlpstsci (dismissed):
Changes are well contained and look fine.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -915,10 +920,25 @@ def find_point_sources(drzname, data=None, mask=None,
...
-                                     pixfrac=0.8,
+                                     pixfrac=1.5,
                                      clean_psfs=clean_psfs)
     drzpsf = fits.getdata(drzpsfname)
-    psf_fwhm = amutils.find_fwhm(drzpsf, 2.0)
+    # try to measure just the core of the PSF
+    # This will be a lot less likely to result in invalid/impossible FWHM values
+    yc, xc = np.where(drzpsf == drzpsf.max())[0]
+    psf_core = drzpsf[yc - box_size: yc + box_size, xc - box_size: yc - box_size]
+    psf_fwhm = amutils.find_fwhm(psf_core, def_fwhm)
```
mdlpstsci, line 930: I just want to make sure line 929 is correct in terms of the limits of the box being extracted (second dimension).

mdlpstsci (commented):
Just need to confirm one item noted below.  If it is as it should be, I will approve.

drizzlepac/haputils/deconvolve_utils.py
stsci-hack: Good catch!  Fat fingers and all that got the better of me there.  Fixed in 8030dc61

mdlpstsci (dismissed):

PR #966, opened 2021-03-16
Insure HDRNAME is always valid

stsci-hack:
A new function has been added to `hapsequencer` to insure that all SCI extensions of all science images have valid, non-empty HDRNAME keyword values.  Invalid values can cause problems with using the headerlet functions in STWCS.  

A couple of small Flake8 fixes were also added.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -69,10 +69,10 @@ from drizzlepac.haputils import processing_utils as proc_utils
 from drizzlepac.haputils import svm_quality_analysis as svm_qa
 from drizzlepac.haputils.catalog_utils import HAPCatalogs
 
+from stsci.tools.fileutil import countExtn
 from stsci.tools import logutil
 from stwcs import updatewcs
 from stwcs import wcsutil
-from stwcs.wcsutil import headerlet
 
 __taskname__ = 'hapsequencer'
 MSG_DATEFMT = '%Y%j%H%M%S'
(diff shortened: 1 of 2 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -1274,3 +1277,33 @@ def delete_ramp_exposures(obj_list, type_of_list):
...
+    fhdu, closefits = proc_utils._process_input(filename)
+
+    # Find all extensions to be updated
+    numext = countExtn(fhdu, extname='SCI')
+
+    for ext in range(1, numext + 1):
+        sciext = ('sci', ext)
+        scihdr = fhdu[sciext].header
+        if 'hdrname' not in scihdr or scihdr['hdrname'] == '':
+            # We need to create a valid value for the keyword
```
mdlpstsci, line 1298: This is going to seem paranoid, but should we make sure it is not a blank string too?

stsci-hack: The logic in the 'if' statement looks to see whether the current value is blank (''), and if so, initiates replacing that value.  In addition, the new value being generated will never be blank since it is being generated from the filename itself and the WCSNAME keyword value (which is never going to be blank at this point in the processing).  So, I believe this code adequately protects against blank strings in both the input and output values of this keyword.

mdlpstsci: Sorry, I am being dense.  I meant contains a blank versus being empty (' ' vs '').

stsci-hack: You are not the only one being dense.  That actually makes sense, and would be a simple change.  It will be included in the next commit.

stsci-hack: Fixed in 16785770

mdlpstsci (approved):
Looks good!

PR #969, opened 2021-03-17
Ensure the PHOT keywords can be found in the SVM product header

mdlpstsci:
When the update to the drizzle products for CAOM compatibility fails, the FITS headers of the files may not be modified to contain all the proper information in the proper header for successful processing.  In particular the PHOT keywords for the WFC3/IR are found in the Primary and must be copied to the SCI extensions which is where these keywords are found for all of the other instruments.  

Added a stopgap routine which will search the Primary header, as well as one other specified header (extname, extver) to search for a floating-point keyword.  This is only a stopgap as generic template should be created to handle multiple datatypes.

### DIFF
--- a/drizzlepac/haputils/processing_utils.py
+++ b/drizzlepac/haputils/processing_utils.py
@@ -346,3 +346,54 @@ def make_section_str(str="", width=60, symbol='-'):
     side_dash = symbol * (dash_len // 2)
     section_str = "{}{}{}".format(side_dash, str, side_dash)
     return section_str
+
+def find_flt_keyword (hdu, keyword, extname="SCI", extver=1):
+    """Look for the FITS keyword in the Primary and specified extension header
+
+    This routine deliberately does not use astropy.io.fits getval() as it
+    should not be used in application code due to its self-documented inefficiency.
+
+    Parameters
+    ----------
+    hdu : HDUList
+    List of Header Data Unit objects
+
+    keyword : str
+    Keyword of interest
+
+    extname : str
+    Named extension to check in case the keyword is not found in the Primary
+    e.g., (extname, extver) ("SCI", 2)
+
+    extver : int
+    Specified extver of an extname to check in case the keyword is not found in the Primary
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/processing_utils.py
```diff
@@ -346,3 +346,54 @@ def make_section_str(str="", width=60, symbol='-'):
     side_dash = symbol * (dash_len // 2)
     section_str = "{}{}{}".format(side_dash, str, side_dash)
     return section_str
+
+def find_flt_keyword (hdu, keyword, extname="SCI", extver=1):
+    """Look for the FITS keyword in the Primary and specified extension header
+
+    This routine deliberately does not use astropy.io.fits getval() as it
+    should not be used in application code due to its self-documented inefficiency.
+
```
stsci-hack, line 355: This docstring should make it clear that this function is only designed to work with keywords that return floating point values, as opposed to 'PHOTMODE' which would return a string.

drizzlepac/haputils/processing_utils.py
```diff
@@ -346,3 +346,54 @@ def make_section_str(str="", width=60, symbol='-'):
...
+    Specified extver of an extname to check in case the keyword is not found in the Primary
+    e.g., (extname, extver) ("SCI", 2)
+
+    Returns
+    -------
+    value : float
+    """
+
+    # Look for the keyword in the Primary header
+    try:
```
stsci-hack, line 378: Instead of an expensive `try...except` block, it may be easier to simply check for existence of the keyword in the headers before trying to get the value using 'if keyword in hdu[0].header`.  Only a suggestion as the current implementation would work just fine.

stsci-hack (approved):
LGTM

drizzlepac/haputils/processing_utils.py
mdlpstsci: Since this will be rewritten as a function more in line with what we will design, I leaving this function alone for now.

stsci-hack (approved):
Good enough for me.

PR #972, opened 2021-03-18
Correct computation of CR threshold factor

stsci-hack:
Computation of the CR threshold used to decide whether the catalogs are valid based on expected numbers of cosmic-rays turned out to have been provided the incorrect exposure time of N=1 exposures.  These changes correct the logic for computing the correct exposure time to be used in the threshold.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -252,11 +252,22 @@ def create_catalog_products(total_obj_list, log_level, diagnostic_mode=False, ph
                     # the catalogs.
                     tot_exposure_time = 0
                     n1_factor = 0.0
+                    n1_dict = {}
+                    for edp in total_product_obj.edp_list:
+                        if edp.filters not in n1_dict:
+                            n1_dict[edp.filters] = {'n': 1, 'texptime': edp.exptime}
+                        else:
+                            n1_dict[edp.filters]['n'] += 1
+                            n1_dict[edp.filters]['texptime'] += edp.exptime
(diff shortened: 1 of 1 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -341,7 +352,7 @@ def create_catalog_products(total_obj_list, log_level, diagnostic_mode=False, ph
             # rate of cosmic-ray contamination for the total detection product
             reject_catalogs = total_product_catalogs.verify_crthresh(n1_exposure_time)
 
-            if not reject_catalogs:
+            if not reject_catalogs or diagnostic_mode:
```
mdlpstsci, line 355: There will need to be a change in catalog_utils.py (write_catalog).  The routines needs to be notified about the reject_catalogs as the routines, with this change, will always write the .ecsv file and only write the .reg file in diagnostic mode.  There will be times when it is only desired to write the .reg file.  I think this should be a different ticket.

mdlpstsci (approved):
The changes look fine.  The issue of the writing the catalogs (.ecsv and/or .reg) should be addressed in a different ticket.

drizzlepac/hapsequencer.py
stsci-hack: Removed in 88a1980 so that this can be addressed more fully in a separate PR.

mdlpstsci (approved):
The clean up of the catalogs (.ecsv and .reg) is covered in HLA-505.

PR #974, opened 2021-03-18
Fixed the logic with regard to "reject_catalogs" and "diagnostic mode".

mdlpstsci:
If "diagnostic mode" is True, all diagnostic files (particularly the *.reg)
are written regardless of the setting of "reject_catalogs.  If
"reject_catalogs" is True, the output catalog files (*.ecsv) are not
written.


stsci-hack (approved):
I like this logic.

PR #983, opened 2021-03-23
Reset a column to its original name "RA" from the temporary name of "#RA"

mdlpstsci:
Ensure a column is reset to its original name ("RA") from the temporary name ("#RA") needed for use with ds9 (only when SVM_QUALITY_TESTING is set and logging is "debug" mode).


stsci-hack (approved):
Makes sense to me.

stsci-hack:
Photutils must be limited to <1.0.2 for this PR.  See #985 on change to 'setup.py' to get this to work.

stsci-hack (approved):
LGTM.

PR #988, opened 2021-03-24
Fix all photutils imports

larrybradley:
I expanded this PR beyond fixing only the `NoDetectionsWarning` import to all `photutils` imports and references in docstrings.  These changes work for all drizzlepac-supported versions (>=0.7) of `photutils`.  I will put in a separate PR of updates specifically for `photutils 1.1.0` enhancements.

I noticed that several private functions from the `photutils.detection` subpackage are being imported here. The private functions are not intended to be used outside of `photutils` and of course are subject to change (e.g., API changes) or removed completely due to refactoring.  I don't plan to change those functions in the near-term, but I do want to refactor `photutils.detection` at some point.  If API stability is critical, then you may want to package those current functions with `drizzlepac`.

### DIFF
--- a/drizzlepac/devutils/sourcelist_generation.py
+++ b/drizzlepac/devutils/sourcelist_generation.py
@@ -9,7 +9,8 @@ import sys
 from astropy.io import fits
 from astropy.stats import mad_std
 import numpy as np
-from photutils import aperture_photometry, CircularAperture, DAOStarFinder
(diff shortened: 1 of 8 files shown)


stsci-hack (approved):
These changes update the use of photutils to get the various modules and classes from the correct locations.  This makes the code a lot cleaner and can be merged.

PR #995, opened 2021-03-25
Updates for photutils >= 1.1.0

larrybradley:
This PR makes updates to the code for `photutils 1.1.0`:

* use the `coverage_mask` keyword in `Background2D`
* use the new `SourceCatalog` class

 Note that these changes *require* `photutils >= 1.1.0` as the minimum dependency.  Please check this PR carefully in case I missed any needed changes.

CC: @stsci-hack


stsci-hack:
We will not want to merge this PR until after our next release.

stsci-hack:
These changes have been picked up in #1096 instead to account for all the changes in the code on master since this PR was first submitted, including one of the files modified by this PR being deleted from the package altogether.  In addition, #1096  also corrects for the change in column names for columns from the SourceCatalog (formerly, source_properties()) table.

PR #996, opened 2021-03-28
Updates to catalog columns

mdlpstsci:
Updates made to the Point and Segment catalogs to make them more consistent in terms of column units, descriptions, and formats.  This references Jira HLA-507.  The X and Y column names (X-Center vs X-Centroid) were explicitly NOT changed at this time.  These name propagate to other modules.  Also, the issue with nulls/nans/"" was not addressed.


stsci-hack (approved):
These all make sense.

PR #1000, opened 2021-03-30
Deleted obsolete modules from drizzlepac repository

mdlpstsci:
haputils/se_source_generation.py was deleted from the repo.  It was the original code written to generate the segmentation catalog.
devutils/sourcelist_generation.py was deleted from the repo.  This code invoked se_source_generation.py.  

I suspect there are other modules which should be cleaned up.


stsci-hack (approved):
Always good to remove code that has already been replaced.

PR #1001, opened 2021-03-30
Add weight column to ref catalog for SVM alignment

stsci-hack:
These changes add a new column named 'weight' to the reference catalog used for aligning the FilterProduct in SVM processing.  The weight is based on the inverse of the errors of the proper motions from the reference catalog.  

The dependencies also get updated to insure that scikit-image gets installed as it is used by the SVM code.


stsci-hack:
More logic needs to be added to handle situations where the reference catalog does not have valid errors for the proper motions, perhaps not even have proper motions at all.

mdlpstsci (commented):
This looks good to me.  It also has the advantage that no interface has to change as the new information is incorporated into an existing table as a new column.

PR #1005, opened 2021-04-05
Insure IDCSCALE always exists

stsci-hack:
Some headerlets from the astrometry database are not entirely complete, lacking at least the IDCSCALE keyword.  These changes look for that keyword when applying an `a priori` WCS solution from the database, and when it is missing, it gets added to the SCI header.  

Association j900a5010 headerlet with hdrname = 'j900a5agq_flt_IDC_0461802ej' does not contain the 'IDCSCALE' keyword and triggered an Exception during standard processing.  This PR resolves the problems and allows this dataset, jdgq05020, and others like it to process successfully.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -1211,9 +1211,9 @@ def verify_gaia_wcsnames(filenames, catalog_name='GSC240', catalog_date=gsc240_d
                                 cd21 = fhdu[('sci', sciext + 1)].header['CD2_1']
                                 fhdu_idscale = round(np.sqrt(np.power(cd11, 2) + np.power(cd21, 2)) * 3600., 3)
                             # Set the value of the IDCSCALE keyword
-                            for sciext in range(num_sci):
-                                msg +=  'Adding IDCSCALE {} to {}[sci,{}]'.format(fhdu_idscale, fhdu.filename(), sciext + 1)
-                                fhdu[('sci', sciext + 1)].header['idcscale'] = fhdu_idscale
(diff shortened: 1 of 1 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -1193,7 +1193,27 @@ def verify_gaia_wcsnames(filenames, catalog_name='GSC240', catalog_date=gsc240_d
...
+                                    _hdrlet = fhdu[('HDRLET', extn)].headerlet
+                                    if 'idcscale' in _hdrlet[('SIPWCS', 1)].header:
+                                        fhdu_idscale = _hdrlet[('SIPWCS', 1)].header['idcscale']
+                                        break
+                            if fhdu_idscale is None:
+                                cd11 = fhdu[('sci', sciext + 1)].header['CD1_1']
+                                cd21 = fhdu[('sci', sciext + 1)].header['CD2_1']
+                                fhdu_idscale = round(np.sqrt(np.power(cd11, 2) + np.power(cd21, 2)) * 3600., 3)
+                            # Set the value of the IDCSCALE keyword
+                            for sciext in range(num_sci):
```
mdlpstsci, line 1214: You are already in a loop "for sciext" starting at line 1113.  It seems that scope rules may apply here, and Python does not mind.  However, for clarity and my sanity, I would use a different variable name than "sciext" for the loop at 1214.

stsci-hack: Good catch.  Fixed in a3de30f5.

mdlpstsci (approved):
Looks fine now.

PR #1006, opened 2021-04-06
Enable user epoch selection in creating GAIA catalogs

stsci-hack:
Requests for astrometric catalogs up until now have resulted in catalogs automatically corrected for the epoch of the observations.  This set of changes enables more control over what epoch the user can request for the astrometric catalog, specifically allowing the user to get the native epoch of the catalog.  These changes also remove the code that was limiting the returned catalog to only sources with measured proper motions, as many sources, in the GAIA catalog at least, don't have proper motions due to being external galaxies or globular clusters, yet they are perfectly valid.


mdlpstsci (approved):
Changes look fine.

PR #1007, opened 2021-04-07
Iterate over SkyMatch options

stsci-hack:
Some datasets triggered Exceptions during sky subtraction when using the 'match' option due to exactly overlapping input images and problems with the spherical geometry package.  This simple update adds a check for problems with the sky computation and if an AssertionError is raised (as happens when the spherical geometry error is triggered), it will simply try again with a simpler SkyMatch option that does not rely on spherical geometry to work.  

A verifiable test case still needs to be identified for testing of this change before it can be merged.

### DIFF
--- a/drizzlepac/sky.py
+++ b/drizzlepac/sky.py
@@ -349,11 +349,15 @@ def _skymatch(imageList, paramDict, in_memory, clean, logfile):
                  flog        = MultiFileLog(console = True, enableBold = False),
                  _taskname4history = 'AstroDrizzle')
     except AssertionError:
-        if paramDict['skymethod'] != 'localmin':
+        if 'match' in paramDict['skymethod']:  # This catches 'match' and 'globalmin+match'
+            if 'globalmin' in paramDict['skymethod']:
+                new_method = 'globalmin'
+            else:
+                new_method = 'localmin'
             # revert to simpler sky computation algorithm
             log.warning('Reverting sky computation to "localmin" from "{}'.format(paramDict['skymethod']))
             skymatch(new_fi,
-                     skymethod='localmin',
+                     skymethod=new_method,
                      skystat=paramDict['skystat'],
                      lower=paramDict['skylower'],
                      upper=paramDict['skyupper'],
--- a/drizzlepac/sky.py
+++ b/drizzlepac/sky.py
@@ -212,7 +212,7 @@ def _skymatch(imageList, paramDict, in_memory, clean, logfile):
 
     nimg = len(imageList)
     if nimg == 0:
-        log.info("Skymatch needs at least one images to perform{0} \
+        log.info("Skymatch needs at least one image to perform{0} \
                     sky matching. Nothing to be done.",os.linesep)
         return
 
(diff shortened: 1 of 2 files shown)


drizzlepac/sky.py
```diff
@@ -327,26 +327,52 @@ def _skymatch(imageList, paramDict, in_memory, clean, logfile):
...
+                 readonly    = not paramDict['skysub'],
+                 dq_bits     = None,
+                 optimize    = 'inmemory' if in_memory else 'balanced',
+                 clobber     = True,
+                 clean       = clean,
+                 verbose     = True,
+                 flog        = MultiFileLog(console = True, enableBold = False),
+                 _taskname4history = 'AstroDrizzle')
+    except AssertionError:
+        if paramDict['skymethod'] != 'localmin':
```
mcara, line 352:
```suggestion
        if 'match' in paramDict['skymethod']:
```

mcara (approved):
I made a small suggestion but this PR seems straightforward and looks good to me.

drizzlepac/sky.py
```diff
@@ -327,26 +327,52 @@ def _skymatch(imageList, paramDict, in_memory, clean, logfile):
...
+            # revert to simpler sky computation algorithm
+            log.warning('Reverting sky computation to "localmin" from "{}'.format(paramDict['skymethod']))
+            skymatch(new_fi,
+                     skymethod='localmin',
```
mcara, line 356:
```suggestion
                     skymethod='globalmin' if 'global' in paramDict['skymethod'] else 'localmin' 
```

stsci-hack: Done in 14adf4aa.

drizzlepac/sky.py
stsci-hack: Resolved in 14adf4aa

drizzlepac/sky.py
```diff
@@ -795,7 +823,7 @@ def getreferencesky(image,keyval):
     _refplatescale=image.header["REFPLTSCL"]
     _platescale=image.header["PLATESCL"]
 
-    return (_subtractedsky * (_refplatescale / _platescale)**2 )
+    return (_subtractedSky * (_refplatescale / _platescale)**2 )
```
mcara, line 826: How did this even work before???

mcara (dismissed):

drizzlepac/sky.py
stsci-hack: I guess it was never called...?!?

drizzlepac/sky.py
```diff
@@ -212,8 +212,8 @@ def _skymatch(imageList, paramDict, in_memory, clean, logfile):
 
     nimg = len(imageList)
     if nimg == 0:
-        ml.logentry("Skymatch needs at least one images to perform{0}" \
-                    "sky matching. Nothing to be done.",os.linesep)
+        log.info("Skymatch needs at least one images to perform{0} \
```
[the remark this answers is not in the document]
stsci-hack: Fixed in 8a93f16

mcara (approved):
I pre-eminently and wholeheartedly approve this PR and any future follow up commits.

PR #1008, opened 2021-04-08
Replaced deprecated STWCS calls

stsci-hack:
This PR replaces all instances of the deprecated function 'stwcs.altwcs.archiveWCS' with the new version 'stwcs.altwcs.archive_wcs'.  The change corrects a problem with WCSNAMEs being set to blank when running 'svm_quality_analysis.report_wcs' leading to direct images in Grism/Prism datasets to having no WCSNAME.  

The original problem was reproduced using dataset 'ib3a03', which was then used to verify that this change corrected the problem.


mdlpstsci (approved):
Straightforward fix - good to go.

PR #1010, opened 2021-04-19
Change API for align.perform_align

stsci-hack:
The API for 'align.perform_align' was updated to require 'catalog_list' and 'num_sources' as inputs from the user or calling function.  This provides clear control over these 2 user-adjustable parameters for this stand-alone function.  The only change in any logic was to pass through the 'num_sources' value to the method which would use that value.  Nothing in this PR should (theoretically, that is) alter the results of any processing.


mdlpstsci (approved):
Changes look appropriate.

PR #1011, opened 2021-04-21
Use correct weights from standardized catalogs

stsci-hack:
The weights used for alignment in SVM (and MVM) alignment were originally based on proper-motion errors, not positional errors.  This PR corrects that mistake while also insuring that 'supported catalogs' present a minimal set of standardized column names for positions, proper motions, and their errors.  

The changes in this PR were used to align SVM visit 'j94sb8' to GSC242 and GAIAeDR3 to verify that the code finally works as expected.


mdlpstsci (approved):
Changes look appropriate and provide good support for the multiplicity of catalogs!

PR #1013, opened 2021-04-22
Modified to have a single align_to_gaia method which loops over catalogs, fit methods, and fit geometries

mdlpstsci:
Implemented looping over catalogs, methods, and fit geometries where the fit geometries are the inner-most loop.  The same align_to_gaia method is used for both SVM and MVM.  Modified configuration files to support the looping.

### DIFF
--- a/drizzlepac/haputils/product.py
+++ b/drizzlepac/haputils/product.py
@@ -10,6 +10,7 @@
       * 2 : filter and total products combined using the improved astrometry, consistent pixel scale, and oriented to North.
       * 3 : (future) multi-visit mosaics aligned to common tangent plane
 """
+import copy
 import logging
 import sys
 import os
@@ -17,6 +18,7 @@ import traceback
 import shutil
 
 from astropy.io import fits
+from collections import OrderedDict
 from stsci.tools import logutil
 from stwcs import updatewcs
 import numpy as np
@@ -27,6 +29,8 @@ from .. import wcs_functions
 from . import align_utils
 from . import astrometric_utils as amutils
 from . import cell_utils
+
+import pdb
 
 MSG_DATEFMT = '%Y%j%H%M%S'
 SPLUNK_MSG_FORMAT = '%(asctime)s %(levelname)s src=%(name)s- %(message)s'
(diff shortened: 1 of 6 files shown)


drizzlepac/haputils/product.py
```diff
@@ -127,6 +131,222 @@ def generate_metawcs(self):
...
+
+                                log.info("Trying '{}' wth method '{}' using fit geometry '{}'.".
+                                         format(catalog_item, method_name, mosaic_fitgeom))
+                                # pdb.set_trace()
+                                align_table.imglist = align_table.perform_fit(method_name, catalog_item, ref_catalog,
+                                                                              fitgeom=mosaic_fitgeom)
+
+                                align_table.select_fit(catalog_item, method_name)
+                                align_table.apply_fit(headerlet_filenames=headerlet_filenames,
+                                                      fit_label=fit_label)
```
stsci-hack, line 268: Selecting a fit at this point will always result in the first attempt being defined as the 'best' (final) solution, then subsequent calls with other methods and catalogs will return an Exception.  Also, since 'determine_fits_quality_mvm_interface' has not been run yet, the determination has not yet been made as to whether the fit in 'select_fit' is even good.  These lines can only be called  after 'determine_fit...' at line 286.

drizzlepac/haputils/product.py
```diff
@@ -127,6 +129,220 @@ def generate_metawcs(self):
...
+                    log.debug('Creating reference catalog {}'.format(self.refname))
+                    ref_catalog = amutils.create_astrometric_catalog(align_table.process_list,
+                                                                     catalog=catalog_item,
+                                                                     output=self.refname,
+                                                                     gaia_only=False,
+                                                                     full_catalog=True)
+
+                    # Add a weight column which is based upon proper motion measurements
+                    if ref_catalog.meta['converted'] and ref_catalog['RA_error'][0] != np.nan:
+                        ref_weight = 1. / np.sqrt(ref_catalog['RA_error'] ** 2 + ref_catalog['DEC_error'] ** 2)
```
stsci-hack, line 203: The definition of the weight should be inverted for best results.  Please remove '1. /' from this line.

drizzlepac/haputils/product.py
mdlpstsci: Issues corrected.

drizzlepac/haputils/product.py
mdlpstsci: Issue corrected.

drizzlepac/haputils/product.py
```diff
@@ -127,6 +129,220 @@ def generate_metawcs(self):
 
         return meta_wcs
 
+    def align_to_gaia(self, catalog_list=[], headerlet_filenames=None, output=True,
```
stsci-hack, line 132: Input parameter 'headerlet_filenames' is never used, and gets overwritten on Line 144 even if it was provided as input.  So, this parameter can be removed as an input parameter.

stsci-hack (changes requested):
The comment about the API would best be handled now then it can be merged as far as I am concerned.

drizzlepac/haputils/product.py
mdlpstsci: Removed the unnecessary parameter in c163f8c.

stsci-hack (approved):
LGTM.

PR #1014, opened 2021-04-23
Apply limits on cross-matches to fit quality

stsci-hack:
These changes connect the varying limits for the fit based on the 'fitgeom' used for the fit to the code used in 'determine_fit_quality' for the 'min_xmatches' check.  This allows both the pipeline processing and SVM/MVM processing to use their own limits defined in the JSON config files.  

The limit for the max_fit_limit is also pulled from the config files in these changes as well.  

This allows dataset 'ic0s17' to end up with a 'shift' fit to GAIAeDR3 with 2 matches rather than ending up with a fit based on GSC242.

### DIFF
--- a/drizzlepac/align.py
+++ b/drizzlepac/align.py
@@ -740,6 +740,9 @@ def determine_fit_quality(imglist, filtered_table, catalogs_remaining, align_par
         radial_offset = math.sqrt(
             float(item.meta['fit_info']['shift'][0])**2 +
             float(item.meta['fit_info']['shift'][1])**2) * item.wcs.pscale  # radial offset in arssec
+        # Without the '+2', this will always fail for xmatches < 3 regardless of how small
+        # the offset is:  2*0.36 == 0.72 vs 0.8 + 0 [perfect alignment]
+        # Adding 2 allows low offset solutions with only 1 or 2 sources to pass this check.
         if float(num_xmatches + 2) * 0.36 > 0.8 + (radial_offset / 10.0)**8:
             radial_offset_check = True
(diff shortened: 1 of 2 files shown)


drizzlepac/align.py
```diff
@@ -729,18 +732,20 @@ def determine_fit_quality(imglist, filtered_table, catalogs_remaining, align_par
...
             nmatches_check = True
 
         radial_offset_check = False
         radial_offset = math.sqrt(
             float(item.meta['fit_info']['shift'][0])**2 +
-            float(item.meta['fit_info']['shift'][0])**2) * item.wcs.pscale  # radial offset in arssec
-        if float(num_xmatches) * 0.36 > 0.8 + (radial_offset / 10.0)**8:
+            float(item.meta['fit_info']['shift'][1])**2) * item.wcs.pscale  # radial offset in arssec
+        if float(num_xmatches + 2) * 0.36 > 0.8 + (radial_offset / 10.0)**8:
             radial_offset_check = True
```
mdlpstsci, line 744: These look like "magic" numbers, especially the newly added "2".  If they are not going to be in a configuration file, please just add a comment as to the origin/meaning of these values.

mdlpstsci (approved):

drizzlepac/align.py
stsci-hack: The change from '0' to '1' simply reflects the use of the shift in both X and Y (as needed when computing a radial value) where before the X value was simply added twice.  The '+2' was commented in ad5c3a7.

PR #1016, opened 2021-04-29
Insure correct versions get reported for WCS

stsci-hack:
Applying a headerlet from an extension will cause these keywords to be overwritten with the values from the headerlet, which will most likely not be the current versions in use.  These simple changes insure that all a posteriori WCS headerlets created during pipeline processing, along with the FLT/FLC image primary header, get updated with the version of STWCS and Astropy actually used to create the new WCS.  Since this change affects code used by both pipeline processing and SVM/MVM processing, we are insuring all future processing will report the correct versions used to create the WCS in use.

### DIFF
--- a/drizzlepac/haputils/align_utils.py
+++ b/drizzlepac/haputils/align_utils.py
@@ -10,6 +10,7 @@ from collections import OrderedDict
 import numpy as np
 from scipy import ndimage
 
+import astropy
 from astropy.io import fits
 from astropy.table import Table
 from astropy.nddata.bitmask import bitfield_to_boolean_mask
@@ -20,6 +21,7 @@ from photutils import background
 from photutils.background import Background2D
 from photutils.utils import NoDetectionsWarning
 
+import stwcs
 from stwcs.wcsutil import HSTWCS
 from stwcs.wcsutil import headerlet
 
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/align_utils.py
```diff
@@ -1067,6 +1069,33 @@ def update_image_wcs_info(tweakwcs_output, headerlet_filenames=None, fit_label=N
...
+            before_kw = 'HISTORY'
+        elif 'ASN_MTYP' in hdulist[0].header:
+            after_kw = 'ASN_MTYP'
+            before_kw = None
+        else:
+            after_kw = hdulist[0].header.cards[-1][0]
+            before_kw = None
+
+        hdulist[0].header.set('UPWCSVER', value=upwcsver,
+                        comment="Version of STWCS used to updated the WCS",
```
mdlpstsci, line 1093: Please fix the typo (updated -> update).

mdlpstsci (changes requested):
Please update the commentary part of the UPWCSVER card. 

I also have a question.  You specifically position these cards in the header, so I would like to ensure that they are overriding the current keywords with these values.

drizzlepac/haputils/align_utils.py
stsci-hack: Corrected in fc66c22.

stsci-hack:
> 
> 
> Please update the commentary part of the UPWCSVER card.
As noted before, fixed in fc66c22.
> 
> I also have a question. You specifically position these cards in the header, so I would like to ensure that they are overriding the current keywords with these values.

The use of 'header.set()' will not only add a new keyword if it was not only present, but modify one in place as well.  Use of the 'before' and 'after' parameters affects the final location of the keyword, moving it from the original location to a new one when updating a current keyword in the header.  This behavior was confirmed through manual verification of this behavior.

mdlpstsci (approved):

PR #1017, opened 2021-04-29
Adjust threshold for point-catalog generation

stsci-hack:
The background used for point-source detection based on PSFs has been revised to use the full detection threshold (bkg + nsigma*rms) instead of just the background by itself.  This will resolve the most egregious cases where peaks in the background get picked up as 'real' sources as noted by INS for 'id7h42' and 'ica9t3'.


mdlpstsci (approved):
It is a straightforward solution.

PR #1018, opened 2021-04-30
Clarified message issued when a downgrade in the fit geometry is done.

mdlpstsci:
Clarified message issued when a downgrade in the fit geometry is done.  The message was fixed to ensure the correct fit geometry name is issued.


stsci-hack (approved):
LGTM

PR #1023, opened 2021-05-05
Add logic to avoid relative fit with singletons

stsci-hack:
This restores the logic to avoid trying to perform a relative fit when there is only 1 exposure to align, as some visits only contain a single exposure (ibe809 and idcm13 for example).


mdlpstsci (approved):
Good catch!!!

PR #1027, opened 2021-05-06
Update default fit pars

stsci-hack:
These adjustments to the parameters used for alignment allow data which was previously unable to be aligned or aligned correctly to actually get aligned to GAIA.  These changes were demonstrated to work for 'ibe809' which ended up having an offset of (236, -347) pixels or 16.6 arcseconds.  In addition, the 2dhist fitting option was restored for mosaic_fitting use since it relies on a very different algorithm from the default fitting for cross-matching and determination of the offset providing complimentary techniques for trying to align single exposures to GAIA.  In fact, the 2dhist option was the only way to align 'ibe809'.


mdlpstsci (approved):
These changes seem fine.  I will note that I have really started to wonder if json was the best choice for the configuration files because according to the spec, json does not allow comments.  Because we have modified these files (thinking of the numeric values) based on our experience with a variety of datasets, it would be nice to know why (aside from trolling through lots of GitHub PRs) a value was modified.  Just floating this thought for a future life/project.

stsci-hack:
Agreed that these files do not allow us to track why the specific values are defined as they are.  Something to consider for later...

PR #1028, opened 2021-05-06
Update inputs for fitting

stsci-hack:
Additional information will be needed (eventually) in order to fit for proper motions as well as offsets during SVM/MVM processing.  These changes add epoch information to the metadata of each input to 'tweakwcs' as the first (of many to come?) changes to support the expanded fitting options.

**The installation requirements for Drizzlepac also were updated with this PR to reflect use of fixes to `spherical_geometry` and `tweakwcs` that were implemented to support or enhance the changes in this PR.**

This PR also includes some revisions to the RTD doc build to try to address problems seen building the docs on RTD.


drizzlepac/haputils/astrometric_utils.py
```diff
@@ -1813,6 +1818,30 @@ def build_wcscat(image, group_id, source_catalog):
...
+    d = hdulist[0].header['date-obs']
+    dt = datetime.strptime(d, '%Y-%m-%d')
+
+    return decimal_year(dt)
+
+
+def decimal_year(date):
+    def since_epoch(date): # returns seconds since epoch
+        return time.mktime(date.timetuple())
+    se = since_epoch
```
mdlpstsci, line 1832: Just a comment - No sure why you encapsulated the computation for seconds since the epoch inside of another function as SSE can be generally very useful.

doc/.rtd-environment.yml
```diff
@@ -27,7 +27,7 @@ dependencies:
       - stsci.imagestats
       - stsci.skypac
       - stsci.stimage
-      - "tweakwcs>=0.4.0"
+      - "tweakwcs>=0.7.0"
       - stregion
```
mdlpstsci, line 31: Does this setting for tweakwcs reconcile with setup.py?  I am not sure the order of evaluation.

mdlpstsci (changes requested):
Not really looking for changes, but want to check/explain an item.

drizzlepac/haputils/astrometric_utils.py
stsci-hack: This is what happens when I blindly cut-and-paste from StackOverflow. ;-)  Actually, using 'astropy.time' to get the decimalyear is cleaner, and the change to use astropy's computation is included in 8f33c83.

doc/.rtd-environment.yml
stsci-hack: Actually, that would not have caused problems since the installation using 'setup.py' would have forced a newer version (0.7.2) to be installed.  However, this has been updated to be consistent with `setup.py` in 8f33c83 just to prevent any problems.

mdlpstsci (approved):
The astropy time certainly is streamlined!

PR #1029, opened 2021-05-07
Update defaults for pipeline catalog use

stsci-hack:
After updating the SVM processing with a new set of catalogs, this updates the options for alignment in standard pipeline processing of ASNs and singletons to be consistent.


mdlpstsci (approved):
Clean update!

PR #1033, opened 2021-05-20
Skip inconsistent fits during relative alignment

stsci-hack:
These changes identify and flag as 'FAILED' any relative alignment solution (performed using `align_utils\match_relative_fit`) where any member exhibits a wildly different rotation.  This typically results from a nearly singular solution with very few (2 or 3) matches or from incorrect cross-matching. 

By setting the `meta["fit_info"]["status"}` to 'FAILED' for all members in the fit, that fit solution will be thrown out and the rest of the code will try the next possible solution.  This logic has only been added to the relative fit solution, since there is an expectation that any corrections will be relatively small in rotation at least.  

These changes were tested using 'ibvg03' under Windows which only uses float64 as long, instead of float128 used under Linux, resulting in a near singular solution for a couple of the 30 exposures being used to create the mosaic.   

In addition, these changes correctly pass in the 'minobj' values for each 'fitgeom' to 'tweakwcs'.

### DIFF
--- a/drizzlepac/align.py
+++ b/drizzlepac/align.py
@@ -688,6 +688,9 @@ def determine_fit_quality(imglist, filtered_table, catalogs_remaining, align_par
     yshifts = []
     overall_valid = True
     overall_comp = False
+
+    do_consistency_check = align_pars['determine_fit_quality'].get('consistency_check', True)
+
     for item in imglist:
         if item.meta['fit_info']['status'].startswith('FAILED') is False:
             xshifts.append(item.meta['fit_info']['shift'][0])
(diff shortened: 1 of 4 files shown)


mdlpstsci (approved):

stsci-hack:
The last commit completes the revisions to correctly pass the 'fitgeom'-specific minobj values to 'tweakwcs' during alignment fitting. The previous code only passed in a single value regardless of the type of fit ('fitgeom').  Now, the code will pass in the value based on the 'fitgeom' selected, as originally designed (if I remember correctly).

mcara:
Too late to comment, but there is no need to specify `minobj` since `align_wcs` automatically determines that based on `fitgeom` - see https://tweakwcs.readthedocs.io/en/latest/imalign.html 

With regard to detecting "bad fits" - I feel computing rms-like quantities of `xshift` and `yshift` are not very relevant. Different images can have very different shifts. This is not necessarily a sign of failure. I think a much better and reliable sign of failure are large differences in computed skews: if all images have the same "handedness" - skews should be within a couple of degrees - not close to 180 degree swings.

PR #1043, opened 2021-06-02
Define bounded WCS for SkyCell footprint

stsci-hack:
This PR adds a new attribute to the SkyFootprint class, bounded_wcs, which represents the WCS of just the section of the SkyCell where the footprint has HST data.  This attribute gets automatically computed upon completion of the '.build()' method, although the method could be called again separately if needed. 

In addition, during testing, it was discovered that the similarity computations have been broken for a few months.  This PR includes a fix for this long-standing bug as well so that that similarity check now gets performed as designed.

### DIFF
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -210,6 +210,7 @@ class SkyFootprint(object):
         self.meta_wcs = meta_wcs
         # bounded_wcs corresponds to WCS of bounding box of footprint
         self.bounded_wcs = None
+        self.bounding_box = None
 
         # the exp_masks dict records the individual footprints of each exposure
         self.exp_masks = {}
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/cell_utils.py
```diff
@@ -300,6 +303,25 @@ def build(self, expnames, scale=False, scale_kw='EXPTIME'):
 
             self.total_mask += self.exp_masks[exposure]['mask']
 
+            # Compute the bounded WCS for this footprint
+            self.find_bounded_wcs()
+
+
+    def find_bounded_wcs(self):
+        """Compute the WCS based on the bounding box of footprint """
```
nden, line 311: This is a terminology comment. In GWCS/astropy and hence JWST and Roman, the bounding box is defined in the inputs space, i.e. it's in pixels, while the footprint is defined in outputs space, i.e. on sky.
So essentially bounding box and footprint are the same thing but in different coordinate frames.
Unless there's a good reason it's best to keep similar terminology to decrease confusion.

stsci-hack: The comment here and earlier has been edited.  The use of the term 'bounding box' seems to be consistent with what is used elsewhere since it is specified here in pixel space, but footprint was misused, so this reference  to 'footprint' was reworded, along with comment associated with the call to this function.  The changes are in 6abb063.  Hopefully, this is sufficient at this time.

mdlpstsci (approved):
Additional comments updated for consistency when referencing pixel coordinates vs equatorial coordinates, and bug when looking for SCI data in Extension vs Primary fixed - looks good!

PR #1044, opened 2021-06-02
make_poller_files.py: minor bugfix

dulude:
Minor bugfix.
make_poller_files.py: fixed incorrectly set variable name. Changed "imgname" to "fullfilepath" at line 60.


stsci-hack (approved):
Subtle bug which would only show up when the image is not in the current directory.  Good catch!  LGTM.

PR #1046, opened 2021-06-04
Insure IDCTAB is current in FLT files

stsci-hack:
The IDCTAB in the FLT/FLC files have reverted to older IDCTABs in the archive.  This implements support for a new environment variable, **PIPELINE_RESET_IDCTAB**, to enable this new function, `reset_idctab_kw`,  to sync the IDCTABs (and corresponding PRIMARY WCS) in the FLT/FLC files with the newest IDCTAB specified in the RAW file header.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -225,6 +225,7 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
 
     # Add support for environment variable switch to automatically
     # reset IDCTAB in FLT/FLC files if different from IDCTAB in RAW files.
+    reset_idctab_switch = False
     if envvar_reset_idctab_name in os.environ:
         val = os.environ[envvar_reset_idctab_name].lower()
         reset_idctab_switch = envvar_bool_dict[val]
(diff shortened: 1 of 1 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -1233,6 +1253,48 @@ def restore_pipeline_default(files):
...
+        newmsg = "Updating IDCTAB {} in {}\n".format(flt_idctab, flt)
+        if logfile:
+            # Write message out to temp file and append it to full trailer file
+            _updateTrlFile(logfile, newmsg)
+        else:
+            print(newmsg)
+
+        # Get info from all hdrlet extensions in file
+        wnames = headerlet.get_headerlet_kw_names('j8mbnlkbq_flc.fits', 'wcsname')
+        hnames = headerlet.get_headerlet_kw_names('j8mbnlkbq_flc.fits', 'hdrname')
```
mdlpstsci, line 1288: Somehow I do not think you meant to hardcode this explicit filename in this method to get the wname and hnames!

stsci-hack: Like a broken clock, it would at least be right 1 time during reprocessing.  
Fixed in f6ffd00.

mdlpstsci (approved):
Changes look fine.

mdlpstsci (changes requested):
Oops - tests found.  If line 228 is not true, then the variable "reset_idctab_switch" is not set.

stsci-hack:
Problem with tests should be resolved with 141ca84.

mdlpstsci (approved):
Looks good now.  Sorry to iterate as I should have caught this lack of initialization in the first place.

PR #1047, opened 2021-06-04
Initialize fit for each attempt

stsci-hack:
This moves the initialization of the fit from outside the iterations over fitgeoms and fit type to inside so that the same starting WCS is used for each alignment attempt.

### DIFF
--- a/drizzlepac/align.py
+++ b/drizzlepac/align.py
@@ -342,8 +342,6 @@ def perform_align(input_list, catalog_list, num_sources, archive=False, clobber=
         else:
             alignment_table.find_alignment_sources(output=output)
 
-        alignment_table.configure_fit()
-
         for imgname in alignment_table.extracted_sources.keys():
             table = alignment_table.extracted_sources[imgname]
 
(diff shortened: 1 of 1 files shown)


drizzlepac/align.py
```diff
@@ -441,13 +440,16 @@ def perform_align(input_list, catalog_list, num_sources, archive=False, clobber=
...
                     log.info(
                         "{} Catalog {} matched using {} {}".format("-" * 18,
                                                                    catalog_name,
                                                                    algorithm_name, "-" * 18))
                     try:
+                        alignment_table.configure_fit()
+                        print("####\n# Running configure fit for AlignmentTable to use: \n")
+                        print([img.wcs for img in alignment_table.imglist])
+                        print("####\n")
+
```
mdlpstsci, line 452: Please change these print statements to log statements of some flavor (info or debug).

mdlpstsci (changes requested):
Just a small change, and then this work is ready to go!

drizzlepac/align.py
stsci-hack: Fixed in e61b53a.

mdlpstsci (approved):

PR #1048, opened 2021-06-08
Fix memory corruption in the linear blot algorithm

mcara:
A port of a fix for a bug originally discovered in the `drizzle` - see https://github.com/spacetelescope/drizzle/pull/66


stsci-hack (approved):
LGTM

mcara:
Thanks! @stsci-hack

PR #1049, opened 2021-06-09
Minimize memory usage of SkyFootprint

stsci-hack:
One of the primary limitations for processing MVM datasets has been the memory requirements for the SkyFootprint class.  These changes reduce the memory use for that class to a single np.int16 array for the total_mask and, when requested, a single np.float32 array for the exposure time scaled total_mask.  These arrays are the size of the final output frame, which for MVM processing is the full size of a SkyCell or 21418 x 21418 pixels.  This memory usage does not go up noticeably with the number of input exposures, unlike the original implementation.  

One of the major benefits of this PR is that the **runtime was also dramatically sped up** now taking only about 1second per input image, **about 10X faster**, give or take.

The **'get_sky_cells' function was also fixed** to work correctly with these changes, while also minimizing it's memory usage and while speeding up its operation as well (by orders of magnitude).  

**_In addition,_** a logic problem was fixed in the GridDefs class for defining the ProjectionCell at the south pole.

### DIFF
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -874,7 +874,7 @@ class ProjectionCell(object):
         for xi in range(mosaic_xr[0], mosaic_xr[1]):
             for yi in range(mosaic_yr[0], mosaic_yr[1]):
                 skycell = SkyCell(x=xi, y=yi, projection_cell=self)
-                skycell.build_mask()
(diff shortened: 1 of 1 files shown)


stsci-hack:
Need to sort out code conflicts with #1043 before it can be reviewed and merged.

mdlpstsci (approved):
Need to give this code whirl.

PR #1051, opened 2021-06-16
WIP: Changes made to use the "bounding box WCS" 

mdlpstsci:
Changes made to use the "bounding box WCS" which are the actual exposed pixels of all the exposures within a skycell.


drizzlepac/hapmultisequencer.py
```diff
@@ -225,7 +225,7 @@ def run_mvm_processing(input_filename, diagnostic_mode=False, use_defaults_confi
 
         # Update the SkyCellProduct objects with their associated configuration information.
         for filter_item in total_obj_list:
-            filter_item.generate_metawcs()
+            _ = filter_item.generate_metawcs()
```
stsci-hack, line 228: This makes me wonder whether 'generate_metawcs' should not be modified to NOT return a value.  After all, it is not labelled 'get_metawcs'.  The only place where this is expected to return a value is 1 time in hapsequencer, and that call could easily be modified.  Just wondering what would be more consistent...

stsci-hack (commented):
Only 1 question regarding 'generate_metawcs'.  Once that is resolved, everything else looks fine to me.

drizzlepac/hapmultisequencer.py
mdlpstsci: I agree that changing generate_metawcs() to only set an attribute and not return a value would be overall more consistent.  Note FilterProduct (SVM) calls its version of this function and uses the return value.  The SkyCellProduct (MVM) calls its own version of this function three times, and only uses the return value once.  Having said this, the value is "meta_wcs" which is getting passed around.  I do think this needs to be cleaned up for consistency, but I am reluctant to do it in this ticket as it should be done very carefully.  Please provide your final assessment as to what should be part of this ticket, and what should go into its own ticket.

stsci-hack (approved):
Changes to 'generate_metawcs' can be performed as part of a separate PR.  The implementation in this PR looks quite logical to me.

PR #1054, opened 2021-06-24
Use LoG filter for detecting alignment sources

stsci-hack:
These changes implement use of Laplacian-of-Gaussian (LoG) filter as the method for detecting sources to be used for alignment, instead of using photutils.detect_sources.   The use of LoG filter here results in a segmentation map compatible with the one generated by photutils and is used as a direct replacement here.  

This simple replacement, though, allows data from 'ibr623' to align successfully whereas use of photutils resulted in severe misalignment of this single-source (saturated source) set of images.  The dataset 'iccq01' should also be used as a test dataset to verify that the correct alignment can be determined using this filter.  

Extensive testing will need to be performed before this change can be merged to make sure that it does not cause problems with alignment in other fields, especially those heavily contaminated by CRs, highly crowded fields, fields with only a few heavily saturated sources, fields with only faint sources, and so on.  These types of fields can affect the results generated using the LoG filter since it will detect the sources very differently.

### DIFF
--- a/drizzlepac/haputils/astrometric_utils.py
+++ b/drizzlepac/haputils/astrometric_utils.py
@@ -65,6 +65,7 @@ from stsci.tools import logutil
 from stsci.tools.fileutil import countExtn
 
 from ..tweakutils import build_xy_zeropoint, ndfind
+from . import laplace_point_cat as lpc
 
 __taskname__ = 'astrometric_utils'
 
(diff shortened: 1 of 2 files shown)


stsci-hack:
This change may be worth testing against the WFPC2 data that failed to align; specifically, data from visit 'u2lx05'.

stsci-hack:
This can be reviewed at a later date to see whether there is any value in pursuing this algorithm.

PR #1056, opened 2021-06-24
Fix list input parsing for get_sky_cells

stsci-hack:
The expnames are now correctly parsed from a file as input to 'get_sky_cells' with these changes.


mdlpstsci (approved):
Update is straightforward and does the job!

PR #1065, opened 2021-07-12
Fix indexing of SkyCell IDs

stsci-hack:
These changes insure that the indices used for the SkyCell ID are always between 1 and 21 (or whatever sc_nxy was defined to be in the pars/allsky_cells.fits table).  The logic prevents negative skycell indices from being tried or used as well.  

The data from `idix03` which overlapped skycells p1894 and p1974 was used to confirm that the negative Y index was corrected.


dulude:
This PR also resolves ticket [HLA-589](https://jira.stsci.edu/browse/HLA-589).

dulude (approved):

PR #1067, opened 2021-07-13
WIP: Address unreasonable deblending times for segments 

mdlpstsci:
Restructured code to allow for modest recomputation with regard to identifying
segments with the custom/Gaussian or RickerWavelet kernel.  If the segmentation
image is found to have big islands or a large source fraction after both
kernels have been tried, the algorithm allows for a recomputation of the background
image (sigma-clipped to Background2D) or raising the threshold for detection if the
Background2D is already in use.  Pending: last ditch thresholds on big islands and/or
source fraction which allow segmentation processing to continue as the deblending
should statistically not take days and days (based upon test results).  I still
need to analyze the stats to determine these thresholds.

Reset some configuration variables, keep track of the background determination
type (zero_background, sigma_clipped_background, twod_background), reset class
attribute to a local variable, switch order of catalog computation (segmentation is
now done first), ensure use of the "custom kernel" for the segmentation detection.
Currently an incomplete update of all configuration files.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -82,7 +82,7 @@ class CatalogImage:
         self.bkg_rms_median = None
         self.footprint_mask = None
         self.inv_footprint_mask = None
-        self.negative_threshold = 0.0
+        self.bkg_type = ""
 
         # Populated by self.build_kernel()
         self.kernel = None
@@ -201,14 +201,19 @@ class CatalogImage:
             Inverse of the footprint_mask
 
         """
+        # MDD
+        # Negative allowance in sigma
+        negative_sigma = -1.0
+        #negative_sigma = -1.5
+
         # Report configuration values to log
         log.info("")
         log.info("Background Computation")
         log.info("File: {}".format(self.imgname))
         log.info("Zero threshold: {}".format(zero_percent))
         log.info("Sigma-clipped Background Configuration Variables")
-        log.info("  Negative threshold: {}".format(self.negative_threshold))
-        log.info("  Negative percent: {}".format(negative_percent))
+        log.info("  Negative percent threshold: {}".format(negative_percent))
+        log.info("  Negative sigma: {}".format(negative_sigma))
         log.info("  Nsigma: {}".format(nsigma_clip))
         log.info("  Number of iterations: {}".format(maxiters))
         log.info("Background2D Configuration Variables")
@@ -245,6 +250,7 @@ class CatalogImage:
             self.bkg_rms_median = stats.tstd(non_zero_pixels, limits=[0, None], inclusive=[False, True])
             self.bkg_background_ra = np.full_like(imgdata, 0.0)
             self.bkg_rms_ra = np.full_like(imgdata, self.bkg_rms_median)
+            self.bkg_type = 'zero_background'
 
             is_zero_background_defined = True
             log.info("Input image contains excessive zero values in the background. Median: {} RMS: {}".format(self.bkg_median, self.bkg_rms_median))
@@ -322,7 +328,8 @@ class CatalogImage:
             self.bkg_rms_ra = np.full_like(imgdata, bkg_rms)
             self.bkg_median = bkg_median
             self.bkg_rms_median = bkg_rms
-            self.negative_threshold = -1.0 * bkg_rms
+            self.bkg_type = 'sigma_clipped_background'
+            negative_threshold = negative_sigma * bkg_rms
 
         # BACKGROUND COMPUTATION 3 (Background2D)
         # The simple_bkg = True is the way to force the background to be computed with the
@@ -359,7 +366,7 @@ class CatalogImage:
                         bkg_rms_ra = bkg.background_rms
                         bkg_rms_median = bkg.background_rms_median
                         bkg_median = bkg.background_median
-                        self.negative_threshold = -1.0 * bkg.background_rms_median
+                        negative_threshold = negative_sigma * bkg.background_rms_median
                         break
 
                 # If computation of a two-dimensional background image were successful, compute the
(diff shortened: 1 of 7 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -201,14 +201,17 @@ def compute_background(self, box_size, win_size,
...
+
         # Report configuration values to log
         log.info("")
         log.info("Background Computation")
         log.info("File: {}".format(self.imgname))
         log.info("Zero threshold: {}".format(zero_percent))
         log.info("Sigma-clipped Background Configuration Variables")
-        log.info("  Negative threshold: {}".format(self.negative_threshold))
-        log.info("  Negative percent: {}".format(negative_percent))
+        log.info("  Negative percent threshold: {}".format(negative_percent))
```
stsci-hack, line 213: I am a little confused about this log statement.  A threshold is a limit used for discrimination, whereas the 'negative_percent' seems to report the percent (number) of pixels that are negative which would not be consistent with the limit.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -1386,19 +1394,16 @@ def identify_sources(self, **pars):
...
 
-            # Compute the threshold to use for source detection
-            threshold = self.compute_threshold(self._nsigma, self.image.bkg_background_ra, self.image.bkg_rms_ra)
-
-            # Define smoothing kernel which overrides the use of the custom kernel
-            sigma = 1.0 * gaussian_fwhm_to_sigma
-            g2d_kernel = Gaussian2DKernel(sigma, x_size=3, y_size=3)
-            g2d_kernel.normalize()
+            # Custom or Gaussian kernel depending upon the results of CatalogImage build_kernel()
+            g2d_kernel = self.image.kernel
```
stsci-hack, line 1406: Would it be possible for 'self.image.kernel' to not get defined?  If so, the old code may need to be used as a default case.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -1414,111 +1419,165 @@ def identify_sources(self, **pars):
...
+                        rw2d_sigma_for_threshold = self._rw2d_nsigma * 2.0
+
+                    # Detect segments and evaluate the detection in terms of big islands or crowded fields
+                    # Round 2
+                    ncount += 1
+                    log.info("")
+                    log.info("With alternate background...using Gaussian kernel to generate a segmentation map.")
+                    del g_segm_img
+                    g_segm_img, g_is_big_crowded, g_bs, g_sf = self.detect_and_eval_segments(imgarr,
+                                                                                             g2d_kernel,
```
stsci-hack, line 1513: The kernel was recomputed on line 1492 when calling 'build_kernel' again.  Did this redefine the kernel pointed to by 'g2d_kernel' or does 'g2d_kernel' need to be reset based on the new results from line 1492?  The code assumes it got redefined in place.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -1414,111 +1419,165 @@ def identify_sources(self, **pars):
...
+                            self.kernel = rw2d_kernel
+                            segm_img = copy.deepcopy(rw_segm_img)
+                            del rw_segm_img
+                        # The segmentation image is problematic and the big island/source fraction limits are exceeded,
+                        # so deblending could take days, and the results would not be viable in any case.
+                        else:
+                            log.info("")
+                            log.info("The second round of segmentation images still contain big islands or a\n"
+                                     "large source fraction of segments.")
+                            log.info("The segmentation algorithm is unable to continue.")
```
stsci-hack, line 1558: Where will the log message be generated that tells the user that no segmentation catalog will be generated?  It would seem that here would be appropriate if it is not being reported elsewhere.

stsci-hack (commented):
There were a few questions that came to mind as I reviewed this code.  Nothing jumps out as an obvious issue that needs to be corrected, but some might require revision of the code.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: The self.kernel, self.kernel_fwhm, and self.kernel_psf are attributes of the CatalogImage class, and represent the Custom (image-specific) or Gaussian kernel, as appropriate.  The self.image is an instantiation of CatalogImage for the HapCatalogs base class and is a parameter used in the constructor for the HAPSegmentCatalog/HAPPointCatalog.  Fundamentally, the kernel is derived in the build_auto_kernel() function of the astrometric_utils.py and seems to always be define deliberately.

If you still think this is risky, I can add the old code back as default.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: Excellent catch!  Indeed the g2d_kernel _**and**_ the rw2d_kernel, need to be reset when a new background type is computed.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: Added the warning that no segmentation catalog will be produced.  Also, changed some  of the former "info" log messages to "warning" log messages.

mdlpstsci:
I thought there was a question/issue with the self.negative_threshold and negative percent threshold.  Here is my response.

The self.negative_threshold was changed to a local variable, negative_threshold, where it is computed and actually used, so it was removed from this "configuration" log information.  Also,
`negative_threshold = negative_sigma * bkg_rms` 
or
`negative_threshold = negative_sigma * bkg.background_rms_median`

The negative_threshold is used in the computation for determining how much of the illuminated portion of the background-subtracted image is negative.  The negative_threshold represents the limit **below** which pixels are considered "negative".  The percentage of "negative" pixels is determined and compared against the acceptable or expected "negative percent threshold".  Effectively, if there are many more negative pixels than expected, the algorithm forces the use of the sigma-clipped background.  No change has been done here.  If this still does not make good sense, then let us discuss.

mdlpstsci (commented):
I did not mean to do "start a review", and I do not know how to abort it. So I will just submit it - duh.

mdlpstsci:
In update 6b923fb I addressed two issues raised in the review (update of kernels and fixed/added log messages).  There were two other comments for which I have provided a response, but there were no corresponding code changes.  There may need to be further discussion.

drizzlepac/haputils/catalog_utils.py
stsci-hack: This just stood out as a significant change, and it wasn't immediately obvious that this was always going to be defined.  Your evaluation seems to verify that these attributes are always defined, so I don't see any problem with this part of the code.

stsci-hack (approved):
Based on testing and updates to code made on previous code comments, this PR can now be merged as far as I am concerned.

PR #1070, opened 2021-07-16
Fix the construction of the manifest filename for MVM processing

mdlpstsci:
Fixed the construction of the manifest filename for MVM processing.  The manifest name is now of the form hst_skycell-p<PPPP>x<XX>y<YY>_manifest.txt.   The name is solely dependent upon the skycell name which should make it predictable for the operations software.


stsci-hack (approved):
Simple enough changes that look fine to me.

Issue #1069, opened 2021-07-14
Make astroquery_utils more useful


stscijgbot-hstdp:
_Issue [HLA-593](https://jira.stsci.edu/browse/HLA-593) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

One of the weaknesses of 'haputils/astroquery_utils' remains the requirement that it can only work with specific ASN or exposure filenames. 

This can be significantly improved by adding the ability to support requests using only the 'obs_id' as a wildcard; for example, by passing in a query such as:

    *obstab = Observations.query_criteria(dataproduct_type=["image"], obs_id="ib3m53\*")*
    *expnames = obstab['obs_id'].tolist()*

to get all the data associated with visit 'ib3m53'.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-593?focusedCommentId=539025#comment-539025):_

Current interface for 'haputils/astroquery_utils' already supports use of wildcards for inputs; for example, 
{{
files = aqutils.retrieve_observations(['ib3m53*'], suffix=['FLC', 'ASN'])
}}
will return all the exposures and ASN files associated with this visit.

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-593?focusedCommentId=539030#comment-539030):_

The docstring for this function has been updated under [Drizzlepac PR #&#x2060;1072](https://github.com/spacetelescope/drizzlepac/pull/1072) to make it more obvious to users how to get data for an entire visit (for example) using wild-cards.

PR #1072, opened 2021-07-16
Update docs for aqutils

stsci-hack:
Clarify documentation for 'haputils/astroquery_utils/retrieve_observation' to make it more obvious how to use the function to retrieve all data associated with a visit.


stsci-hack:
Closes #1069

mdlpstsci (approved):
Definitely better documentation for this particular (and important) parameter now.

PR #1073, opened 2021-07-16
Fix the code in cell_utils.py which is generating a FutureWarning

mdlpstsci:
Fixed the code in cell_utils.py which was generating a FutureWarning of "Using a non-tuple sequence for multidimensional indexing is deprecated. Use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.".  Fixing this issue eliminates unnecessary clutter in the output trailer files.


stsci-hack (approved):
LGTM.

PR #1077, opened 2021-07-21
Skip drizzlepac PyTests for HSTDP-2021.3.3. Build

mdlpstsci:
Pytest skip drizzlepac tests due to mismatch between code version and artifactory truth files.


stsci-hack (approved):
LGTM

PR #1078, opened 2021-07-22
Only apply DRIZPARS when returned by tweakwcs

stsci-hack:
This corrects the problem with trying to update the DRIZPARS keyword in an empty DRZ file or one for which AstroDrizzle did not create a DRZ product at all.  

The problem was identified with dataset 'ib8c5k'.


mdlpstsci (approved):
Straightforward solution.

PR #1079, opened 2021-07-22
Fix fitgeom use in log msg

stsci-hack:
This fixes the problem where a log message was trying to be written out using a field that 'tweakwcs' did not create for an image (reference image?).  

Found using dataset 'ibcd61'.


mdlpstsci (approved):
Straightforward fix.

PR #1080, opened 2021-07-22
Fix to handle an undefined manifest filename

mdlpstsci:
Fix to handle a dataset where no images are viable for processing so
no total objects are created.  As a result, the manifest name cannot
be generated in the standard way and upon exit, the process ends on
an error. Addressed the situation so the process ends in a tidy manner.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -548,6 +548,7 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
     cat_switches = {sw: _get_envvar_switch(sw, default=envvar_cat_svm[sw]) for sw in envvar_cat_svm}
 
     total_obj_list = []
+    manifest_name = ""
     try:
         # Parse the poller file and generate the the obs_info_dict, as well as the total detection
         # product lists which contain the ExposureProduct, FilterProduct, and TotalProduct objects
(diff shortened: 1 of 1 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -686,6 +687,14 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
         logging.exception("message")
 
     finally:
+        # Try to ensure there is a manifest filename if the code exits in the poller_utils.py
+        # so the shutdown is tidy
+        ntokens = len(input_filename.split("_"))
+        if manifest_name == "" and input_filename.lower().endswith("input.out") and ntokens == 4:
```
stsci-hack, line 693: It doesn't need to be done as part of this PR, but this does add another dependency in the code on the input file ending with 'input.out'.  This is not a 'bad' think, but simply something that should be clearly documented somewhere for this code.

stsci-hack (approved):
Logic seems robust enough for me.

drizzlepac/hapsequencer.py
mdlpstsci: I understand. The pipeline files are currently defined in this manner which is what this PR is addressing, and there is the fallback for a generic naming convention which could be supplied by a user for their personal processing.

PR #1085, opened 2021-07-28
Create Drizzled MVM FLC files

stsci-hack:
These changes should result in drizzled versions of the MVM input FLC files to be produced as output products of MVM processing.

### DIFF
--- a/drizzlepac/hapmultisequencer.py
+++ b/drizzlepac/hapmultisequencer.py
@@ -117,6 +117,11 @@ def create_drizzle_products(total_obj_list):
         # Add individual single input images with updated WCS headers to manifest
         for exposure_obj in filt_obj.edp_list:
             product_list.append(exposure_obj.full_filename)
+            # Create Drizzled images for each input on SkyCell pixels
+            exposure_obj.wcs_drizzle_product(meta_wcs)
+            # Add drizzled FLC images to manifest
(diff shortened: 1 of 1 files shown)


drizzlepac/hapmultisequencer.py
```diff
@@ -117,6 +117,11 @@ def create_drizzle_products(total_obj_list):
         # Add individual single input images with updated WCS headers to manifest
         for exposure_obj in filt_obj.edp_list:
             product_list.append(exposure_obj.full_filename)
+            # Create Drizzled images for each input on SkyCell pixels
+            exposure_obj.wcs_drizzle_product(meta_wcs)
+            # Add drizzled FLC images to manifest
+            product_list.append(exposure_obj.drizzle_filename)
+
```
mdlpstsci, line 124: Will trailer files be produced for the skycell exposures?

mdlpstsci (commented):
Just need to verify if a trailer file will or will not be produced for the MVM visit.  If so, should it be retained as output?

drizzlepac/hapmultisequencer.py
stsci-hack: Good catch.  Added in d83bd00

mdlpstsci (approved):

PR #1087, opened 2021-07-29
Allow local files in poller files

stsci-hack:
This additional logic allows the creation of MVM poller files when the input files are already in the current working directory without having to specify an external environment variable. 

This change is NOT critical to any pipeline functionality, but rather allows poller files to be generated for local use; primarily, for testing.


mdlpstsci (approved):
Looks fine.

PR #1088, opened 2021-07-29
MVM: Skip any image/chip in the MVM visit list with no skycell overlap

mdlpstsci:
Modification to skip any image/chip in the MVM visit list which does not really overlap with the skycell footprint so the data is not included in the processing and cause a failure.


stsci-hack (approved):
Simple enough change.

PR #1091, opened 2021-07-31
Update HAPLEVEL for MVM products

stsci-hack:
Simple change to update the value of the HAPLEVEL keyword for MVM products (not exposure level MVM products) to a value of 4, from 3.


mdlpstsci (approved):
Agreed - simple!

PR #1095, opened 2021-08-03
Turn off CTX creation for MVM products

stsci-hack:
As requested by INS, this turns off the creation of CTX extensions for MVM products.


drizzlepac/haputils/product.py
```diff
@@ -868,6 +868,7 @@ def wcs_drizzle_product(self, meta_wcs):
         drizzle_pars["preserve"] = False
         drizzle_pars['rules_file'] = self.rules_file
         drizzle_pars['resetbits'] = "0"
+        drizzle_pars['context'] = False
```
mdlpstsci, line 871: Please add a comment here noting the reason for setting this parameter as you have done for the code below.

mdlpstsci (approved):

PR #1096, opened 2021-08-05
Update to using PHOTUTILS 1.1.0

stsci-hack:
These are the changes from #995 as applied to the code on master as of 4-Aug-2021.  In addition, these changes account for differences in the column names for columns in the SourceCatalog (previously, source_properties()) table that came with the update to v1.1.0.  

These changes were tested by processing dataset 'ic9t35' which contains both UVIS and IR data under Python 3.7, with Python 3.7 or greater being a requirement for installing photutils 1.1.0.  

**NOTE:  Python 3.7 or greater preferred, but will work with Python 3.6**


mdlpstsci (dismissed):
Changes to upgrade to Photoutils 1.1.0 look fine to me.

drizzlepac/buildmask.py
```diff
@@ -81,7 +81,7 @@ def buildDQMasks(imageObjectList,configObj):
 
 def buildMask(dqarr, bitvalue):
     """ Builds a bit-mask from an input DQ array and a bitvalue flag """
-    return bitfield_to_boolean_mask(dqarr, bitvalue, good_mask_value=1,
+    return bitfield_to_boolean_mask(dqarr.astype(np.int16), bitvalue, good_mask_value=1,
```
mcara, line 84: Why are you modifying `buildMask` in a PR that deals with PHOTUTILS. But rebase it agains the new `master` because `buildmask.py` now uses `astropy.nddata.bitmask` module.

mcara: Hm... no need to rebase - it looks like there are no conflicts.

mcara:
@stsci-hack @larrybradley @nden I wonder if it would be better not to hard-code in the `setup.py` `photutils` version requirement and instead continue to support various Python (i.e., 3.6 and above) and `photutils` versions?

stsci-hack:
These changes only prevent people from continuing to use Python 3.6, a version which is no longer even provided by conda.

larrybradley (approved):
Overall, this looks fine to me.  However, as I mentioned in https://github.com/spacetelescope/drizzlepac/pull/988, I am a bit worried about importing private functions from `photutils`. Because they are private, there is no guarantee that they will not suddenly change or disappear without deprecation.  I don't have any immediate or near-term plans to change these private functions, but I did have some ideas about improving this code (and perhaps making it public in the future).  In any case, the future-proof solution would be to copy those private functions here instead of calling them from `photutils`.

drizzlepac/haputils/deconvolve_utils.py
```diff
@@ -5,21 +5,15 @@
...
-if LooseVersion(photutils.__version__) < '1.1.0':
-    from photutils.detection.findstars import (_DAOFindProperties, _StarCutout,
-                                               _StarFinderKernel, _find_stars)
-else:
-    from photutils.detection._utils import (_StarCutout, _StarFinderKernel,
-                                            _find_stars)
-    from photutils.detection.daofinder import _DAOFindProperties
+from photutils.detection._utils import (_StarCutout, _StarFinderKernel,
+                                        _find_stars)
+from photutils.detection.daofinder import _DAOFindProperties
```
larrybradley, line 16: These are the three lines have the private imports from `photutils`.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -4,22 +4,30 @@
...
-from astropy.stats import sigma_clipped_stats, gaussian_fwhm_to_sigma
+from astropy.stats import sigma_clipped_stats
 from astropy.table import Column, MaskedColumn, Table, join, vstack
-from astropy.convolution import Gaussian2DKernel, RickerWavelet2DKernel
+from astropy.convolution import RickerWavelet2DKernel
 from astropy.coordinates import SkyCoord
 import numpy as np
 from scipy import ndimage, stats
 
+import photutils
```
mdlpstsci, line 17: Line 17 looks like cruft.

mdlpstsci (changes requested):
Just one small typo to clean up.

drizzlepac/haputils/catalog_utils.py
stsci-hack: Removed in dc66b81.

mcara (approved):
I haven't run any tests but it is great that unit tests pass and now this PR also supports older `photutils`. Thanks!

mcara:
Merge at will

mdlpstsci (approved):
Agreed - ready to merge.

PR #1097, opened 2021-08-06
Create code to generate custom mosaics that span multiple skycells

dulude:
**Relevant tickets and issues**
- [HLA-587](https://jira.stsci.edu/browse/HLA-587)

**High-level summary**
- Created make_custom_mosaic.py, a new script that facilitates the creation of custom MVM-mosaics from an arbitrary set of user-specified input images. The mosaic products produced can span multiple skycells (but not multiple projection cells) if necessary.
- In order to support this new functionality, a number of scripts were modified. These updates were verified to not impact the functionality  or output products of existing MVM code when executing normal single-skycell runs in any way.


drizzlepac/make_custom_mosaic.py
```diff
@@ -0,0 +1,428 @@
...
+        for ra_dec in proj_cell_dict[sc_name].wcs.calc_footprint().tolist():
+            x_y = proj_cell_dict[sc_name].projection_cell.wcs.all_world2pix([ra_dec], 0).tolist()[0]
+            ra_values[i] = ra_dec[0]
+            dec_values[i] = ra_dec[1]
+            x_values[i] = x_y[0]
+            y_values[i] = x_y[1]
+            i += 1
+
+    # Return mosaic bounding rectangle X min and max and Y min and max values
+    mosaic_limits = [x_values.min(), x_values.max(), y_values.min(), y_values.max()]
```
stsci-hack, line 224: Logic needs to be added to insure that these computed limits do not extend beyond the limits of the projection cell; namely, only go from 0 to NX and 0 to NY.

stsci-hack (commented):
Only 1 very minor point needs to be addressed.

drizzlepac/make_custom_mosaic.py
dulude: Issue resolved by commit [a977171](https://github.com/spacetelescope/drizzlepac/pull/1097/commits/a977171d520d1f46acf89f997e76998ae699e1b3)

stsci-hack (approved):
LGTM

PR #1101, opened 2021-08-12
Fix multiprocessing incompatibility with Python >= 3.8

### LINKED ISSUE #1068: Drizzlepac: Update to Python 3.8 multiprocessing library
_Issue [HLA-592](https://jira.stsci.edu/browse/HLA-592) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

The Drizzlepac package relies on Python's multiprocessing library to perform some drizzling in parallel.  This code needs to be modified to be compatible with the version of the library found in Python 3.8 and later.  
 
This ticket will focus on making the changes to the current Drizzlepac implementation to use the new multiprocessing library found in Python 3.8 and later.   

The dataset 'iacr51010' (specifically, iacr51010_asn.fits and all its member images) as used in the Drizzlepac test 'wfc3/test_wfc3/test_uvis_asn' will be suitable for testing that the multiprocessing code works correctly.  In fact, any association (or list of multiple input images) will serve as a test of this capability since the single drizzle step uses multiprocessing to create the individual (single drizzled) products which are used for CR identification.

mcara:
This PR fixes compatibility issues in `Python >= 3.8` when running `astrodrizzle` in parallel mode.

From [Python 3.8 Documentation](https://docs.python.org/3.8/library/multiprocessing.html#contexts-and-start-methods):

> Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess. See bpo-33725.

In order to pass open `HDULists` (or other complicated objects) to subprocesses, we have to continue using `'fork'` for now.

In the future, it may be worth exploring switching to `'spawn'` or `'forkserver'` which may improve security and possibly add support for multiprocessing when running on Windows.

Fixes #1068

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,8 @@ of the list).
 - Fix a memory corruption issue in ``interpolate_bilinear()`` in
   ``cdrizzleblot.c`` which could result in segfault. [#1048]
 
+- Fixed multiprocessing incompatibility with ``Python >= 3.8``. [#1101]
+
 In addition to a couple dozen bug fixes for the new SVM processing code
 to expand the number of visits which can be aligned to GAIA, the
 following significant changes to the code or some of the APIs have also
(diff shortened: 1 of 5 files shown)


stsci-hack (approved):
Straight-forward changes that work for all versions of Python >=3.6.  LGTM.

PR #1102, opened 2021-08-12
Use intrinsic int, bool, and float types instead of numpy

mcara:
This replaces the deprecated `numpy` global types

```python
np.float
np.bool
np.int
```

with the Python types

```python
float
bool
int
```
 
to avoid (reduce) deprecation warnings.

### DIFF
--- a/drizzlepac/devutils/alignment_viewer.py
+++ b/drizzlepac/devutils/alignment_viewer.py
@@ -286,5 +286,6 @@ def create_product_page(prodname, zoom_size=128, wcsname="",
 
 def get_col_val(hdrtab, keyword, default=None):
     val = hdrtab[0][keyword.upper()] if keyword.upper() in hdrtab.columns.names else default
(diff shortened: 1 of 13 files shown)


drizzlepac/devutils/alignment_viewer.py
```diff
@@ -286,5 +286,6 @@ def create_product_page(prodname, zoom_size=128, wcsname="",
 
 def get_col_val(hdrtab, keyword, default=None):
     val = hdrtab[0][keyword.upper()] if keyword.upper() in hdrtab.columns.names else default
-    if isinstance(val, bool) or isinstance(val, np.bool_): val = str(val)
+    if isinstance(val, bool) or isinstance(val, np.bool_):
```
pllim, line 289:
```suggestion
    if isinstance(val, (bool, np.bool_)):
```

pllim (dismissed):
Nice clean-up overall. One minor nitpick below.

PR #1103, opened 2021-08-12
Update dependency requirements. Remove unused dependencies

### LINKED ISSUE #1099: Remove unused dependencies
CC: @stsci-hack `setup.py` lists dependencies that are not used by the `drizzlepac`, such as `tables` and `PyYAML` (and maybe other packages) - see https://github.com/spacetelescope/drizzlepac/blob/67232d48b7736ddba64dbe9b4e4ab0c881acdfc5/setup.py#L106-L107

These should be removed.

mcara:
This PR updates dependency requirements. Now that `stsci.tools v4.0.0` has been released, we can undo upper limit requirement for the `numpy` introduced in https://github.com/spacetelescope/drizzlepac/pull/1051/commits/898c1e1510cc3106dfe2c6996b4eeefcd8c7f8c1

This PR restores `numpy` requirement back to the original `numpy>=1.18`.

Also this PR removes apparently unused dependencies: `tables` and `PyYAML`. Closes #1099


stsci-hack (approved):
This makes a lot of sense to me.

PR #1104, opened 2021-08-17
Initial version of MVM tools: find files in the region of interest,

mdlpstsci:
VERY MUCH WIP.  Create utilities to: (1) find files in the region of interest, (2) download the files, and (3) make cutouts from the files.

### DIFF
--- a/drizzlepac/haputils/hapcut_utils.py
+++ b/drizzlepac/haputils/hapcut_utils.py
@@ -3,14 +3,16 @@
 
 from astrocut import fits_cut
 from astropy import units as u
-from astropy.units.quantity import Quantity
 from astropy.coordinates import SkyCoord
-from drizzlepac.haputils import cell_utils as cu
-from astroquery.mast import Observations
+from astropy.io import fits
 from astropy.table import Table, vstack
+from astropy.units.quantity import Quantity
+from astroquery.mast import Observations
+from drizzlepac.haputils import cell_utils as cu
 from stsci.tools import logutil
-import numpy as np
+
 import math
+import numpy as np
 import os
 import shutil
 import sys
@@ -109,8 +111,7 @@ def mvm_id_filenames(sky_coord, cutout_size, verbose=False):
     good_rows = []
     updated_table = None
     for i, old_row in enumerate(query_table):
-        #if old_row["obs_id"].startswith("hst_skycell"):
-        if old_row["obs_id"].startswith("hst_"):
+        if old_row["obs_id"].startswith("hst_skycell"):
             if old_row["s_ra"] >= ra_min and old_row["s_ra"] <= ra_max and \
                old_row["s_dec"] >= dec_min and old_row["s_dec"] <= dec_max:
                 good_rows.append(old_row)
@@ -232,8 +233,8 @@ def mvm_retrieve_files(products, archive=False, clobber=False, verbose=False):
 
 def make_the_cut(input_files, sky_coord, cutout_size, single_outfile=True, output_dir=".", verbose=False):
     """
-    This function makes the actual cut in the input images. As such it isa a high-level
-    interface for the `˜astrocut.cutout.fits_cut` functionality.
+    This function makes the actual cut in the input MVM drizzled filter-level images. As such it is
+    a high-level interface for the `˜astrocut.cutout.fits_cut` functionality.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,324 @@
...
+                     "_{:.0f}_cutout".format(radius.value) + ".ecsv" 
+
+    if verbose:
+        print("Writing out the filter product list table to {}.".format(query_filename))
+    filtered_dp_table.write(query_filename, format="ascii.ecsv")
+
+    return filtered_dp_table
+
+
+def mvm_retrieve_files(products, archive=False, clobber=False, verbose=False):
```
stsci-hack, line 157: The name of this function implies that it will actually get the images themselves (files) as opposed to only reporting the filenames back to the user.  'mvm_retrieve_filenames' would be clearer.

mdlpstsci: Unless the file is already found on disk, this function does retrieve the files.

stsci-hack: Please update the docstring to make this behavior (more?) apparent.

mdlpstsci: Fixed in 6620833.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,383 @@
...
+    # collection (HST).  Use the wildcard to get all the detectors for the instrument.  Multiple
+    # instruments cannot be searched at the same time.  Use the diagonal of the cutout to define
+    # the search radius for the archive.  Images which fall outside the desired cutout need to
+    # be filtered from the solution later.
+    radius = math.ceil(math.sqrt(math.pow(cutout_size.value[0], 2) + math.pow(cutout_size.value[1], 2)) / 2.0)
+
+    # Careful - the radius must be a str or Quantity
+    radius *= u.arcsec
+    if verbose:
+        print("Performing query for ACS images. Radius: {}.".format(radius))
```
stsci-hack, line 78: The logging has been defined at the top of the module.  Subsequent print statements should at least be accompanied by the relevant 'log.info' or 'log.warning' or 'log.debug' statements.  Alternatively, the print statements could be replaced entirely by the log statements with the log level set based on the value of verbose as done in other modules.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,383 @@
...
+    ra_max = sky_coord.ra.value + deg_cutout_size.value[0]
+    dec_min = sky_coord.dec.value - deg_cutout_size.value[1]
+    dec_max = sky_coord.dec.value + deg_cutout_size.value[1]
+
+    # Filter the output as necessary to include only MVM filenames (MVM prefix: hst_skycell).
+    # Also, filter out images which are not actually in the requested cutout region as the
+    # archive search had to be done using a radius.
+    good_rows = []
+    updated_table = None
+    for i, old_row in enumerate(query_table):
```
stsci-hack, line 113: Does 'i' get used at all (I couldn't see where it did)?  If not, you could remove the use of 'enumerate' just to keep the code a little simpler.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,383 @@
...
+        if old_row["obs_id"].startswith("hst_skycell"):
+            if old_row["s_ra"] >= ra_min and old_row["s_ra"] <= ra_max and \
+               old_row["s_dec"] >= dec_min and old_row["s_dec"] <= dec_max:
+                good_rows.append(old_row)
+    updated_query_table = Table(rows=good_rows, names=query_table.colnames)
+
+    # Catch the case where no files are found which satisfy the criteria
+    if not updated_query_table:
+        print("WARNING: Query for objects within {} of {} returned NO RESULTS!".format(radius, sky_coord))
```
stsci-hack, line 122: This print statement could be replaced by 'log.warning'.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,383 @@
...
+    # Filter on MVM drizzled product
+    suffix = ["DRZ", "DRC"]
+    if verbose:
+        print("Filter the product list table for only {} filenames.".format(suffix))
+    filtered_dp_table = Observations.filter_products(dp_table,
+                                                     productSubGroupDescription=suffix,
+                                                     extension="fits")
+
+    if not filtered_dp_table:
+        print("WARNING: No drizzle product files (DRZ/DRC) found.")
```
stsci-hack, line 139: Another place where 'log.warning' would be appropriate.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,383 @@
...
+        return filtered_dp_table
+
+    # Write the filtered data product table out to a file.  This allows for further
+    # manipulation of the information before a list of filenames is distilled from
+    # the table.
+    # Output filename in the form: mvm_query_ra.dddd_sdec.dddd_radius_cutout.ecsv.
+    #                              mvm_query_84.9208_s69.1483_71_cutout.ecsv
+    ns = "s" if sky_coord.dec.degree < 0.0 else "n"
+    query_filename = "mvm_query_" + str(sky_coord.ra.degree) + "_" + ns + str(abs(sky_coord.dec.degree)) + \
```
stsci-hack, line 148: The number of digits (after the decimal, of course) reported in the RA and Dec can become excessive.  Is it safe to blindly use all digits of RA/Dec for the filename, or should the RA/Dec values be limited to a set number of digits, perhaps 6 or 8.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,383 @@
...
+    all_images = products['productFilename'].tolist()
+    if not clobber:
+        rows_to_remove = []
+        for row_idx, row in enumerate(products):
+            fname = row['productFilename']
+            if os.path.isfile(fname):
+                print(fname + " already exists. File download skipped.")
+                rows_to_remove.append(row_idx)
+        products.remove_rows(rows_to_remove)
+
```
stsci-hack, line 197: What happens if you remove all the rows at this point; namely, the user already had all the images they want locally and just wasn't sure or merely following a prescribed flow without understanding it?

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -0,0 +1,383 @@
...
+    verbose : bool
+        Default False. If True, additional intermediate information is printed.
+
+    Returns
+    -------
+    response : str or list
+        If single_outfile is True returns the single output filepath. Otherwise returns a list of all 
+        the output filepaths.
+        If memory_only is True a list of `~astropy.io.fit.HDUList` objects is returned instead of
```
stsci-hack, line 277: Where is the 'memory_only' parameter set?

stsci-hack (changes requested):
Overall, this looks really logical and efficient.  There were just a couple of points which stood out to me that should be addressed one way or another before merging this code.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: Residual cruft has been removed.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: This parameter is not exposed in the public interface, and this value is hard-coded in make_the_cut() to True.  By setting the value to True when invoking the low-level software, HDUList objects are returned to the high-level software.  This allows the high-level software to modify the FITS file headers before the data is written out.  We had decided we always wanted to write out a file.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: Ooops.  Clearly this is not a good idea. I am addressing this by allowing 6 digits in the decimal portion of the coordinates.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: Added a check and log messages to let the user know the if files are downloading or all the files are already resident on disk.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: Use logging as is done for the HAP tasks.  The "verbose" parameter only applies to the underlying "astrocut" functionality.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: Addressed.

drizzlepac/haputils/hapcut_utils.py
stsci-hack: This can confuse an end-user if they are being told about a parameter which they do not have access to modify.  The phrase 'If memory_only is True' could/should be removed to avoid this confusion.

stsci-hack (approved):
This looks good now.

PR #1106, opened 2021-08-20
Compute mvm sregion

stsci-hack:
The S_REGION keyword computation has been simplified and generalized to not only work for pipeline and SVM products, but also for MVM products.  

The simplification comes from correctly implementing use of the Harris corner detection algorithm (from skimage) to detect the corners in the mask of SCI array while removing the computation based on each input to the output product.  This allows the code to work on images where chips extend beyond the bounds of the output image, as is the case with MVM products.  

The output value of the S_REGION keyword also now separates each region with it's own POLYGON ICRS tag, leading to multiple POLYGON sets of coordinates in the keyword value.  

In addition, a new method was added to the SkyFootprint class to enable creation of a footprint mask directly from the SCI extension of an exposure without having to read in any other images (like all the input exposures for an SVM or MVM product).   

This computation of the S_REGION keyword was used to successfully determine the corners for a variety of exposures, namely:

- **hst_skycell-p0212x04y03_acs_wfc_f658n_all** : MVM product with 5 separate regions where ACS exposures partially land in the output image
- **j8mbnl** : single pointing ACS/WFC SVM product with 2 separate regions (1 for each chip)
- **ib3m53**:  WFC3/IR mosaic with complex pointing creating 1 large irregularly shaped footprint/region
- **ibvg03** : WFC3/IR mosaic with  complex pointing creating 1 large irregularly shaped footprint/region
- **ib1f85** : WFC3/IR single pointing product complete with hole
In each SVM case, the number of corners was comparable to or less than the number of corners previously computed. 

**NOTE:**
This PR should not be merged until we get confirmation from ASB that this form of the S_REGION keyword value can be correctly interpreted by the archive's portal.

### DIFF
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -376,9 +376,19 @@ class SkyFootprint(object):
 
     def extract_mask(self, filename):
         """Extract a total_mask from the SCI data directly"""
-        sciext = ("SCI", 1)
-        # Try to use the WHT array as a msk
-        arr = fits.getdata(filename, ext=sciext)
+        # Determine what extension contains the SCI array
+        # It could be 0 if 'build=no' for drizzling
+        fhdu = fits.open(filename)
+        if len(fhdu) > 1:
+            sciext = ("SCI", 1)
+        else:
+            sciext = 0
+        # Get the SCI array to use as a mask
+        arr = fhdu[sciext].data.copy()
+        # Done with image, so close immediately.
+        fhdu.close()
+        del fhdu
+
         # If working with drizzled data which has NaN as non-exposed pixel values
         if np.isnan(arr.min()):
             total_mask = (np.isnan(arr) == 0).astype(np.int16)
@@ -536,7 +546,7 @@ class SkyFootprint(object):
                 # insure there is a border all around the region
                 # THIS IS CRITICAL to being able to identify corners correctly in slice
                 label_mask = ndimage.binary_erosion(label_mask)
-                print('extracting corners for label {} in slice {}'.format(label, mask_slice))
+                print('extracting corners for region {} in slice {}'.format(label, mask_slice))
                 # Perform corner detection on each region/chip separately.
                 mask_corners = corner_peaks(corner_harris(label_mask),
                                        min_distance=1,
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -381,14 +381,14 @@ class SkyFootprint(object):
         arr = fits.getdata(filename, ext=sciext)
         # If working with drizzled data which has NaN as non-exposed pixel values
         if np.isnan(arr.min()):
-            self.total_mask = (np.isnan(arr) == 0).astype(np.int16)
+            total_mask = (np.isnan(arr) == 0).astype(np.int16)
         else:
             total_mask = (arr != 0).astype(np.int16)
-            # Remove 'small' holes in the image due to noise to avoid
-            # creating extraneous 'regions' from the image when it is really
-            # all one region/chip.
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/cell_utils.py
```diff
@@ -480,111 +507,75 @@ def find_corners(self, member='total', sensitivity=0.1):
...
-                # Start matching the xy_corner positions, one-by-one, to pixels
-                # along the edge.  We only want the index of the corner that matches
-                # as we travel along the edge in order which will be used to
-                # sort the corner positions.
+            # insure footprint has enough signal to detect corners
+            fp = np.clip(self.footprint, 0, 1).astype(np.int16)
+
+            # simple trick to remove noise and small regions 3x3 or less.
+            scmask = ndimage.binary_dilation(ndimage.binary_erosion(fp, iterations=3), iterations=2)
+            # Label each major contiguous region in the mask
```
mdlpstsci, line 517: Cool trick.

drizzlepac/haputils/processing_utils.py
```diff
@@ -240,19 +247,21 @@ def find_footprint(hdu, extname='SCI'):
 
     """
     # Extract list of input files from Drizzle keywords
-    data_kws = hdu[0].header['d*data'] if isinstance(hdu, fits.HDUList) else fits.getval(hdu, 'd*data')
-    input_files = [kw.split('[')[0] for kw in data_kws.values()]
+    # data_kws = hdu[0].header['d*data'] if isinstance(hdu, fits.HDUList) else fits.getval(hdu, 'd*data')
+    # input_files = [kw.split('[')[0] for kw in data_kws.values()]
```
mdlpstsci, line 252: Did you mean for lines 250, 251, and 258 to be removed or ?

mdlpstsci (commented):
I really only have two questions as understanding of the algorithm itself requires significant knowledge and experience with numpy, ndimage, and masking (at the least) to follow the logic.  It is clear the author has worked in this realm previously.  I inquire in the code about some lines which are commented out.  Also, I am wondering if the compute_sregion() function is needed by the skycell product classes in product.py.

drizzlepac/haputils/product.py
```diff
@@ -663,6 +667,9 @@ def wcs_drizzle_product(self, meta_wcs):
                                   output=self.drizzle_filename,
                                   **drizzle_pars)
 
+        # Compute S_REGION keyword and add it to the header of this product
+        putils.compute_sregion(self.drizzle_filename)
+
         # Rename Astrodrizzle log file as a trailer file
         log.debug("Exposure image {}".format(self.drizzle_filename))
         try:
```
mdlpstsci, line 675: Do the calls for s_region need to be in the SkycellExposure and SkycellProduct classes.?

drizzlepac/haputils/cell_utils.py
```diff
@@ -372,6 +374,31 @@ def build(self, expnames, scale=False, scale_kw='EXPTIME'):
         # Compute the bounded WCS for this mask of exposed pixels
         self.find_bounded_wcs()
 
+    def extract_mask(self, filename):
+        """Extract a total_mask from the SCI data directly"""
+        sciext = ("SCI", 1)
+        # Try to use the WHT array as a msk
```
mdlpstsci, line 380: Correct comment regarding use of WHT array.

mdlpstsci (changes requested):
One very minor change to update a comment, and this PR is ready to go.

drizzlepac/haputils/cell_utils.py
stsci-hack: Not only did I fix this comment in 8592ee9, I updated the logic to work with simple FITS files as well, since DRZ files could be simple when 'build=no' in astrodrizzle.

mdlpstsci (approved):

PR #1112, opened 2021-08-31
Support general input poller file name

stsci-hack:
Update SVM and MVM processing code to support input filenames with any extension, as opposed to throwing an Exception if the filename did not end in '.out'.


mdlpstsci (approved):
Straightforward and gets the job done!

stsci-hack:
Closed #1113

PR #1115, opened 2021-09-01
SVM: Handle multiple total data products (TDP) when first TDP processed has no sources

mdlpstsci:
Ensure the segmentation image is not None upon return from detect_and_eval_segments()and handle the situation appropriately though additional logic.  Added informational log messages regarding lack of sources found and catalog status. Protect against divide by zero when computing the ratio of "Gaussian big islands" vs "RW big islands".

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1617,7 +1617,7 @@ class HAPSegmentCatalog(HAPCatalogBase):
 
                 # No segments were detected in the total data product - no further processing done for this TDP,
                 # but processing of another TDP should proceed.
-                else not rw_segm_img:
+                elif not rw_segm_img:
                     return
 
             # The first round custom/Gaussian segmentation image is good, continue with the processing
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1477,7 +1477,7 @@ class HAPSegmentCatalog(HAPCatalogBase):
 
                 # Compute the ratio of big sources/islands using Custom/Gaussian kernel vs Rickerwavelet kernel
                 # This value can be used as a discriminant between overlapping point sources and nebulousity fields
-                ratio_cg2rw = 0.0
+                ratio_cg2rw_bigsource = 3.0
                 if rw_bs > 0.0:
                     ratio_cg2rw_bigsource = g_bs / rw_bs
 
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -1479,10 +1477,11 @@ def identify_sources(self, **pars):
 
                 # Compute the ratio of big sources/islands using Custom/Gaussian kernel vs Rickerwavelet kernel
                 # This value can be used as a discriminant between overlapping point sources and nebulousity fields
-                ratio_cg2rw_bigsource = g_bs / rw_bs
+                if rw_bs > 0.0:
+                    ratio_cg2rw_bigsource = g_bs / rw_bs
```
stsci-hack, line 1481: If 'rw_bs' == 0, the variable 'ratio_cg2rw_bigsource' never gets initialized with a value, yet it gets used later (line 1586).  On the face of it, this would cause an exception unless having 'rw_bs' == 0 somehow prevents the logic on line 1586 from ever being called.  Please clarify to insure no Exceptions get thrown.

stsci-hack (changes requested):
Overall looks fine.  Just one bit of logic which needs to be clarified before approval.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: Oops.  Fixed in 9a8b0d5.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -1479,10 +1477,12 @@ def identify_sources(self, **pars):
 
                 # Compute the ratio of big sources/islands using Custom/Gaussian kernel vs Rickerwavelet kernel
                 # This value can be used as a discriminant between overlapping point sources and nebulousity fields
-                ratio_cg2rw_bigsource = g_bs / rw_bs
+                ratio_cg2rw = 0.0
```
stsci-hack, line 1480: Should this be 'ratio_cg2rw_bigsource' instead?

stsci-hack (commented):
It is not clear that the variable was initialized correctly... please confirm.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -1746,15 +1746,15 @@ def detect_and_eval_segments(self, imgarr, kernel, ncount, size_source_box, nsig
             if segm_img is None:
                 log.warning("End processing for the segmentation catalog due to no sources detected with the current kernel.")
                 log.warning("No segmentation catalog will be produced for this total detection product, {}.".format(self.imgname))
-                is_big_crowded = False
-                big_island = 100.0
-                source_fraction = 100.0
+                is_big_crowded = True
+                big_island = 1.0
+                source_fraction = 1.0
```
stsci-hack, line 1751: The logic here for 'seg_img is None' duplicates the logic at the start of the method '_evaluate_segmentation_image'.  Removing this entire if...else block and adding these log messages to the method's own 'if seg_img is None' block of code would simplify the code by putting this check in only one place.

stsci-hack (approved):
My comment doesn't require another revision, but is something to consider for later.  Otherwise, I think this PR is ready to merge.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: Fixed in ada6517.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: Does not hurt anything.

PR #1119, opened 2021-09-03
WIP: Ensure hapcut utilities will work on MVM exposure-level products

mdlpstsci:
Removed the option of a single output MEF file for image cutouts.
The output cutout filename now contains "p####" vs "####" for consistency with
other MVM filenames.
Ensure proper treatment of WFC3/IR files which have both a "fine" (default) and
"coarse" platescales where only "coarse" is part of the input and output filenames.
Instead of adding the input filename to the PHDU header of the output file as a
HISTORY keyword, put this value in keyword ORIG_FLE in the PHDU.  This keyword
can already be found in the EHDUs.
Delete stale variables asap.


drizzlepac/haputils/hapcut_utils.py
```diff
@@ -492,32 +485,89 @@ def mvm_combine(cutout_files, img_combiner=None, output_dir=".", log_level=logut
...
+    # Walk the sorted input list and create a group list for each new detector/filter
+    temp_list = []
+    detector_filter_list = []
+    prev_detector = ""
+    prev_filter = ""
+    for cfile in cutout_files:
+        cf = cfile.replace(".fits", "")
+        detector = cf.split("_")[3]
+        filter = cf.split("_")[4].split("-")[0]
+        if detector != prev_detector:
```
stsci-hack, line 519: This logic lumps both filter and exposure level input files into 1 group, making it more difficult to combine just one or the other later, unless I am overlooking something...

stsci-hack (commented):
The logic seems to lump both filter and exposure inputs into 1 combined product, whereas the filter inputs already include the exposure files.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: This has been addressed in commit 2343efc.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -437,50 +464,133 @@ def make_the_cut(input_files, sky_coord, cutout_size, single_outfile=True, outpu
...
+    """
 
+    img_combiner = None
 
-def mvm_combine(cutout_files, img_combiner=None, output_dir=".", log_level=logutil.logging.INFO):
+    # set logging level to user-specified level
+    log.setLevel(log_level)
+
+    # Make sure the cutout_files are really a list of MULTIPLE filenames
+    if type(cutout_files) == str or type(cutout_files) == list and len(cutout_files) < 2:
```
stsci-hack, line 508: The next iteration of this function could support a filename as input, where the file provides the list of exposures to be combined.

drizzlepac/haputils/hapcut_utils.py
```diff
@@ -492,32 +602,50 @@ def mvm_combine(cutout_files, img_combiner=None, output_dir=".", log_level=logut
...
+    OUTPUT_PREFIX = "hst_combined_skycells-"
+
+    for key, file_list in input_dict.items():
+
+        # If there are multiple files to combine, then do it
+        if len(file_list) > 1:
+
+            # Construct the combined filename based on the first file in the list
+            # Example: hst_combined_skycells-ra84d9207-decs69d8516_uvis_f275w.fits
+            filename = fits.getheader(file_list[0], ext=0)['FILENAME']
```
stsci-hack, line 620: Please confirm the format of the input filenames.  I would expect that the input files to be MVM product filenames of the format `hst_skycell-p0081x14y14_acs_wfc_f658n-all_drc.fits`.  Perhaps I misunderstand how this code was supposed to be called or when it was supposed to be used.

stsci-hack (approved):
I did not see anything unexpected, and the logic looks fine to me now.

drizzlepac/haputils/hapcut_utils.py
mdlpstsci: OK - will consider this.  This function was meant to accept as input the output from the make_the_cut routine which is a list of filenames.

stsci-hack (approved):

PR #1125, opened 2021-09-10
MVM: Insure SVM-derived WCS gets used as input to MVM processing

dulude:
**Related JIRA Ticket**
- [HLA-598](https://jira.stsci.edu/browse/HLA-598)

**High-level Summary**
- Modified MVM pipeline code so that it can process both non-pipeline-processed flc/flt fits files (as it did previously) and SVM pipeline-processed flc/flt fits files as well. This gives the MVM pipeline the ability to utilize the WCS solutions that were previously derived by the SVM pipeline.   The vast majority of the changes were to account for the differences in the filename formatting between the SVM pipeline-processed  and the non-SVM pipeline-processed input flc/flt fits files. 

**Additional Code Additions**
- Added new script which_skycell.py. This script is basically just a command-line interface for cell_utils.get_sky_cells() and as the name of the script suggests, allows users to determine which skycell(s) a given set of observations occupy.

### DIFF
--- a/drizzlepac/haputils/which_skycell.py
+++ b/drizzlepac/haputils/which_skycell.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+"""Returns the name(s) of the projection cell(s)/skycell(s) that the observations in the user-specified input
+ file occupy."""
+
+import argparse
+import glob
+import os
+import pdb
+import sys
+
+from drizzlepac.haputils import cell_utils
+
+
+__version__ = 0.1
+__version_date__ = '08-Sept-2021'
+# ------------------------------------------------------------------------------------------------------------
+
+def identify_pc_sc(img_list):
+    """identifies the projection cell(s) and skycell(s) that the observations in the user-specified input
+    list occupy.
+
+    Parameters
+    ----------
+    img_list : list
+        list of the images to process
+
+    Returns
+    -------
+    Nothing!
+    """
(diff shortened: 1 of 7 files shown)


drizzlepac/haputils/which_skycell.py
```diff
@@ -0,0 +1,61 @@
...
+
+from drizzlepac.haputils import cell_utils
+
+
+__version__ = 0.1
+__version_date__ = '08-Sept-2021'
+# ------------------------------------------------------------------------------------------------------------
+
+
+def identify_pc_sc(img_list):
```
stsci-hack, line 18: This function name seems a little...obscure.  Changing this to something a little more explicit would help the user understand what it can do; perhaps, 'report_sky_cells', where 'report' indicates it is telling the user the information.

stsci-hack (changes requested):
Just one comment which should be addressed before merging this code.  Otherwise, the rest of it appears to be just fine to me.

drizzlepac/haputils/which_skycell.py
dulude: Issue resolved by commit [27fa35f](https://github.com/spacetelescope/drizzlepac/pull/1125/commits/27fa35ff65fb597a10a1c0d751752d5e7bd6e40c).

stsci-hack (approved):
LGTM

PR #1127, opened 2021-09-14
Remove dependence on photutils private functions

larrybradley:
This PR will remove once and for all the current dependence on `photutils` private functions.  I'm planning to refactor some of these private functions in `photutils` to allow the addition of new features (prompted by https://github.com/astropy/photutils/pull/1231).  Moving the current versions of the private functions here will prevent `drizzlepac` from breaking after these private functions are changed in `photutils`.  I don't want to break `drizzlepac`.

This would need to go in the 3.3.0 release, otherwise I'm stuck in `photutils` without breaking `drizzlepac`.

### DIFF
--- a/drizzlepac/haputils/_detection_utils.py
+++ b/drizzlepac/haputils/_detection_utils.py
@@ -1,0 +1,647 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+This module implements common utility functions and classes for the star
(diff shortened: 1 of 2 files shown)


larrybradley:
CC @stsci-hack

stsci-hack (approved):
Simple way to internalize these private functions from photutils.  LGTM.

PR #1128, opened 2021-09-14
which_skycell.py simple bugfix

dulude:
This is just a simple one-line bugfix. which_skycell.py had an incorrectly set argument that I didn't catch before merging PR #1125.


stsci-hack (approved):
Easy thing to miss the first time.  LGTM

PR #1132, opened 2021-09-21
Function mvm_analyze_wrapper determines the viability of an image for MVM processing

mdlpstsci:
Added a new function, mvm_analyze_wrapper, which uses the functionality in the underlying
analyze_data routine to examine FITS keywords and determine the viability of the data to
be used in the MVM processing.  The mvm_analyze_wrapper works on a single image at a time.
Added an optional parameter to analyze_data so the routine can determine when MVM processing
is wanted as MVM processing does not allow anything to be done for GRISM/PRISM images.
Updated doc strings.

### DIFF
--- a/drizzlepac/haputils/analyze.py
+++ b/drizzlepac/haputils/analyze.py
@@ -25,7 +25,7 @@ __taskname__ = 'analyze'
 
 MSG_DATEFMT = '%Y%j%H%M%S'
 SPLUNK_MSG_FORMAT = '%(asctime)s %(levelname)s src=%(name)s- %(message)s'
-log = logutil.create_logger(__name__, level=logutil.logging.NOTSET, stream=sys.stdout,
+log = logutil.create_logger(__name__, level=logutil.logging.DEBUG, stream=sys.stdout,
                             format=SPLUNK_MSG_FORMAT, datefmt=MSG_DATEFMT)
 
 __all__ = ['analyze_data', 'analyze_wrapper', 'mvm_analyze_wrapper']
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/analyze.py
```diff
@@ -104,6 +142,9 @@ def analyze_data(input_file_list, log_level=logutil.logging.NOTSET):
         The desired level of verboseness in the log statements displayed on the screen and written to the .log file.
         Default value is 20, or 'info'.
 
+    type : string
```
stsci-hack, line 145: It might be helpful for us later (if no one else) to include the list of supported possible values for this string here.

drizzlepac/haputils/analyze.py
```diff
@@ -303,7 +344,7 @@ def analyze_data(input_file_list, log_level=logutil.logging.NOTSET):
         # for all the good data.
         split_sfilter = sfilter.upper().split('_')
         for item in split_sfilter:
-            if item.startswith(('G', 'PR')) and not is_zero:
+            if item.startswith(('G', 'PR')) and not is_zero and type.upper() != "MVM":
```
stsci-hack, line 347: It is not clear to me what message would be generated for Grism or Prism data, if any, as it seems that if `type == MVM`, the image would not be flagged to be skipped at all with this change.  This change would leave, if I understand correctly, Grism/Prism data with `no_proc_key`= None, or in other words, it would still be processed.

stsci-hack (commented):
I could not follow the logic entirely and had a question about how Grism/Prism data are handled that should be clarified before approval.

drizzlepac/haputils/analyze.py
mdlpstsci: Fixed in commit 111c243.

drizzlepac/haputils/analyze.py
mdlpstsci: The logic here is correct, but I make the if-elif more elaborate in order to accommodate additional log messages describing the action.  Also add more comments in commit 111c243.

stsci-hack: The extra log messages are perfect!

stsci-hack (approved):

PR #1136, opened 2021-09-23
Add FWHM measurement to the output Segmentation filter catalogs

mdlpstsci:
Added the FWHM measurement to the output Segmentation filter catalog. The FWHM measurement
only exists in versions of Photutils > 1.1.0.  The code accommodates older versions of
Photutils by including the FWHM column in the output Segmentation catalog, but the entire
column is zero-filled.


stsci-hack (approved):
This seems to cover everything nicely in adding this column and making it backwards-compatible.

PR #1138, opened 2021-09-24
Improve make_poller_files.py

dulude:
**Relevant Ticket**
-  [HLA-572](https://jira.stsci.edu/browse/HLA-572)

**Summary of changes**
This PR addresses issues raised by @mdlpstsci  regarding a couple of UX bugs in make_poller_files.py. Fixes/improvements include the following:
- Added additional logic to throw an exception if users don't specify a skycell name when creating a MVM poller file
- Added additional logic to throw an exception if users don't specify a properly formatted skycell name when creating a MVM poller file
- Refactored the fits file location subroutine to make it more flexible and the logic more robust.
- Assorted docstring improvements and updates

### DIFF
--- a/drizzlepac/haputils/make_poller_files.py
+++ b/drizzlepac/haputils/make_poller_files.py
@@ -61,7 +61,7 @@ def generate_poller_file(input_list, poller_file_type='svm', output_poller_filen
         else:
             # Warn user if no fits file can be located for a given rootname, and skip processing of the file.
             if rootname.endswith(".fits"):
-               item_type = "filename"
+                item_type = "filename"
             else:
                 item_type = "rootname"
             print("WARNING: No fits file found for {} '{}'. This {} will be omitted from the poller file.".format(item_type, rootname, item_type))
--- a/drizzlepac/haputils/make_poller_files.py
+++ b/drizzlepac/haputils/make_poller_files.py
@@ -5,6 +5,7 @@ user-specified list file."""
 
 import argparse
 import os
+import re
 import sys
 
 from astropy.io import fits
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/make_poller_files.py
```diff
@@ -117,35 +133,39 @@ def locate_fitsfile(search_string):
          full file path + image name of specified search_string.
     """
     if search_string.endswith("_flt.fits") or search_string.endswith("_flc.fits"):  # Process search_string as a full filename
-        # Look for files in CWD first
-        if os.path.exists(search_string):
+        # Look in user-provided path (assuming they provided one)
+        if os.path.exists(search_string) and "/" in search_string:
```
stsci-hack, line 137: The use of the hard-coded value '/' here should be replaced instead with 'os.sep' in order to make this code more platform-independent.

drizzlepac/haputils/make_poller_files.py
```diff
@@ -117,35 +133,39 @@ def locate_fitsfile(search_string):
...
+            return search_string
+        # Look for files in CWD
+        if os.path.exists(search_string) and "/" not in search_string:
             return os.getcwd()+"/"+search_string
         # If not found in CWD, look elsewhere...
         if not os.getenv("DATA_PATH"):
             sys.exit("ERROR: Undefined online cache data root path. Please set environment variable 'DATA_PATH'")
-        fullfilepath = "{}/{}/{}/{}".format(os.getenv("DATA_PATH"), search_string[:4], search_string, search_string)
+        fullfilepath = "{}/{}/{}/{}".format(os.getenv("DATA_PATH"), search_string[:4], search_string[:-9], search_string)
```
stsci-hack, line 145: Again, a more platform-independent way to generate filenames would be to use 'os.sep'..

stsci-hack (changes requested):
Please take a look at making this code work on all Python platforms as noted in my comments.

drizzlepac/haputils/make_poller_files.py
dulude: issue resolved by commit [6626a45](https://github.com/spacetelescope/drizzlepac/pull/1138/commits/6626a452591a0f7331157f0954be141be06cfa4f)

drizzlepac/haputils/make_poller_files.py
dulude: issue resolved by commit [6626a45](https://github.com/spacetelescope/drizzlepac/pull/1138/commits/6626a452591a0f7331157f0954be141be06cfa4f)

stsci-hack (approved):
LGTM

PR #1142, opened 2021-09-27
Add the Kron radius as an output measurement for the catalog

mdlpstsci:
Added the Kron radius as an output measurement for the catalog.  This change required providing
"kron_params" to the SourceCatalog instantiation.  The "kron_params" are a list of two values that
are used to determine how the Kron radius and flux are calculated.  The two values, kron_scaling_radius
and kron_minimum_radius, are stored as two new configuration values in the catalog configuration file
associated with each of the detectors.


stsci-hack (approved):
Looks like all the bases are covered with this simple fix.

PR #1146, opened 2021-09-28
Fixed a bug which caused the proper naming of the output manifest

mdlpstsci:
Fixed a bug which caused the proper naming of the output manifest file to be botched.  The manifest filename is now properly constructed if the main input is a poller file (no specific filename format expected), a user-defined filename
containing a list of files, or a Python list.  In the latter two cases, the manifest filename is constructed by interrogating
the primary header of the first file for information.  In the poller file case, the total data product has a manifest filename attribute which is used.  If something goes completely amiss, the manifest filename will be "manifest.txt".  In all other cases the manifest filename will be of the form instrument_programID_obsetID_manifest.txt (e.g.,wfc3_b46_06_manifest.txt).

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -47,6 +47,7 @@
 """
 import datetime
 import fnmatch
+import glob
 import logging
 import os
 import pickle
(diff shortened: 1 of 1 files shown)


stsci-hack (approved):
Nice change to the logic.

PR #1148, opened 2021-10-05
Improvements/updates made to MVM utilities while documenting the code

mdlpstsci:
For mvm_id_filenames, make the output MAST query results file have a
consistent naming convention with the remaining hapcut utilities.
For make_the_cut, loop over the files for which cutouts have been requested.
In this way if one of the files causes an exception for any reason (e.g.,
the file does not contain the cutout region), an error is issued, but
the remaining files continue to process. For make_the_cut and mvm_combine,
added the instrument designation to the output files for completeness (though
now the filenames are even longer).  For mvm_combine, added the FILENAME
keyword to the PHDU of the output combined file.  Also, improved the EXTNAMEs
of all the EHDUs to reflect the data they contain.  Added or improved clarity
of comments.

### DIFF
--- a/drizzlepac/haputils/hapcut_utils.py
+++ b/drizzlepac/haputils/hapcut_utils.py
@@ -316,7 +316,7 @@ def make_the_cut(input_files, sky_coord, cutout_size, output_dir=".", log_level=
         must be in angular units.
 
     output_dir : str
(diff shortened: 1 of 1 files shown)


stsci-hack (approved):
LGTM

PR #1149, opened 2021-10-05
Update v3.3.0 user docs

stsci-hack:
The readthedocs documentation explaining the changes in (mostly) user-friendly terms, as opposed to just describing the API, has been updated with descriptions of most of the significant changes made under v3.3.0.


stsci-hack:
The changes made to align_utils are obviously not documentation updates, but got included accidentally and was more trouble than it was worth to remove from this PR.  The new code is not called by operational code now, but may enable a solution for single-source alignment that currently is not solved.

drizzlepac/haputils/align_utils.py
```diff
@@ -1318,3 +1318,113 @@ def register_photutils_function(name):
...
+    Output size will be image size + 1/2 template size in each dimension.
+    valid: The output consists only of those elements that do not rely on the zero-padding.
+    same: The output is the same size as image, centered with respect to the ‘full’ output.
+    :return: N-D array of same dimensions as image. Size depends on mode parameter.
+    """
+
+    # If this happens, it is probably a mistake
+    if np.ndim(template) > np.ndim(image) or \
+            len([i for i in range(np.ndim(template)) if template.shape[i] > image.shape[i]]) > 0:
+        print("normxcorr2: TEMPLATE larger than IMG. Arguments may be swapped.")
```
mdlpstsci, line 1355: Should this be log.warning() vs print()?

drizzlepac/haputils/align_utils.py
```diff
@@ -1318,3 +1318,113 @@ def register_photutils_function(name):
...
+# Based on Octave implementation by: Benjamin Eltzner, 2014 <b.eltzner@gmx.de>         #
+# Octave/Matlab normxcorr2 implementation in python 3.5                                #
+# Details:                                                                             #
+# Normalized cross-correlation. Similiar results upto 3 significant digits.            #
+# https://github.com/Sabrewarrior/normxcorr2-python/master/norxcorr2.py                #
+# http://lordsabre.blogspot.ca/2017/09/matlab-normxcorr2-implemented-in-python.html    #
+########################################################################################
+
+import numpy as np
+from scipy.signal import fftconvolve
```
mdlpstsci, line 1334: Although it seems this code has been absorbed into the drizzlepac source from the quoted reference, I wonder if the imports should still all be at the top of the module to avoid duplication (import numpy as np) or "hidden" imports.

mdlpstsci (changes requested):
Just some minor changes or reason to leave "as is".

PR #1152, opened 2021-10-05
Define segment thresholds for basically empty images

stsci-hack:
This adds logic to catch the leftover case for an if_elif statment when setting detection thresholds for segment catalog source identification.  The previous logic only looked for 2 types of backgrounds based on '.bkg_type' when there were actually 4 possibilities.  

This addresses the error seen in the 2021-10-05 1k test results.


mdlpstsci (approved):
How did this case get overlooked?  It is good to have it addressed now!

PR #1156, opened 2021-10-08
Modify publish-to-pypi.yml to build wheels

zanecodes:
This change uses the newly added custom steps in [`spacetelescope/action-publish_to_pypi`](https://github.com/spacetelescope/action-publish_to_pypi/) to build and publish wheels for new releases.

See https://jira.stsci.edu/browse/SPB-679 for more information.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -5,22 +5,37 @@ on:
     types: [released]
 
 jobs:
-  build:
-      name: Publish release to PyPI
-      env:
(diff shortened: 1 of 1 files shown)


zanecodes:
> We will wait on merging this until after our next build is finished. This will provide us with the opportunity at that point to make sure we understand the full impact of this change compared to what we are already successfully working with now while avoiding any unintended impact on our current build.

That makes sense to me. One additional thing though; since we opened this PR, we made the decision to use the new reusable GitHub Workflows feature to improve the maintainability of this workflow and implement non-platform-wheel builds.

I think it would make sense to update this PR to use the reworked workflow, as it will afford us the ability to pin the OS version and update other details within Bromine's workflow repository, without having to come back to every other repository and update their workflow files.

PR #1163, opened 2021-10-14
Remove unmodified input files from disk and manifest file

stsci-hack:
Input FLC/FLT images which never get aligned during MVM processing will be deleted from disk and not included in the output manifest file since they are simply unmodified copies of the SVM updated versions of the FLC/FLT files.  These changes are implemented to only be performed when not running in diagnostic_mode (for debugging purposes).  

The logic is currently only implemented for updates to the input files applied by aligning to GAIA, but could be expanded later to keep the files for other reasons such as after the input files had their DQ arrays updated to reflect cosmic-rays identified during MVM processing.  As it is, regardless of DQ array updates, the files will be removed if they are not aligned to GAIA.


stsci-hack:
The changes to the function 'run_align_to_gaia' were implemented in order to align ALL exposures from ALL filters at the same time to each other, then to GAIA.  The code had originally only aligned the first filter before returning the name of the reference catalog used.

PR #1166, opened 2021-10-18
Modified azure-pipeline.yml to use ubuntu-latest versus ubuntu-16.04

mdlpstsci:
Modified azure-pipeline.yml to use ubuntu-latest versus ubuntu-16.04.  This was necessary
as support for Ubuntu 16.04 was removed from Azure DevOps on 18 October 2021.  See this
URL for details: https://github.com/actions/virtual-environments/issues/3287

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -27,7 +27,7 @@ jobs:
 
 - job: 'PEP8'
   pool:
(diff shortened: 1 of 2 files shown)


mdlpstsci:
My bad - blinded by the light.  Addressed the problem by changing ALL instances of ubuntu-16.04 to ubuntu-latest.

PR #1178, opened 2021-10-27
Fix the units of a few variables in the output Point and Segmentation catalogs

mdlpstsci:
The MSkyAp2 and StdevAp2 are based on an estimate of the background in an annulus beyond Aperture 2 (the outer aperture).  Bck looks like it is per unit area.  The descriptions units were updated for these values in any catalog where they are present.


stsci-hack (approved):
Simple enough changes to the descriptions.

PR #1181, opened 2021-11-03
Create an empty manifest file for datasets which do not process (e.g., Grism/Prism only)

mdlpstsci:
- Moved initialization of grism and ramp lists, as well as the found_data indicator before the call to poller_utils.  Moving the list initialization ensures no exception will be generated upon exit.
- Do not add the "no_data_trl" (no data being processed trailer file) to the product_list.
- Only append the the astrodriz trailer file to the total trailer file if data_found = True.
- Updated comments as necessary.


stsci-hack (approved):
Change make sense to me.

PR #1185, opened 2021-11-05
Added "demonstration" routine, test_svm_demo.py, which uses fixtures

mdlpstsci:
Added "demonstration" routine, test_svm_demo.py, which uses fixtures for basic functionality.  This is a WIP.  The poller file needs to be in the directory with the tests.  Warren, Mike, and Michele need to discuss what they want in terms of generic support fixtures.


tests/hap/test_svm_demo.py
```diff
@@ -0,0 +1,146 @@
...
+            flt_list.append(fn.split("_")[0])
+
+    # Get test data through astroquery - only retrieve the pipeline processed FLC/FLT files
+    # (e.g., j*_flc.fits).  Soon the SVM and MVM products will be pipeline produced!
+    if flc_list:
+        flcfiles = aqutils.retrieve_observation(flc_list, suffix=["FLC"], product_type="pipeline")
+    if flt_list:
+        fltfiles = aqutils.retrieve_observation(flt_list, suffix=["FLT"], product_type="pipeline")
+
+    flcfiles.extend(fltfiles)
```
stsci-hack, line 63: The function 'retrieve_observation' supports specifying multiple suffix values in the same call, even if there are no products with a specified suffix.  A simple list of rootnames in 'ipppssoot' format with 'suffix=["FLT","FLC"]' provided to a single call to 'retrieve_observation' would replicate this entire function if I am not mistaken.  This would work, but I am left wondering what reason there was for breaking down this functionality like this.

tests/hap/test_svm_demo.py
```diff
@@ -0,0 +1,146 @@
...
+      * Note: When running this test, the `--basetemp` directory should be set to a unique
+        existing directory to avoid deleting previous test output.
+      * The POLLER_FILE exists in the tests/hap directory.
+
+    *** The --basetemp does not seem to be working at this time!!!
+
+"""
+
+WCS_SUB_NAME = "FIT_SVM_GAIA"
+POLLER_FILE = "acs_8f6_61_input.out"
```
stsci-hack, line 29: This effectively hard-codes the test for this single dataset.  Why not assign this poller file a more generic name, since the code itself doesn't really care, then the test can be updated later by simply replacing the contents without having to change this test as well?  For example, rename the current file as 'svm_test1_input.out', then use that as "POLLER_FILE".  What is here will still work, but is not any different from simply having the contents of the file in this module directly with the way it is defined.

stsci-hack (commented):
A couple of comments came to mind so far while reviewing.

tests/hap/test_svm_demo.py
mdlpstsci: This was my way of only downloading the files actually specified in the SVM poller file for processing.  I could no longer remember if we processed the FLC to generate the SVM FLC and then added the WCS solution to the SVM FLT, or if we do not produce (or no longer produce) an SVM FLT.  Looking at Matt's latest processing, it looks like we do not produce an SVM FLT when CTE processing is done.

tests/hap/test_svm_demo.py
mdlpstsci: This is true.  Since I mean for this code to be a demonstration and an example for discussion before I parameterize or do other complex stuff, I thought this was OK for now.  I see there are JWST examples where they have done this hard-coding for testing of particular results.  Passing parameters to fixtures is awkward (as I have been researching do this very thing).

mdlpstsci:
I think the former team-burgundy should have a meeting to discuss what is wanted before any further development is done.  The algorithm that should be implemented is what we all want and not just one person's perspective.

tests/hap/test_svm_demo.py
stsci-hack: Makes sense now.  Thanks for the explanation.

tests/hap/test_svm_demo.py
stsci-hack: Good for this test then.

stsci-hack (dismissed):
Nice framework.

mdlpstsci:
The file generated for this PR, test_svm_demo.py, is a demonstration Pytest for HAP.  At this time, each dataset to be tested will have its own test module.  This is to say the test modules will not be parametrized at this time.

stsci-hack (approved):
Works for me...

PR #1187, opened 2021-11-10
Pipeline align problems

stsci-hack:
This includes several updates to the alignment code that affect the standard pipeline processing as well as the subsequent SVM 
and MVM processing.  Changes to logic include:

- Immediately flagging a fit as 'unsuccessful' if number of final matches is less than MIN_FIT_MATCHES which has been reset to 4.
- only performing magnitude correlation checking when too few matches were included. 
- correctly build the dqmask for source identification with 'bitwise_and' instead of 'bitwise_or' 
- when performing single image CR classification, protect against some sources having moments with a value of np.nan and combine all criteria correctly while insuring correct sources are identified as CRs
- force a deepcopy of a new focus_dict each time it gets checked to avoid continuing to mix the focus measures from all attempts so that each alignment attempt gets correctly(?) evaluated on its own.
- insure all input images have the same WCSNAME when performing pipeline alignment to avoid imprinting differences from one catalog to another on final fit and destroying the relative alignment. 
- Preferentially select fit with most matches even if RMS is not lower (seen in SBC fits in j8ep07 where lower RMS was for bad fit)
- Change criteria for identifying singular fits to be large rot or scale with large skew as well.
- Changed fit_quality=1 criteria to be based on config file parameter 'MAX_FIT_RMS', which was also increased from 10 to 20
- Revised background window size for HRC, SBC, and WFC3/IR to better match smaller field-of-view and plate-scale to improve source finding

In addition, some initial work has been started to change how the logging gets initialized (and can be closed) for the drizzlepac.align.


stsci-hack:
These changes were tested against data from the following visits.

ia0m04:
  - Misaligned in 25-Oct-2021 tests, aligned in Jul-2021
  - WFC3/IR visit which failed to align due to CRs being included for as alignment sources
  ==> Resolution: change 'bitwise_or' to 'bitwise_and' in 'build_dqmask of HAPImage class. [Confirmed]

ibeveh:
  - WFC3/IR data misaligned in 25-Oct-21 test
  - misalignment based on nmatches < 4 for fit to GAIA
  ==> Resolution: Added logic to consider nmatches < 4 (pars['general']['MIN_FIT_MATCHES']) as a failure and try another catalog instead

ib3m53:
  - WFC3/UVIS data of 47Tuc? or some other globular cluster (or similar field)
  - F814W and F555W exposures misaligned to each other
  - F555W exposures were a mix of short and very long (saturated) exposures in the SAME ASN
  ==> Resolution: Reprocessed ASNs with runastrodriz to obtain new, consistent fit to GAIAeDR3
                  after modifying magnitude check on fit for nmatches. [Confirmed]

ibfn01:
  - WFC3/IR data aligns fine
  - WFC3/UVIS data misaligned in 25-Oct-21 test
  - Problem with alignment:
    => F275W ASN empty after performing single-image CRclean algorithm triggering Exception in classify_sources
  ==> Resolution: Update logic in 'classify_sources' to handle sources with np.nan values for moments.


iblo29:
  - WFC3/IR single star exposure
  - misaligned in 25-Oct-21 test
  ==> Aligned perfectly with latest changes. [Confirmed]

j9ns01:
  - Misaligned, reported by RLW
  - ACS/HRC

    From MAST:
      j9ns01mlq_flt.fits CLEAR1S F435W IDC_q692007bj-FIT_REL_GAIADR2 J9NS01050
      j9ns01mmq_flt.fits CLEAR1S F435W IDC_q692007bj-FIT_REL_GAIADR2 J9NS01050
      j9ns01mnq_flt.fits F850LP CLEAR2S IDC_q692007bj-FIT_REL_GAIADR2 J9NS01060
      j9ns01moq_flt.fits F850LP CLEAR2S IDC_q692007bj-FIT_REL_GAIADR2 J9NS01060
      j9ns01mpq_flt.fits CLEAR1S F330W IDC_q692007bj-HSC30 J9NS01070
      j9ns01mqq_flt.fits CLEAR1S F330W IDC_q692007bj-HSC30 J9NS01070
      j9ns01mrq_flt.fits CLEAR1S F330W IDC_q692007bj-FIT_REL_GAIADR1 J9NS01080
      j9ns01msq_flt.fits CLEAR1S F330W IDC_q692007bj-FIT_REL_GAIADR1 J9NS01080
  ==> PROBLEM: Identified as stemming from some images NOT having the same a priori WCS defined (e.g., HSC30 vs GSC240).
               This results in an offset between the multiple WCS systems being imprinted into the final single fit due 
               to the conversion of (x,y) into (RA, Dec) for the fit.

drizzlepac/align.py
```diff
@@ -483,28 +489,35 @@ def perform_align(input_list, catalog_list, num_sources, archive=False, clobber=
...
                                 log.info("Better solution found!")
                                 best_fit_rms = fit_rms
+                                best_num_matches = fit_num
                                 best_fit_label = (catalog_name, algorithm_name)
 
                                 best_fit_status_dict = fit_status_dict.copy()
                                 best_fit_qual = fit_quality
-                            elif fit_quality == best_fit_qual:  # new solution same level of fit_quality. Choose whichever one has the lowest total rms as "best" and keep looping.
+                            elif fit_quality == best_fit_qual:
+                                # new solution same level of fit_quality. Choose whichever one has the lowest total rms as "best" and keep looping.
```
mdlpstsci, line 511: Update comment as the solution also needs better number of matches.

drizzlepac/align.py
```diff
@@ -694,6 +707,7 @@ def determine_fit_quality(imglist, filtered_table, catalogs_remaining, align_par
     overall_comp = False
 
     do_consistency_check = align_pars['determine_fit_quality'].get('consistency_check', True)
```
mdlpstsci, line 709: Is "consistency_check" an obsolete or deprecated value in the parameter files as it does not currently appear in any of the json parameter files.  I understand there is a default value here, but are there times when we do not want the consistency check done?

Ok, I looked a bit more after I got confused between the use of "do_consistency_check" and "consistency_check".  Is "consistency_check" really "is_consistent"?

drizzlepac/runastrodriz.py
```diff
@@ -709,7 +716,9 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
             pass
 
     # Append final timestamp to trailer file...
-    _final_msg = '%s: Finished processing %s \n' % (_getTime(), inFilename)
+    end_time = _getTime()
+    _delta_time = time.time() - init_time
+    _final_msg = '%s: Finished processing %s in %.2f seconds \n' % (end_time, inFilename, _delta_time)
```
mdlpstsci, line 721: Just curious why the initial time is in SSE, but the end time is in a human-readable time string?

stsci-hack: The (human readable) 'end_time' value is a string which can not be used for math operations, so the 'init_time' is used to compute the elapsed processing time instead.

mdlpstsci (changes requested):
Just some simple questions or changes for comments.

drizzlepac/align.py
stsci-hack: Updated in db1f6d1.

drizzlepac/align.py
stsci-hack: The parameter 'consistency_check' here was renamed to 'do_consistency_check' in db1f6d1 to avoid any future confusion with the variable of the same name and different meaning later in the function.

drizzlepac/runastrodriz.py
mdlpstsci: Ok.  Just a practical use of time formats.

mdlpstsci (approved):

PR #1189, opened 2021-11-15
Added a new PyTest test_svm_j97e06 

mdlpstsci:
Added a new PyTest, test_svm_j97e06 which is an ACS WFC, Grism and Direct image, and six filters dataset.


tests/hap/test_svm_j97e06.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+@pytest.fixture(scope="module")
+def gather_output_data(construct_manifest_filename):
+    # Determine the filenames of all the output files from the manifest
+    print("\nManifest Filename: {}".format(construct_manifest_filename))
+    files = []
+    with open(construct_manifest_filename, 'r') as fout:
+        for line in fout.readlines():
+            files.append(line)
```
stsci-hack, line 127:
```suggestion
            files.append(line.rstrip("\n")
```
This should be needed in order for test code to work when using '.endswith(".fits")' (like on Line 184).

tests/hap/test_svm_j97e06.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def test_svm_wcs(gather_output_data):
+    # Check the output primary WCSNAME includes FIT_SVM_GAIA as part of the string value
+    tdp_files = [files for files in gather_output_data if files.lower().find("total") > -1 and files.lower().endswith(".fits")]
+
+    for tdp in tdp_files:
+        wcsname = fits.getval(tdp, "WCSNAME", ext=1).upper()
+        print("\ntest_svm_wcs.  WCSNAME: {} Output file: {}".format(wcsname, tdp))
+        assert WCS_SUB_NAME in wcsname, f"WCSNAME is not as expected for file {tdp}."
```
stsci-hack, line 188: Will this test stop immediately (not evaluate any more files) on the FIRST instance where the assert fails?  Is this the best behavior for this type of test?  This comment also applies to the rest of the tests here as well.

tests/hap/test_svm_j97e06.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def test_svm_point_cat_numsources(gather_output_data):
+    # Check the output catalog length
+    cat_files = [files for files in gather_output_data if files.lower().endswith("point-cat.ecsv")]
+
+    for cat in cat_files:
+        table_length = len(ascii.read(cat, format="ecsv"))
+        print("\ntest_svm_point_cat_numsources. Number of sources in catalog {} is {}.".format(cat, table_length))
+        assert table_length > MIN_CAT_LENGTH and table_length < MAX_CAT_LENGTH_POINT, f"Catalog file {cat} is unexpectedly empty"
```
stsci-hack, line 198: The failure message is potentially mis-leading, as the failure can be triggered by the catalog having more sources than expected, yet the message indicates the catalog is "unexpectedly empty".  The comment should be revised to be more accurate for the purposes of better understanding the actual mode of failure.  This actually also applies to the comments for the next test as well.

stsci-hack (changes requested):
A couple of issues need to be addressed to make this test as useful as intended.

tests/hap/test_svm_j97e06.py
mdlpstsci: Fixed so the assert is not done in a loop in any of the tests.

tests/hap/test_svm_j97e06.py
mdlpstsci: Fixed failure comments.

tests/hap/template_svm_demo.py
```diff
@@ -155,17 +159,20 @@ def test_svm_wcs(gather_output_data):
     # Check the output primary WCSNAME includes FIT_SVM_GAIA as part of the string value
     tdp_files = [files for files in gather_output_data if files.lower().find("total") > -1 and files.lower().endswith(".fits")]
 
-    for tdp in tdp_files:
-        wcsname = fits.getval(tdp, "WCSNAME", ext=1).upper()
-        print("\ntest_svm_wcs.  WCSNAME: {} Output file: {}".format(wcsname, tdp))
-        assert WCS_SUB_NAME in wcsname, f"WCSNAME is not as expected for file {tdp}."
+    # This check is for all total data products which have the same "type" of WCSNAME -
```
stsci-hack, line 162: The test no longer verifies that the WCSNAME is coming out aligned to GAIA with this most recent change.  Having one test check for WCS_SUB_NAME insures that the alignment is coming out as expected, while checking to see that they are all the same can pass even if they are all -HSC30 apriori WCSs, which may actually indicate a failure to align.  This will work just fine, but the purpose of this test is now more limited and really most useful for validating the WCSs between detectors.

stsci-hack (approved):
My comments on the WCSNAME test are only meant to confirm the intent of the test.  If you have no concerns after reviewing my comments, then this can be merged as-is.

tests/hap/template_svm_demo.py
mdlpstsci: I am OK with this as my intention is to update this "template" with the new tests we have put into the individual dataset tests so in the future one can just mine this file for ideas.  The upgrade should entirely be its own ticket.

PR #1190, opened 2021-11-15
Add new pytest for SVM HRC and SBC processing

stsci-hack:
This adds a new pytest derived from the template implemented under #1185 to validate SVM processing of ACS/HRC and ACS/SBC data.  The test only checks a couple of simple things, such as WCSNAME with FIT_SVM_GAIA and that the total catalogs exist with a rough (rtol=0.1) number of sources.


stsci-hack:
The checks are failing for expected reasons; specifically, that #1187 has not yet been merged while the test is using PASS/FAIL criteria (number of sources and expected WCSNAME, to be precise) derived from running under #1187.

mdlpstsci (approved):
Looks good.

PR #1191, opened 2021-11-16
Add pytest for WFC3/IR SVM processing

stsci-hack:
This adds some pytests for the SVM processing of WFC3/IR data based on data from visit ib6807 and using the template for the tests developed under #1185 .

### DIFF
--- a/tests/hap/test_svm_wfc3ir.py
+++ b/tests/hap/test_svm_wfc3ir.py
@@ -2,7 +2,6 @@
 
 """
 import datetime
-import glob
 import os
 import pytest
 import numpy as np
@@ -43,7 +42,11 @@ def read_csv_for_filenames():
 
 
 @pytest.fixture(scope="module")
-def gather_data_for_processing(read_csv_for_filenames):
+def gather_data_for_processing(read_csv_for_filenames, tmp_path_factory):
+    # create working directory specified for the test
+    curdir = tmp_path_factory.mktemp(os.path.basename(__file__))
+    os.chdir(curdir)
+
     # Establish FLC/FLT lists and obtain the requested data
     flc_flag = ""
     flt_flag = ""
--- a/tests/hap/test_svm_wfc3ir.py
+++ b/tests/hap/test_svm_wfc3ir.py
@@ -21,12 +21,14 @@ from pathlib import Path
         existing directory to avoid deleting previous test output.
       * The POLLER_FILE exists in the tests/hap directory.
 
-    *** The --basetemp does not seem to be working at this time!!!
-
 """
 
 WCS_SUB_NAME = "FIT_SVM_GAIA"
 POLLER_FILE = "wfc3_ir_ib6807_input.out"
+
+# Gather all expected values used for determining pass/fail criteria here
+expected_point_sources = {'ir': 358}
+expected_seg_sources = {'ir': 298}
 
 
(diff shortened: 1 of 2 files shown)


tests/hap/test_svm_wfc3ir.py
```diff
@@ -0,0 +1,213 @@
...
+    test_svm_demo.py
+
+    This test file can be executed in the following manner:
+        $ pytest -s --basetemp=/internal/hladata/yourUniqueDirectoryHere test_svm.py >& test_svm.log &
+        $ tail -f test_svm.log
+      * Note: When running this test, the `--basetemp` directory should be set to a unique
+        existing directory to avoid deleting previous test output.
+      * The POLLER_FILE exists in the tests/hap directory.
+
+    *** The --basetemp does not seem to be working at this time!!!
```
mdlpstsci, line 24: Suggest to update the commentary to be specific to this test and remove the --basetemp warning which is obsolete due to the tmpdir addition.

tests/hap/test_svm_wfc3ir.py
```diff
@@ -0,0 +1,213 @@
...
+        table_length = len(ascii.read(cat, format="ecsv"))
+        print("\ntest_svm_cat_sources. Number of sources in catalog {} is {}.".format(cat, table_length))
+        valid_tables[cat] = table_length > 0
+    bad_tables = [cat for cat in cat_files if not valid_tables[cat]]
+    assert len(bad_tables) == 0, f"Catalog file(s) {bad_tables} is/are unexpectedly empty"
+
+def test_svm_point_cats(gather_output_data):
+    # Check that the point catalogs have the expected number of sources
+    cat_files = [files for files in gather_output_data if files.lower().endswith("point-cat.ecsv")]
+    expected_sources = {'ir': 358}
```
mdlpstsci, line 187: Just a suggestion for this and the next test as this is developer preference.  Putting expected_sources here keeps the values with the test, or putting the variable at the top of the module keeps the  variable which likely to change as algorithms evolve in one handy location.

mdlpstsci (changes requested):
Only modest suggestions though I do think updating the doc at the beginning of the file should reflect the current test.

tests/hap/test_svm_wfc3ir.py
```diff
@@ -0,0 +1,213 @@
...
+
+    num_sources = {cat:len(ascii.read(cat, format="ecsv")) for cat in cat_files}
```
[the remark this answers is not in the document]
stsci-hack: Thanks!

tests/hap/test_svm_wfc3ir.py
stsci-hack: They have been renamed (to unique values) and moved to the top of the module in 2893c91.  This does make it easier to maintain these tests.

tests/hap/test_svm_wfc3ir.py
stsci-hack: Fixed in 2893c91.

stsci-hack:
The checks are failing for expected reasons; specifically, that #1187 has not yet been merged while the test is using PASS/FAIL criteria (number of sources and expected WCSNAME, to be precise) derived from running under #1187.

mdlpstsci (approved):
Changes are approved.

PR #1192, opened 2021-11-19
Adding a new WFC3 test for both UVIS and IR images and catalogs.

mdlpstsci:
PyTest test_svm_ibqk07 which is a WFC3 dataset using both IR and UVIS detectors, though the IR and UVIS have different WCS solutions.


stsci-hack (approved):
Looks like it uses everything we developed for the other tests.  Nice!

PR #1193, opened 2021-11-19
Fix an issue with WFPC2 flat field file extension name

mcara:
This PR addresses an issue with WFPC flat filed files having extension name `FLT` instead of the expected (by the code) extension `SCI`. When trying to drizzle WFPC2 image with IVM weighting, this unexpected extension type would result in a crash.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,7 +15,7 @@ The following notes provide some details on what has been revised for each
 version in reverse chronological order (most recent version at the top
 of the list).
 
-3.3.1 (unreleased)
+3.3.2 (unreleased)
 ==================
 
 - Detect extension name from WFPC2 flatfield files. [#1193]
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,8 +15,14 @@ The following notes provide some details on what has been revised for each
 version in reverse chronological order (most recent version at the top
 of the list).
 
-3.3.0 (28-Sep-2021)
+3.3.1 (unreleased)
 ==================
+
+- Detect extension name from WFPC2 flatfield files. [#1193]
+
+
+3.3.0 (28-Sep-2021)
+===================
 This version includes all the functionality needed to generate
 source catalogs, both point source and extended (segment) source
 catalogs, during single-visit mosaic (SVM) processing.  In fact,
--- a/drizzlepac/wfpc2Data.py
+++ b/drizzlepac/wfpc2Data.py
@@ -164,7 +164,7 @@ class WFPC2InputImage (imageObject):
         # Convert the science data to electrons
         self.doUnitConversions()
 
-    def getflat(self,chip):
+    def getflat(self, chip, flat_ext=None):
         """
         Method for retrieving a detector's flat field.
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -15,8 +15,14 @@ The following notes provide some details on what has been revised for each
 version in reverse chronological order (most recent version at the top
 of the list).
 
-3.3.0 (28-Sep-2021)
+3.3.1 (unreleased)
 ==================
+
+- Detect extension name fro WFPC2 flatfield files. [#1193]
```
stsci-hack, line 21:
```suggestion
- Detect extension name from WFPC2 flatfield files. [#1193]
```

stsci-hack (approved):
Simple enough change to make the method to get the flatfield data more general.

drizzlepac/wfpc2Data.py
```diff
@@ -164,7 +164,7 @@ def setInstrumentParameters(self, instrpars):
         # Convert the science data to electrons
         self.doUnitConversions()
 
-    def getflat(self,chip):
+    def getflat(self, chip, flat_ext=None):
```
stsci-hack, line 167:
```suggestion
    def getflat(self, chip, flat_ext=''):
```
This would avoid the need for adding lines 179-180.  Right?

stsci-hack (approved):
My comment can be ignored if you don't feel it makes a difference.

drizzlepac/wfpc2Data.py
mcara: `''` is a valid extension (i.e., no `EXTNAME` in the header) while `None` is not valid. This allows to use the default `SCI` extension for non-WFPC2 flat field files.

mcara:
I had to make another change because `flat_ext` was incorrectly used on the (science) image itself.

@stsci-hack Please comment on the following change: adding `writefits=False` to `fileutil.openImage()`. I did this in order to reduce the number of times flat field FLT file is written to the disc. I have noticed that for 4 WFPC2 files with 4 chips, the FITS version of the flat field file was written out 32 times (when converting from GEIS). Proposed change keeps the file in memory and this should not be an issue since `getflat()`returns just the data array from the FITS file.

Another alternative is to set `clobber=False`. This would create FITS file (from GEIS) but only once. If a file with the same name is detected - it will simply be opened (without writing it out again). My worry about this approach is that if FITS file somehow diverged from the original GEIS file - the code will load the modified FITS instead of the original GEIS.

What is your opinion?

mcara:
The last commit adds more complicated logic to cache MEF flat field file names and extensions. This update will convert GEIS->MEF only once every time `drizzlepac` is run but will re-open existing MEF for each extension. So, for example, when drizzling 10 WFPC2 images that all use the same flat field file, it will be converted only once to MEF and then for each image file and extension, flat field will be read from the stored MEF.

Basically, the only overhead for WFPC2 files is conversion from GEIS to MEF which will be done only once for each distinct flat field file compared to how it was done previously once for each _image file_ AND _extension_ (!).

mcara:
Failures in HAP/SVM tests seem to be unrelated to this PR.

PR #1198, opened 2021-11-24
Ignore largest

stsci-hack:
This set of changes addresses problems with some datasets taking longer than necessary (days instead of minutes) to deblend the images when generating the segment source catalogs during SVM processing.  The changes include:

- ignoring the largest segment when deblending, when the largest is greater than the deblend_limit (1.5% area of image)
- background-subtracting and clipping negative fluxes from input science image when detecting sources with RickerWavelet kernel
- only pass in the array from the RickerWavelet kernel to be consistent with how the Gaussian kernel gets used by detect_segments
- increase the rw2d_nsigma parameter value from 3.0 to 10.0 for segment source detection to minimize the influence of the background or nebulosity during segment catalog source identification

These changes were used to process the following datasets in less than 90minutes, all of which never completed SVM processing with v3.3.1 due to deblending taking so long:

- jcol06
- j8oc01
- jcdhb7


stsci-hack:
**NOTE: Support for astropy was pinned to <5.0 in order to avoid problems with astropy in the next release of this package due to incompatibilities introduced by astropy v5.0.0.**

mdlpstsci (approved):
Clever algorithm!

PR #1199, opened 2021-11-29
Add logic to create empty catalogs for blank data

stsci-hack:
These changes insure that blank total products generate empty catalogs instead of throwing an Exception.


mdlpstsci (approved):
Sorry it has taken so long.  I had to follow up with what happens with the filter catalogs.  I see the code to handle the filter catalogs was installed two months ago!

PR #1201, opened 2021-12-02
Add support for MVM header rules files

stsci-hack:
These changes add support for the addition and use of fitsblender rules files specifically for MVM processing alongside the rules files already provided for SVM processing.  

In addition, the SCELLID keyword is being added to the MVM product headers, a change which was necessary since this information is new to the headers and not already provided by the pipeline during previous processing steps.  

**NOTE:**  This PR requires the merging of [FITSBLENDER PR #30](https://github.com/spacetelescope/fitsblender/pull/30) in order to work correctly.


drizzlepac/haputils/product.py
```diff
@@ -1165,11 +1166,12 @@ def wcs_drizzle_product(self, meta_wcs):
...
                                   **drizzle_pars)
 
-        # Update product with SVM-specific keywords based on the footprint
+        # Update product with MVM-specific keywords based on the footprint
         with fits.open(self.drizzle_filename, mode='update') as hdu:
             for kw in self.mask_kws:
                 hdu[("SCI", 1)].header[kw] = tuple(self.mask_kws[kw])
-
+            # Add SCELLID keyword to MVM product
+            hdu[0].header['SCELLID'] = self.cell_id
```
lisaesherbert, line 1174: Maybe I am wrong, but seems simple enough to make that 0 '"SCI", 1' (but may have to move into a loop?)

stsci-hack: Done, as requested, in 36282ee.

mdlpstsci (approved):
Code looks reasonable and straightforward.  Admittedly, I am not familiar with the rules files themselves.

PR #1202, opened 2021-12-02
SVM: Ensure list of headerlets contains all the WCS solutions available

mdlpstsci:
Reworked to ensure *all* the alternate WCS solutions, except for OPUS, are
stored as headerlets.  As the alternate WCS solutions may not have an HDRNAME,
a HDRNAME is generated on-the-fly by using the filename, stripping off the
".fits" suffix, and adding the WCSNAME and "-hlet.fits". For example, a HDRNAME
might be hst_9029_01_acs_wfc_f775w_j8ca01at_flc_IDC_0461802ej-FIT_SVM_GAIAeDR3-hlet.fits.
This change ensures the list of headerlets contains all of the solutions, except
perhaps for the OPUS solution.


stsci-hack (approved):
That logic seems perfectly reasonable to me.

PR #1205, opened 2021-12-03
Avoid empty HDRTAB rows

stsci-hack:
These changes revise the logic used to update the HDRTAB extension table with the HAPEXPNAME column to avoid creating empty rows in the HDRTAB.  Previously, a new column including the new HAPEXPNAME values was created by searching for the input exposures in ALL the total products resulting in a column that was the sum of all the total drizzle products input exposures in length.  This logic was revised to only update the HDRTAB table for the inputs associated with  the total product image being updated.


mdlpstsci (approved):
Straightforward change seems fine to me.

PR #1208, opened 2021-12-06
Add Python 3.10 to Azure test matrix

stsci-hack:
These changes are the changes needed to add Python 3.10 testing to the nightly Jenkins regression tests, while also documenting how to add Python 3.10 to the Azure tests as well.


stsci-hack:
Need to merge in order to test.

PR #1211, opened 2021-12-07
Special handling of SBC/HRC MVM data, clarify messsages, return codes

mdlpstsci:
Reworked the analyze_wrapper function to accept two new parameters, an indicator for SVM or MVM processing and an indicator as to whether or
not SBC/HRC data should be used for MVM processing.

The MVM indicator has two uses: (1) Marks Grism/Prism data as "do not
process" in the filtered table associated with the poller file, and
(2) acts as a gate for further checking of the filtered table data when
SBC/HRC data should not be used for processing.

Clarified messages which are generated for the log for the cases with
associated return codes:

 No viable images in single/multi-visit table - no processing done.
 Return code = 65

 Only non-viable or SBC/HRC images in the multi-visit table - no processing done.
 Return code = 55

### DIFF
--- a/drizzlepac/hapmultisequencer.py
+++ b/drizzlepac/hapmultisequencer.py
@@ -43,13 +43,16 @@ import fnmatch
 import logging
 import os
 import pickle
+import re
 import sys
 import traceback
 
+from astropy.io import ascii
 from astropy.table import Table
 import numpy as np
 import drizzlepac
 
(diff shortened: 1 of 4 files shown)


stsci-hack (dismissed):
I like how you implemented the return code as an enumerated object.  All the code looked good to me.  This can be merged when you feel it is ready.

stsci-hack (approved):

PR #1217, opened 2021-12-15
Compute own NMATCHES value for fit validation

stsci-hack:
This change stops relying on the tweakwcs fit information for the 'nmatches' value altogether, since it has proven to be inconsistent with what actually gets used for the final fit due to outlier rejection used within tweakwcs.


stsci-hack:
This change has been successfully used on data from visit 'icphp2' which originally reported NMATCHES=4 when only 2 were reportedly being used.

PR #1218, opened 2021-12-16
Set NMATCH limits for fitgeoms to INS requested values

stsci-hack:
This codifies the limits on minimum number of cross-matches that are acceptable for each type of fit (e.g., rscale, rshift,...) that gets performed during alignment as requested by INS.


stsci-hack:
Additional tips on how to update the code to more directly control the use of 'nmatches' during the fitting by 'tweakwcs' can be found in the comments for [Tweakwcs issue #149](https://github.com/spacetelescope/tweakwcs/issues/149).

mdlpstsci (approved):

PR #1219, opened 2021-12-16
Reduce memory use in hapmultisequencer

stsci-hack:
These changes fix problems with memory use during processing with hapmultisequencer.  Specifically, footprint masks are properly defined one per layer, not one per exposure, and they are deleted as soon as they are no longer needed by the processing.  This reduces the memory usage during this stage of processing to a maximum of about 2Gb based on the memory used by an array of a single full skycell layer.


mdlpstsci (approved):
Reminds us all to be more careful and delete variables which represent lots of memory when no longer in use!

PR #1220, opened 2021-12-17
Prevent WCS consistency update from overwriting latest idctab

stsci-hack:
Logic needed to be changed to limit the check for consistent WCSs (implemented in #1187) to only the most current IDCTAB, not WCSs based on older calibrations.  This check is needed to avoid imprinting mis-alignment stemming from different zero-points of the different WCSs, a problem seen while developing #1187.


mdlpstsci (approved):
The logic looks fine.

Issue #1216, opened 2021-12-14
update how make_poller_files.py generates visit number 


stscijgbot-hstdp:
_Issue [HLA-680](https://jira.stsci.edu/browse/HLA-680) was created on JIRA by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude):_

Used to just get get it from the header value "LINENUM". In the case of paralell observations, should check "INSTRUME" against "PRIMESI" and if they don't match, get visit num from the filename.

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-680?focusedCommentId=562691#comment-562691):_

Discovery datasets:
j8fndi* (any of these)
j8nbk9

PR #1221, opened 2021-12-23
update how make_poller_files.py generates visit numbers to account an edge case

dulude:
**RELATED ISSUES AND TICKETS**
- #1216 
- [HLA-680](https://jira.stsci.edu/browse/HLA-680)

**BACKGROUND**
- For a small number of edge case images (those with header values "PRIMESI" != "INSTRUME"), the first two characters of header value "LINENUM" do not match the values in the 5th and 6th places of the filename or header value "ROOTNAME".
- make_poller_files.py was populating the "visit number" column of .out files with the incorrect value because it was using the first two characters of header value "LINENUM" for these edge case images.

**SUMMARY OF CHANGES**
- Added logic to check for the edge case and if detected, use the 5th and 6th characters of "ROOTNAME" header value rather than "LINENUM" to produce "visit number" values. 
- For all other non-edge case images, "visit number" values are still derived as they were before based on the "LINENUM" header value.
- A similar issue was found in diagnostic_utils.py. A similar fix was applied to account for the edge case when generating a value for visit number.

### DIFF
--- a/drizzlepac/haputils/diagnostic_utils.py
+++ b/drizzlepac/haputils/diagnostic_utils.py
@@ -176,7 +176,10 @@ class HapDiagnostic(object):
             self.out_dict['general information'][dict_keys[key]] = self.header[key]
         # Now, add items which require more interpretation
         try:
-            self.out_dict['general information']['visit'] = self.header['linenum'].split(".")[0]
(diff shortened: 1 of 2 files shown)


stsci-hack (approved):
LGTM.

PR #1222, opened 2021-12-26
Fix incorrect drizzling of 0-weight input pixels

mcara:
This PR fixes a bug in the C code of the core drizzle algorithm due to which flux from the input image could be dropped onto the output (resampled) image even though those input pixels have zero weights.

This is a port of the bug fix from https://github.com/spacetelescope/drizzle/pull/79

### DIFF
--- a/tests/drizzle/test_drizzle.py
+++ b/tests/drizzle/test_drizzle.py
@@ -1,0 +1,37 @@
+import numpy as np
+from astropy import wcs
+
+from drizzlepac import cdriz
+
+def test_zero_input_weight():
+    """
+    Test do_driz square kernel with grid
+    """
+    insci = np.ones((200, 400), dtype=np.float32)
+    inwht = np.ones((200, 400), dtype=np.float32)
+    inwht[:, 150:155] = 0
+    tflux = np.sum(inwht)
+
+    for kernel in ['square', 'point', 'turbo']:
+        # initialize output:
+        outsci = np.zeros((210, 410), dtype=np.float32)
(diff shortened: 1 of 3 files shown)


mcara:
This PR should be merged only after all regression tests are fixed and passing in order to understand its effect on the tests.

mcara:
Unit test failures are unrelated to this PR and are the same as described in https://github.com/spacetelescope/drizzlepac/issues/1225

tests/drizzle/test_drizzle.py
```diff
@@ -0,0 +1,37 @@
+import numpy as np
+from astropy import wcs
+
+from drizzlepac import cdriz
+
+def test_zero_input_weight():
```
eslavich, line 6: I tried running this test on the master branch and it passed there too.  Did I make a mistake, or does the test need adjustment so that it covers the problem?

eslavich (approved):
Confirmed that the test does indeed fail on master with the latest commit.  LGTM!

tests/drizzle/test_drizzle.py
mcara: No mistake. The test was bad.

mcara:
Anyone? @stsci-hack @nden

stsci-hack (dismissed):
This fix makes sense to me.  Having an extra test is also a good idea.

mcara:
@stsci-hack Can you officially approve this PR in the review?

stsci-hack (approved):
Everything seems to work as designed. LGTM.

Issue #1223, opened 2021-12-27
Master is broken due to imports of debug modules


mcara:
When attempting to import `drizzlepac` from the `master`, I get this error:

```
E   ModuleNotFoundError: No module named 'guppy'
```

This is due to [this import.](https://github.com/spacetelescope/drizzlepac/blob/d97700f5aa4c8742085989bdf10f432e71f3f01d/drizzlepac/haputils/cell_utils.py#L6)

Due to this all tests in https://github.com/spacetelescope/drizzlepac/pull/1222 Pr (and probably all other open PRs) are failing.

CC: @stsci-hack

stscijgbot-hstdp:
_This issue is tracked on JIRA as [HLA-682](https://jira.stsci.edu/browse/HLA-682)._

stscijgbot-hstdp:
_Comment by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack) on [JIRA](https://jira.stsci.edu/browse/HLA-682?focusedCommentId=606116#comment-606116):_

Resolved with [Drizzlepac PR #&#x2060;1224](https://github.com/spacetelescope/drizzlepac/pull/1224).

PR #1224, opened 2021-12-27
Remove guppy import

mcara:
Fixes https://github.com/spacetelescope/drizzlepac/issues/1223


mcara:
However, this PR still does not fix all failing unit tests. I think that test that fail permanently (for a long time without fixing them) should be disabled so that new PRs could be tested effectively.

CC: @stsci-hack @mdlpstsci

nden:
👍 
This import was breaking also readthedocs.

PR #1230, opened 2022-01-03
Remove the prefix "hst_" from the manifest filename for MVM products.

mdlpstsci:
This prefix was inadvertently re-introduced when handling the special cases
of no data being available for processing or some other error when reading
the poller file so the total data product attributes were never set.  Hence,
no default manifest filename was defined.


sevenslices (approved):

dulude (approved):
LGTM

stsci-hack:
Testing this by interactively processing data from skycell-p1889x07y19 to verify that this change does not cause problems with the default success use case.

stsci-hack (approved):
LGTM.

PR #1231, opened 2022-01-03
Temporarily "skip" the HAP PyTests which fail in their assertions.

mdlpstsci:
The issues are reference files are no longer in the appropriate locale, or
the expected WCS is not derived.  Interestingly, the tests run successfully
locally.

I will note the majority of the HAP tests may still fail **_as an ensemble_** due to the way the logging has been implemented in the HAP routines combined with the fact the PyTests are run in a single Python session.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -83,7 +83,7 @@ setup(
         'astropy<5.0.0',
         'fitsblender',
         'nictools',
-        'numpy>=1.19',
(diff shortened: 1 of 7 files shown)


stsci-hack (approved):
Simple enough.  Too bad that they all have to be marked as 'skip'.

PR #1232, opened 2022-01-05
Add explicit dependency on spherical-geometry and set min version

mcara:
This PR explicitly adds `spherical_geometry` to the list of dependencies in the `setup.py` because it is used by HAP code and I also set the minimum  version requirement for the `spherical_geometry` to `1.2.22` as previous versions may cause memory corruption.


stsci-hack (approved):
Since all tests pass, this looks good to me.

PR #1233, opened 2022-01-05
Update apriori test to be more explicit with errors

stsci-hack:
These changes make the 'test_apriori' tests less stringent while also reporting a lot more information on failures or Exceptions.  The less stringent limits for this test do not impact the validity of the test since they are looking for situations where the fitting ended up acting singular (wildly wrong, yet successful).  Small variations in the results can come from differences in the proper motions used in the catalog source positions as well as implemented (expected) changes to the sigma-clipping in the fitting itself amongst other valid reasons.

### DIFF
--- a/drizzlepac/haputils/testutils.py
+++ b/drizzlepac/haputils/testutils.py
@@ -1,5 +1,6 @@
 import sys
 import os
+import traceback
 
 from astropy.io import fits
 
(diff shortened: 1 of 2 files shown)


mdlpstsci (approved):
These changes should help us diagnose issues more readily!

PR #1238, opened 2022-01-07
Insure both FLT and FLC headers have same a posteriori fit keywords

stsci-hack:
These changes copy the header keywords from the FLC file header that report the quality of the a posteriori WCS fit into the header of the FLT file.  This insures both headers have the same information so that the DRZ and DRC files will also be identical/consistent.  The set of keywords updated includes: PYWCSVER, UPWCSVER, HDRNAME, NMATCHES, FITGEOM, RMS_RA, RMS_DEC, CRDER1, and CRDER2.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -999,6 +999,8 @@ def verify_alignment(inlist, calfiles, calfiles_flc, trlfile,
             if align_update_files and align_table:
                 # Apply headerlets from alignment to FLT version of the files
                 for fltfile, flcfile in zip(align_update_files, alignfiles):
+                    # Update non-headerlet-based keywords in fltfile
+                    _update_wcs_fit_keywords(fltfile, flcfile)
                     row = align_table[align_table['imageName'] == flcfile]
                     headerlet_file = row['headerletFile'][0]
                     if headerlet_file not in ["None", '']:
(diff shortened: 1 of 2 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -1677,6 +1679,45 @@ def update_active_wcs(filename, wcsname, logfile=None):
...
+        after_kw = 'ASN_MTYP'
+        before_kw = None
+    else:
+        after_kw = hdulist[0].header.cards[-1][0]
+        before_kw = None
+
+    hdulist[0].header.set('UPWCSVER', value=hdulist_flc[0].header['UPWCSVER'],
+                          comment="Version of STWCS used to update the WCS",
+                          after=after_kw, before=before_kw)
+    hdulist[0].header.set('PYWCSVER', value=hdulist_flc[0].header['UPWCSVER'],
```
mdlpstsci, line 1705: Should the value from "FLC" be "PYWCSVER"?  Also, should the comment be "Version of PYWCS used to update the WCS "?  I am unsure about the commentary portion of the FITS Card, but I thought I should confirm.

mdlpstsci (changes requested):
There is just one item which I believe needs to be addressed.

drizzlepac/runastrodriz.py
stsci-hack: It looks as if you found a cut-and-paste error I overlooked.  Fixed in d79d1af.

mdlpstsci (approved):

Issue #1207, opened 2021-12-06
MVM: Make tool to quantify quality of GAIA alignment generic for general use


stscijgbot-hstdp:
_Issue [HLA-674](https://jira.stsci.edu/browse/HLA-674) was created on JIRA by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude):_

This ticket focuses on taking the tool developed as part of [HLA-405](https://jira.stsci.edu/browse/HLA-405) and fleshed out in [HLA-655](https://jira.stsci.edu/browse/HLA-655) and making it generic enough that it can be used in general use.

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-674?focusedCommentId=561539#comment-561539):_

See [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude)'s drizzlepac branch[ HLA-655](https://github.com/dulude/drizzlepac/tree/HLA-655) for the relevant code.

PR #1241, opened 2022-01-13
MVM: Make tool to quantify quality of GAIA alignment generic for general use

dulude:
**Relevant Tickets & Issues**
- [HLA-674](https://jira.stsci.edu/browse/HLA-674)
- #1207

**Summary of Changes**
- Created script _analyze_mvm_gaia_alignment.py_ to statistically quantify how well MVM products are aligned to GAIA sources found in the image footprint defined by the SVM-processed fl(c/t).fits input files.
- Returns unclipped and 3x3-sigma clipped mean, median, and standard deviation for X-axis, Y-axis and RA/Dec on-sky GAIA position residuals
- Optionally generates histograms of all above position residuals, and a vector plot to visualize X and Y GAIA position residuals as a function of location in the field of view.


drizzlepac/devutils/analyze_mvm_gaia_alignment.py
```diff
@@ -35,10 +34,102 @@
...
+    finally:
+        for item in [det_coo_filename, gaia_coo_filename]:
+            if os.path.exists(item):
+                log.debug("Removing temp coord file {}".format(item))
+                os.remove(item)
+    return detection_table, matches_gaia_to_det, matches_det_to_gaia
+# ============================================================================================================
+
+
+def perform(mosaic_imgname, flcflt_list, diagnostic_mode=False, log_level=logutil.logging.INFO,
```
stsci-hack, line 131: The specification of 'flcflt_list' as a parameter is currently misleading.  This parameter has been implemented as a filename (str) whereas it has 'list' in the name and the docstring specifies it should be input as a list.  This parameter probably should be renamed so that both a list and a filename should (and can easily) be supported.  The docstring should be updated to correctly reflect the final implementation.

dulude: issue resolved by commit [79addb4](https://github.com/spacetelescope/drizzlepac/pull/1241/commits/79addb45729f674dffa6c553bbb13dc70e46044b)

drizzlepac/devutils/analyze_mvm_gaia_alignment.py
```diff
@@ -102,60 +228,43 @@ def perform(mosaic_imgname, flcflt_list, diagnostic_mode=False, log_level=loguti
...
-            log.error(err_msg)
-            raise Exception(err_msg)
-    finally:
-        for item in [det_coo_filename, gaia_coo_filename]:
-            if os.path.exists(item):
-                log.debug("Removing temp coord file {}".format(item))
-                os.remove(item)
-
+    # 4: compute FWHM for source finding based on sources in image
+    mpeaks, fwhm = decutils.find_point_sources(mosaic_imgname, mask=np.invert(dao_mask_array),
```
stsci-hack, line 239: You need to force 'dao_mask_array' to be bool for 'np.invert' to work correctly, then pass the result as an np.int16 array for the masking to work correctly.  Inverting an np.int array results in negative values.  It can be done using the following suggestion: 
```suggestion
    mpeaks, fwhm = decutils.find_point_sources(mosaic_imgname, mask=np.invert(dao_mask_array.astype(bool)).astype(np.int16),
```

drizzlepac/devutils/analyze_mvm_gaia_alignment.py
```diff
@@ -102,60 +228,43 @@ def perform(mosaic_imgname, flcflt_list, diagnostic_mode=False, log_level=loguti
...
+        detection_table, matches_gaia_to_det, matches_det_to_gaia = find_and_match_sources(fwhm_value,
+                                                                                           mosaic_hdu,
+                                                                                           mosaic_wcs,
+                                                                                           mosaic_imgname,
+                                                                                           dao_mask_array,
+                                                                                           gaia_table,
+                                                                                           xy_gaia_coords,
+                                                                                           diagnostic_mode=diagnostic_mode,
+                                                                                           log_level=log_level)
+        if len(matches_gaia_to_det) > 10:
```
stsci-hack, line 257: Does this limit need to be hard-coded?  Would it be useful to allow the user to set what limit they are comfortable with, or should it vary depending on how many GAIA sources actually overlap the skycell?  For example, only 8 sources overlap the HST data for p1101x11y05 out of a total of 25 for the entire skycell.

stsci-hack (changes requested):
Final note:  this code can only be successfully used on MVM products generated by MVM processing that was done with diagnostic mode turned on.  This requirement stems from the recent change in MVM processing to delete unmodified FLT/FLC files from being passed on as an output product after renaming them as skycell inputs and from the fact that the source identification routine used by this code relies on having those products available.  This is not a problem, per se, with this PR, but rather the code it relies on.  As such, this PR can be merged and this interaction can be improved with another PR.   **Once the current set of comments have been addressed, this PR can formally approved for merging.**

drizzlepac/devutils/analyze_mvm_gaia_alignment.py
dulude: Issue resolved by commit [094a1cd](https://github.com/spacetelescope/drizzlepac/pull/1241/commits/094a1cda714ccb414cc2f7c5d50bd96650a0866b)

drizzlepac/devutils/analyze_mvm_gaia_alignment.py
dulude: issue resolved by commit [819e7f0](https://github.com/spacetelescope/drizzlepac/pull/1241/commits/819e7f06bfd0e684d383ba0c216da21a5608e92b)

stsci-hack (approved):
LGTM.

PR #1244, opened 2022-01-17
Build process

nden:
Closes #1248 

This PR attempts to streamline and modernize drizzlepac's build process.
Currently it uses three different build systems which interact in bad ways leading to problems like the ones described in SPB-838 and SPB-826.

- Removed relic
- removed distutils
- moved dependencies to setup.cfg so they can be used by pip
- using setuptools_scm and its way to create `__version__.py` from tags
- removed numpy as an install dependency from setup since it's defined asa build dependencyy in pyproject.toml
- Since astropy is constrained in setup, also constrained it in pyproject.toml, otherwise the resolver goes into a very long loop of installing different astropy versions. (Consider removing it from setup.)
- Removed all `__version_date__` strings

Questions:

- Should installs with python< 3.8 be allowed?
- Needs someone to test it,  ~especially the numerous `__version_date__` strings, some of which I may have dropped.~

PR #1245, opened 2022-01-18
Remove nictools as a dependency

nden:
`nictools` does not appear to be maintained, removing it makes drizzlepac's build system just a bit more robust.
Copied the code fro nictools to drizzlepac/nicmosData.py.

Should be merged after #1244.

### DIFF
--- a/drizzlepac/nicmosData.py
+++ b/drizzlepac/nicmosData.py
@@ -6,10 +6,12 @@ Class used to model NICMOS specific instrument data.
 :License: :doc:`LICENSE`
 
 """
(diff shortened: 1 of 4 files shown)


stsci-hack (dismissed):
LGTM, as far as the changes themselves go.  However, an explanation should be provided for why the tests are now failing for this PR, as it may indicate an interaction that needs to be resolved.

stsci-hack (approved):
I see no problems with pulling this code directly into this package or with how it was done.  LGTM.

PR #1246, opened 2022-01-18
RickerWavelet Kernel for SBC needs to have dimensions which are odd

mdlpstsci:
RickerWavelet Kernel for SBC needs to have dimensions which are odd.
Changed to 23 from 24.  Tested with j8ep07.


stsci-hack (approved):
Simple enough change.  LGTM.

PR #1247, opened 2022-01-19
Address issue of logging for PyTest tests

mdlpstsci:
Modified the align.py module, perform_align() and determine_fit_quality(), by defining global variables to use for the log file handler and log filename.  This was done to address the PyTestsmfrom having an error looking for a directory associated with the first test as all PyTests are done in a single Python session.  The error would cause the test to fail, and the tests need to be executed to ensure the algorithms continue to work as expected.  Essentially, the log file handler is reset as necessary so a fresh handle is assured for each new test.  Also, some PyTests were updated to PASS.  However, updating the tests further, particularly to remove "skip", is a topic for a different PR.


drizzlepac/haputils/align_utils.py
```diff
@@ -1222,9 +1222,14 @@ def update_image_wcs_info(tweakwcs_output, headerlet_filenames=None, fit_label=N
...
-            for sci_ext_ctr in range(1, num_sci_ext + 1):
-                sci_ext_dict["{}".format(sci_ext_ctr)] = fileutil.findExtname(hdulist, 'sci', extver=sci_ext_ctr)
+            try:
+                sci_ext_dict = {}
+                for sci_ext_ctr in range(1, num_sci_ext + 1):
+                    sci_ext_dict["{}".format(sci_ext_ctr)] = fileutil.findExtname(hdulist, 'sci', extver=sci_ext_ctr)
+            except Exception:
+                exc_type, exc_value, exc_tb = sys.exc_info()
+                traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
+                logging.exception("message")
```
stsci-hack, line 1232: I am sure I don't follow everything that is going on here.  Is 'logging.exception' reporting the exception with the explanation of 'message'?  If so, should 'message' be replaced with something more explanatory?  Just checking...

drizzlepac/haputils/product.py
```diff
@@ -288,13 +288,21 @@ def align_to_gaia(self, catalog_list=[], output=True, process_type='SVM',
...
                                                                                                        more_catalogs,
                                                                                                        num_cat,
                                                                                                        alignment_pars,
                                                                                                        print_fit_parameters=True,
-                                                                                                       loglevel=self.log_level)
+                                                                                                       loglevel=self.log_level,
+                                                                                                       runfile=log_file)
+                                
+                                # Clean up the temporary log file as the contents are captured
+                                os.remove(log_file)
```
stsci-hack, line 305: Be careful with the use of 'os.remove' as this can fail under Windows but work perfectly well for the same data under Linux or Mac.  As a result, you may want to trap the OSError exception specifically here and report it separately from other possible reasons for failure at this point.

drizzlepac/haputils/align_utils.py
mdlpstsci: The code looks misleading, but here is a simple example to show what would be printed.  The "traceback.print_exception() generates the first 3 lines.  The logging.exception("message") generates the output starting from "ERROR:root:message".  In this instance (perhaps in all instances), it is not necessary to have both "print" statements.

import sys, traceback, logging
try:
    1/0
except Exception:
    exc_type, exc_value, exc_tb = sys.exc_info()
    traceback.print_exception(exc_type, exc_value, exc_tb, file=sys.stdout)
    logging.exception("message")

  File "<stdin>", line 2, in <module>
ZeroDivisionError: division by zero
ERROR:root:message
  File "<stdin>", line 2, in <module>
ZeroDivisionError: division by `zero``

tests/hap/test_svm_hrcsbc.py
```diff
@@ -198,32 +201,43 @@ def test_svm_empty_cats(gather_output_data):
...
+    num_sources = {tdp:len(ascii.read(tdp, format="ecsv")) for tdp in tdp_files}
     valid_cats = {}
-    for cat in expected_point_sources.keys():
-        for file in cat_files:
-            if cat in file and "total" in file:
-                valid_cats[cat] = (np.isclose(num_sources[file], expected_point_sources[cat], rtol=0.1), num_sources[file])
+    for tdp in expected_total_point_sources.keys():
+        for file in tdp_files:
+            if tdp in file:
+                tol_limit = tolerance * expected_total_point_sources[tdp]
```
stsci-hack, line 218: It would be more efficient to simply rely on numpy's built-in comparison code to do this; specifically, np.isclose.  The 'tolerance' specified here corresponds to the 'rtol' parameter for 'np.isclose' or perhaps simply use 'tol_limit' as the value for the 'atol' parameter instead (may be more explicit).  Just an observation...

stsci-hack (dismissed):
Everything looks to be fine, but a couple of items did stand out for comment.  Should those comments not require any changes, then this can be merged.

drizzlepac/haputils/product.py
mdlpstsci: Modified to be:
```
# Clean up the temporary log file as the contents are captured
try:
    os.remove(log_file)
except OSError as error:
l   log.warning("Unable to remove file {}.".format(log_file))
    log.warning("Output trailer files may contain duplicate information.")
```

drizzlepac/haputils/align_utils.py
stsci-hack: Sounds good to me.

drizzlepac/haputils/product.py
stsci-hack: Nice log messages.

stsci-hack (approved):
LGTM.

PR #1250, opened 2022-01-25
Remove logic from align

stsci-hack:
This simple change addresses the bug reported in HLA-689 that related to checking for alignment results in align.py when it was apparently not necessary.   Removing this logic should allow more data to successfully align to GAIA.


mdlpstsci (approved):
Solution looks fine.

PR #1255, opened 2022-01-31
Insure focus comparison works on Linux

stsci-hack:
This simple fix insures that 'nan' values in the drizzle products do not show up when performing the focus comparisons.  When 'nan' values are present, the max focus value can not be determined and it causes an Exception to be thrown.

### DIFF
--- a/drizzlepac/haputils/astrometric_utils.py
+++ b/drizzlepac/haputils/astrometric_utils.py
@@ -2104,6 +2104,7 @@ def build_focus_dict(singlefiles, prodfile, sigma=2.0):
     for f in singlefiles:
         imgarr = fits.getdata(f)
         imgarr[~full_sat_mask] = 0
+        imgarr = np.nan_to_num(imgarr, 0)  # Insure, on Linux, all 'nan' values are removed
         focus_val, focus_pos = determine_focus_index(imgarr, sigma=sigma)
         focus_dict['exp'].append(float(focus_val))
         focus_dict['exp_pos'] = (int(focus_pos[0][0]), int(focus_pos[1][0]))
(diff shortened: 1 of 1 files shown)


stsci-hack:
This problem was identified as a result of pipeline regression testing using ASN 'ibs335030'.

drizzlepac/haputils/astrometric_utils.py
```diff
@@ -2104,13 +2104,15 @@ def build_focus_dict(singlefiles, prodfile, sigma=2.0):
     for f in singlefiles:
         imgarr = fits.getdata(f)
         imgarr[~full_sat_mask] = 0
+        imgarr = np.nan_to_num(imgarr, 0)  # Insure, on Linux, all 'nan' values are removed
```
mdlpstsci, line 2107: Question as this always throws me off.  The signature for nan_to_num is 
numpy.nan_to_num(x, copy=True, nan=0.0, posinf=None, neginf=None).  So does your "0" map to copy as False?  Is this what you intend?  I have the same question for line 2115.

mdlpstsci (changes requested):
Actually, this is a request for clarification.

drizzlepac/haputils/astrometric_utils.py
stsci-hack: I updated the calls in [77449e2](https://github.com/spacetelescope/drizzlepac/pull/1255/commits/77449e27c494b9b14d9e49d7e5dae314c9803fe1) to that function to explicitly set 'nan=0' to be perfectly clear with how this function should be used in this instance.

mdlpstsci (approved):
Thank you.  I think setting the value explicitly is best.

PR #1256, opened 2022-02-01
Remove extraneous version variables

stsci-hack:
These changes follow up on #1244 by removing module-level definitions of `__version__` where that variable is not used for logging and replaces it with the package definition of that variable for modules that do report it for logging.


drizzlepac/mapreg.py
```diff
@@ -16,10 +16,9 @@
 from stsci.tools import teal
 from . import util
 from .regfilter import fast_filter_outer_regions
+from .version import __version__
 
 
-# This is specifically NOT intended to match the package-wide version information.
-__version__ = '0.1'
 __taskname__ = 'mapreg'
 __author__ = 'Mihai Cara'
```
mdlpstsci, line 24: This module leaves __taskname__ and __author__ intact, but photeq.py removes these entries.  Are we removing all the dunders?  We should be consistent, unless there is some rationale that I am missing.

mdlpstsci (commented):
I do not disagree with any of these changes, but I do have a question.  How is it decided which modules contain "from .version import dunder version"?  I cannot empirically determine the decision process.

stsci-hack:
Typically, only those modules which report the version in log messages, or modules which **used** to include the version in log messages, will include this import statement.  Otherwise, it may have been included in preparation for such log messages and may not have been used.  I tried to clean this up some in this PR, but probably was not completely thorough.

mdlpstsci (approved):

PR #1262, opened 2022-02-07
Improve source detection for SBC data 

stsci-hack:
A custom algorithm has been implemented specifically for use in identifying sources in SBC data for the purposes of alignment.  This algorithm takes advantage of the fact that SBC data has no cosmic-rays by applying the Laplacian-of-Gaussian operator to the image and selecting the subsequent peaks as sources.  This results in a vast majority of the actual sources getting identified without relying on comparisons to PSFs (model or empirical).  Increasing the number of valid sources identified in SBC exposures provides an opportunity for the alignment algorithm to find matches to sources from astrometric catalogs, while passband differences will still result in confusion in cross-matching leading to failures to align the exposure.  Still, these changes can only benefit alignment of SBC data.

### DIFF
--- a/drizzlepac/haputils/align_utils.py
+++ b/drizzlepac/haputils/align_utils.py
@@ -20,7 +20,7 @@ from astropy.coordinates import SkyCoord, Angle
 from astropy import units as u
 from astropy.modeling import models, fitting
 
-from photutils import background
+from photutils import background, DAOStarFinder
 from photutils.background import Background2D
 from photutils.utils import NoDetectionsWarning
 
@@ -35,6 +35,7 @@ from stsci.tools import fileutil
 from .. import updatehdr
 from . import astrometric_utils as amutils
 from . import analyze
+from . import deconvolve_utils as decutils
 
 import tweakwcs
 
@@ -677,7 +678,6 @@ class HAPImage:
 
         return dqmask
 
-
     def find_alignment_sources(self, output=True, dqname='DQ', crclean=False,
                                **alignment_pars):
         """Find sources in all chips for this exposure."""
@@ -733,13 +733,11 @@ class HAPImage:
 class SBCHAPImage(HAPImage):
     def __init__(self, filename):
         super().__init__(filename)
+        self.exptime = max(self.imghdu[0].header['exptime'], 1.0)
 
     def find_alignment_sources(self, output=True, dqname='DQ', crclean=False,
                                **alignment_pars):
         """Find sources in all chips for this exposure."""
-        if crclean:
-            self.imghdu = fits.open(self.imgname, mode='update')
-
         # Only 1 chip, no need to loop
         chip = 1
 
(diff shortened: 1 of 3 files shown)


stsci-hack:
These changes were developed and tested using data from ASN 'j9ou55010' and 'jc6e03ufq'.

drizzlepac/haputils/align_utils.py
```diff
@@ -725,6 +730,87 @@ def find_alignment_sources(self, output=True, dqname='DQ', crclean=False,
...
+        else:
+            outroot = None
+
+        sciarr = self.imghdu[("SCI", chip)].data.copy()
+
+        # Remove all background noise
+        # This background noise is effectively integerized by the SBC detector
+        sci_gauss = ndimage.gaussian_filter(sciarr, sigma=3.0)
+        bkg_noise = astropy.stats.sigma_clipped_stats(sci_gauss, maxiters=1)
+        # bkg_noise = astropy.stats.sigma_clipped_stats(sciarr[sciarr > 0], maxiters=1)
```
mdlpstsci, line 756: Since bkg_noise is really three variables, I would add a comment to make this clear (e.g., mean, median, stddev) at Line 755.

drizzlepac/haputils/align_utils.py
```diff
@@ -725,6 +730,87 @@ def find_alignment_sources(self, output=True, dqname='DQ', crclean=False,
...
+        bkg_limit = bkg_noise[1] + 3 * bkg_noise[2]  # Set limit as 1-sigma above mean
+        log.debug(f"BKG_LIMIT for source identification: {bkg_limit} based on {bkg_noise}")
```
mdlpstsci, line 758: Fix the comment at Line 757 to be generic (n sigma above median).

drizzlepac/haputils/align_utils.py
```diff
@@ -725,6 +730,87 @@ def find_alignment_sources(self, output=True, dqname='DQ', crclean=False,
...
+        # create mask of all background pixels and zero out all those pixels
+        bkg_mask = sci_gauss <= bkg_limit
+        sci_bkgsub = sciarr.copy()
+        sci_bkgsub[bkg_mask] = 0
+
+        # A high LoG sigma value is needed to smooth over the pixelated nature of the SBC PSFs
+        # This avoids having PSF halo peaks being detected as 'real' sources
+        slabels, snum, logimg = amutils.detect_point_sources(sci_bkgsub, background=0, log_sigma=5.0)
+        # create mask based on these sources only...
```
mdlpstsci, line 767: Line 766. If "logimg" is really not ever used, replace with "_".

mdlpstsci (changes requested):
Just a few minor tweaks requested.

drizzlepac/haputils/align_utils.py
stsci-hack: Fixed in 7b869e6.

drizzlepac/haputils/align_utils.py
stsci-hack: Fixed in 7b869e6.

drizzlepac/haputils/align_utils.py
stsci-hack: Fixed in 7b869e6.

PR #1264, opened 2022-02-10
Update toml file to for NumPy support

mdlpstsci:
Modified toml specifically to use "oldest-supported-numpy".


stsci-hack (approved):
Just what we talked about... LGTM

PR #1265, opened 2022-02-10
Generate correct HDRNAME (finally?)

stsci-hack:
This simple change corrects the use of 'rstrip' in the generation of the HDRNAME keyword so that it comes out as expected.  

Using '.rstrip(".fits")' on a full filename results in chopping the last character of the filename before the ".fits" (don't ask me why!).  However, using '.rstrip("fits")' leaves the '.' from the full filename at the end of the string.  This simply uses that behavior to correctly strip the '.fits' from a full filename and not chop the 'flt' or 'flc' to be 'fl'.


mdlpstsci (approved):
Interesting (and unexpected) behavior is now fixed!

Issue #1260, opened 2022-02-02
Update/fix HAP code "exit" statement handling


stscijgbot-hstdp:
_Issue [HLA-712](https://jira.stsci.edu/browse/HLA-712) was created on JIRA by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude):_

# Make sure all relevant "sys.exit()" statements return the appropriate return value
 # 
hapsequencer.py lacks the "except BaseException" lines of code that hapmultisequencer.py has around line 480. This will have to be added.

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-712?focusedCommentId=568906#comment-568906):_

<https://www.evernote.com/shard/s306/sh/ad89ef22-48c9-fc6b-ecfe-9079fc0586dd/6644556d05bec3fe14bb89cf0a6a4477](https://www.evernote.com/shard/s306/sh/ad89ef22-48c9-fc6b-ecfe-9079fc0586dd/6644556d05bec3fe14bb89cf0a6a4477)>

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-712?focusedCommentId=569028#comment-569028):_

sys.exit() calls that require revision because they don't currently return an int value:
 * devutils/comparison_tools/compare_sourcelists.py:1618:                sys.exit("ERROR: Unrecognized format. Exiting...")
 * devutils/comparison_tools/compare_sourcelists.py:1638:            sys.exit("ERROR: Unrecognized format. Exiting...")
 * haputils/comparison_utils.py:414:                sys.exit("ERROR: Unrecognized format. Exiting...")
 * haputils/comparison_utils.py:434:            sys.exit("ERROR: Unrecognized format. Exiting...")
 * haputils/config_utils.py:52:            sys.exit("ERROR: Input and output parameter files must have unique names!")
 * haputils/config_utils.py:170:                        sys.exit("INVALID ACS DETECTOR!")
 * haputils/config_utils.py:204:                        sys.exit("INVALID WFC3 DETECTOR!")
 * haputils/config_utils.py:206:                    sys.exit("INVALID HST INSTRUMENT!")

[...]

Issue #1212, opened 2021-12-08
Check the HAP code "exit" statements


stscijgbot-hstdp:
_Issue [HLA-677](https://jira.stsci.edu/browse/HLA-677) was created on JIRA by [Michele De La Pena](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mdelapena):_

All *exit* statements in the HAP code should be checked to ensure the proper choice of *exit* has been used, **and** an appropriate exit code has been provided.  The choices are at least: SystemExit(), sys.exit(), exit()/quit(), os.exit().  *_  All of these, except os._exit(), raise SystemExit and will be caught by _except BaseException*.

There should be a discussion of agreed upon return codes between developers and operations.

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-677?focusedCommentId=566298#comment-566298):_

Note on the difference between exception and BaseException:
- "except Exception:" catches "raise" calls
- "except BaseException:" catches **only** sys.exit() calls and the like that execute a full exit.

See this [StackOverflow thread](https://stackoverflow.com/questions/7160983/catching-all-exceptions-in-python) for more details

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-677?focusedCommentId=566518#comment-566518):_

Examples of correct syntax:
 - Towards the end of hapmultisequencer.run_mvm_processing(), around line #&#x2060;480

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-677?focusedCommentId=566519#comment-566519):_

<https://www.evernote.com/shard/s306/sh/ad89ef22-48c9-fc6b-ecfe-9079fc0586dd/6644556d05bec3fe14bb89cf0a6a4477>

stscijgbot-hstdp:
[...]

PR #1267, opened 2022-02-15
SVM/MVM: Update/fix HAP code "exit" statement handling

dulude:
**Relevant Issues and Tickets**
- [HLA-677](https://jira.stsci.edu/browse/HLA-677)
- Issue #1212
- [HLA-712](https://jira.stsci.edu/browse/HLA-712)
- Issue #1260 

**Summary of Changes**
- The main goal here was to standardize execution of the sys.exit() command by any and all code ultimately run by either the SVM or MVM pipeline.
- At the site of any sys.exit() command in  any code that is ultimately run by either the SVM or MVM pipeline I...
      - ensured that there was a descriptive message in a log.error() statement
      - ensured that the sys.exit() command returned an integer value, which was typically 1, but could also be 0, 55 or 65 in specific cases.
 - A small section of code was copied from hapmultisequencer.py to  hapsequencer.py to handle BaseExceptions raised when a sys.exit() command is executed.

### DIFF
--- a/drizzlepac/haputils/config_utils.py
+++ b/drizzlepac/haputils/config_utils.py
@@ -115,7 +115,7 @@ class HapConfig(object):
             if hasattr(prod_obj, "skycell"):
                 n_exp = int(round(prod_obj.mask_kws['MEANNEXP'][0]))  # Use mean number of exposures in skycell instead of just simple number of input images for MVM
             else:
-                n_exp = len(prod_obj.edp_list) # use simple number of input images for SVM
+                n_exp = len(prod_obj.edp_list)  # use simple number of input images for SVM
 
         # determine product type, initialize and build conditions list
         if hasattr(prod_obj, "edp_list") and hasattr(prod_obj, "fdp_list"):  # For total products
@@ -168,7 +168,7 @@ class HapConfig(object):
                         if n_exp >= 6:
                             self.conditions.append("acs_wfc_any_n6")
                     else:
-                        log.error("INVALID ACS DETECTOR!")
+                        log.error("{} is an invalid ACS detector!".format(self.detector))
                         sys.exit(1)
                 elif self.instrument == "wfc3":
                     if self.detector == "ir":
@@ -203,10 +203,10 @@ class HapConfig(object):
                             if n_exp >= 6:
                                 self.conditions.append("wfc3_uvis_any_pre_n6")
                     else:
-                        log.error("INVALID WFC3 DETECTOR!")
+                        log.error("{} is an invalid WFC3 detector!".format(self.detector))
                         sys.exit(1)
                 else:
-                    log.error("INVALID HST INSTRUMENT!")
+                    log.error("{} is an invalid HST instrument!".format(self.instrument))
                     sys.exit(1)
         else:  # For single-exposure products
             self.conditions = ["single_basic"]
--- a/drizzlepac/haputils/config_utils.py
+++ b/drizzlepac/haputils/config_utils.py
@@ -49,7 +49,8 @@ class HapConfig(object):
         """
         log.setLevel(log_level)
         if input_custom_pars_file and input_custom_pars_file and input_custom_pars_file == output_custom_pars_file:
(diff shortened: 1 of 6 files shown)


drizzlepac/haputils/config_utils.py
```diff
@@ -167,7 +168,8 @@ def _determine_conditions(self, prod_obj):
                         if n_exp >= 6:
                             self.conditions.append("acs_wfc_any_n6")
                     else:
-                        sys.exit("INVALID ACS DETECTOR!")
+                        log.error("INVALID ACS DETECTOR!")
```
mdlpstsci, line 171: For messages like this, I have been trying to print the information the program currently has in memory in case there is something amiss.  I suggest printing out the value of the variable (e.g., self.detector).  This comment should be applied to other similar statements.

drizzlepac/haputils/get_git_rev_info.py
```diff
@@ -108,10 +107,10 @@ def get_rev_id(local_repo_path):
             else:
                 raise ValueError("Git revision information not found.")
     except Exception:
-        pass
+        log.warning("Problem encountered getting git revision ID")
+        sys.exit(1)
```
mdlpstsci, line 111: Since this Exception has been promoted to be more serious, I have to ask if this failure serious enough to stop all processing and exit?  Shall we (Warren, Mike, Michele) discuss this?

mdlpstsci (changes requested):
Just a few comments and a question.

drizzlepac/haputils/config_utils.py
dulude: Issue resolved by commit [fafc9d2](https://github.com/spacetelescope/drizzlepac/pull/1267/commits/fafc9d20e0626b42eb8c6356308f374d1cf93761)

drizzlepac/haputils/get_git_rev_info.py
dulude: So the interesting thing here is that the git revision id isn't reported by this piece of code until after the SVM code as basically finished processing, and not even directly by either pipeline. it's actually invoked during by the SVM QA test that compares HAP-created sourcelists to their HLA classic counterparts, so it's a very very very specific failure mode that won't occur for observations that weren't processed by HLA, or if a certain environment variable exists and is set to "True". So for all these reasons, you are proably right, @mdlpstsci that the severity on the logging statement in question could be knocked down a notch from "error" to "warning".
All of this sort of begs the question though...is the git revision ID reported earlier in the code in a place that it will always be logged? this seems like a good thing to have in either the log files or somewhere in the headers of the output products.

mdlpstsci: > @dulude I agree. The Git tag is put in the output log of the HSTCAL processing (calxxx) pipelines. I do think this is extremely useful to do for the HAP processing, but this is an issue for discussion and a different PR!

mdlpstsci: @dulude I agree. The Git tag is put in the output log of the HSTCAL processing (calxxx) pipelines. I do think this is extremely useful to do for the HAP processing, but this is an issue for discussion and a different PR!

dulude: ok. I'll pull out the 'sys.exit()' and decrease the severity of the logging statement from "error" to "warning".

mdlpstsci (approved):
Cool.

drizzlepac/haputils/get_git_rev_info.py
dulude: I created Ticket [HLA-718](https://jira.stsci.edu/browse/HLA-718) to deal with this problem as per your suggestion.

dulude: Issue resolved by commit [9a25d71](https://github.com/spacetelescope/drizzlepac/pull/1267/commits/9a25d7117646588b44d958d25db2572acdf0887d)

PR #1273, opened 2022-02-21
Figures for the MVM utilities descriptions

mdlpstsci:
These are the four figures which were inadvertently left out when the MVM utilities descriptions were committed.


stsci-hack (approved):
Nice figures to have.

PR #1274, opened 2022-02-22
Modified three tests to remove the "skip" to enable the tests to execute

mdlpstsci:
Remove the "skip" to enable the tests to execute.  PyTest criterion may have been modified to make the test succeed or be more general.

### DIFF
--- a/tests/hap/test_run_svmpoller.py
+++ b/tests/hap/test_run_svmpoller.py
@@ -15,7 +15,6 @@ from astropy.io import ascii
 from drizzlepac import runsinglehap
 from astropy.table import Table
 from drizzlepac.haputils import astroquery_utils as aqutils
-from importlib import reload
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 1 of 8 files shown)


tests/hap/test_run_svmpoller.py
```diff
@@ -60,29 +54,34 @@ def test_run_svmpoller(tmpdir, dataset):
...
     print("TEST_RUN_SVMPOLLER. Dataset: ", dataset)
 
     current_dt = datetime.datetime.now()
     print(str(current_dt))
 
+    # Create working directory specified for the test
+    #install_dir = os.path.dirname(__file__)
+    #local_path = os.path.join(install_dir, dataset)
+    #os.chdir(install_dir)
+
```
stsci-hack, line 84: Code that has been commented out can be simply deleted if it is not needed for the final working version.

stsci-hack (changes requested):
Just a couple of things stood out to be addressed.

tests/hap/test_run_svmpoller.py
```diff
@@ -3,55 +3,49 @@
...
-from astropy.table import Table
 
+from astropy.io import ascii
 from drizzlepac import runsinglehap
-
-
-log = logutil.create_logger('test_run_svmpoller', level=logutil.logging.INFO, stream=sys.stdout)
+from astropy.table import Table
+from drizzlepac.haputils import astroquery_utils as aqutils
+from importlib import reload
```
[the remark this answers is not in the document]
mdlpstsci: Fixed in e29c47d.

tests/hap/test_run_svmpoller.py
mdlpstsci: Fixed in e29c47d.

stsci-hack (approved):
Looks fine to me now.

PR #1278, opened 2022-02-24
Protect actions being performed on a potentially NoneType object

mdlpstsci:
Update to handle the situation where the one chip has sources, but the other chip has no sources, and the data is stored in a dictionary. The "table" is actually a dictionary so the len function on len(table[chipnum]) where table[chipnum] is NoneType caused an exception.


stsci-hack (approved):
Makes sense.

PR #1279, opened 2022-02-24
Generate SVM QA JSONs again

stsci-hack:
The JSON files now can be generated for SVM processing with **'SVM_QUALITY_TESTING'** turned on as a result of these changes.  Updating _photutils_ to v1.2.0 resulted in the table of point sources to be returned as a **QTable** instead of a simpler **Table** object, and the **QTable** object can not (at this time anyway) be serialized as a **JSON** object.  Casting the output from _photutils_ to a simpler **Table** object fixes this problem with no consequences.  This allows the SVM testing using the _'svm_quality_analysis'_ module to succeed regardless of the version of _photutils_.  

In addition, a couple of changes were implemented to allow this code to be more platform independent; specifically, the git information can now be reported successfully even when run on Windows.  

These changes were tested under both Windows 10 and Linux on data from visit 'iacs01'.

### DIFF
--- a/drizzlepac/haputils/get_git_rev_info.py
+++ b/drizzlepac/haputils/get_git_rev_info.py
@@ -66,7 +66,7 @@ def print_rev_id(local_repo_path):
             # for Windows, 'head' command doesn't exist
             instream = os.popen('git log --oneline |more')
             for line in instream:
-                log.info("{}\n".format(instream[0].strip()))
+                log.info("{}\n".format(line.strip()))
                 # Only print out first entry, then quit
                 break
         else:
--- a/drizzlepac/haputils/get_git_rev_info.py
+++ b/drizzlepac/haputils/get_git_rev_info.py
@@ -62,9 +62,17 @@ def print_rev_id(local_repo_path):
             log.info("{}".format(streamline.strip()))
 
         log.info("\n===== Most Recent Commit =====")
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/get_git_rev_info.py
```diff
@@ -62,9 +62,17 @@ def print_rev_id(local_repo_path):
...
 
         log.info("\n===== Most Recent Commit =====")
-        instream = os.popen('git log |head -1')
-        for streamline in instream:
-            log.info("{}\n".format(streamline.strip()))
+        if 'win' in sys.platform:
+            # for Windows, 'head' command doesn't exist
+            instream = os.popen('git log --oneline |more')
+            for line in instream:
+                log.info("{}\n".format(instream[0].strip()))
```
mdlpstsci, line 69: Sorry to be dense, but why is this "for" loop necessary as "line" is never used/

mdlpstsci (changes requested):
This is either just an answer to a question or a request for a change!

drizzlepac/haputils/get_git_rev_info.py
stsci-hack: Good catch.  Fixed in 9fb037a.

mdlpstsci (approved):

PR #1280, opened 2022-02-25
Update requirements to allow astropy 5.0 use

stsci-hack:
This simply updates the dependency requirements for the package to allow use of the latest version of astropy; namely, astropy 5.0.  The addition of the 'typing_extensions' dependency is needed since astropy 5.0.1 does not include this itself, as it was only added to astropy installation with [Astropy PR #12812](https://github.com/astropy/astropy/pull/12812) after 5.0.1 was released.  

The results of using astropy 5.0.1 are identical to using astropy 4.3.post1 based on comparisons of data from visit 'iacs01' along with running the 'hap' pytests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools>=42.0",
             "setuptools_scm[toml]>=3.4",
             "wheel",
             "oldest-supported-numpy",
-            "astropy<5.0"]
+            "astropy>=4.3"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools_scm]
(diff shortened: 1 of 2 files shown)


nden:
@stsci-hack You need to change also pyproject.toml or at some point there will be version mismatches.

mdlpstsci (commented):
I am just double-checking...is there a change needed to the pyproject.toml file?

nden:
@mdlpstsci Not sure whom you are asking but yes, the version restriction in pyproject.toml needs to be removed.

mdlpstsci (approved):

nden (approved):
Before youmake a PyPi release, it may be good to run a quick test with all packages.

PR #1285, opened 2022-03-02
Reset tasknames to work with TEAL

stsci-hack:
These changes reset the values of the `__taskname__` variable for each task/module in drizzlepac so that code which calls teal to get the default parameters using these tasknames will actually work again.  This will enable many of the pytest tests to also run successfully again as well.

### DIFF
--- a/drizzlepac/ablot.py
+++ b/drizzlepac/ablot.py
@@ -31,7 +31,7 @@ from . import __version__
 
 __all__ = ['blot', 'runBlot', 'help', 'getHelpAsString']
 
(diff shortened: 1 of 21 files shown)


mdlpstsci:
@stsci-hack There are errors collecting the tests.

stsci-hack:
The PR was updated in 967381c to fix the problems preventing the tests to run.

mdlpstsci (approved):
Yay -- the changes are clean, and the tests succeeded.

PR #1292, opened 2022-03-03
Fix problem with astropy 5.0 table interpretation

stsci-hack:
These changes implement the code necessary for correctly interpreting binary tables with astropy regardless of the astropy version.  Astropy versions < 5.0 returned binary table data as byte arrays, while versions > 5 did not.  This code interprets that table data correctly in either case.  This now allows `astrodrizzle.AstroDrizzle()` to be run directly from python scripts (or command line) when astropy >=5.0 is installed as tested using data from ASN `ibvz01020`.

### DIFF
--- a/drizzlepac/processInput.py
+++ b/drizzlepac/processInput.py
@@ -33,6 +33,7 @@ import os
 import shutil
 import string
 import sys
+from packaging.version import Version
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


mdlpstsci (approved):
Glad this was a straightforward fix!

PR #1295, opened 2022-03-03
Protect computations when photflam is equal to 0.0

mdlpstsci:
When photflam == 0.0, the magnitude, magnitude error, and CI are set to -9999.0.  Reworked the computation of CI for both Point and Segment catalogs.  Also, ensure "ci_data" and "ci_mask" are created even if an exception is thrown.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -33,6 +33,7 @@ from stsci.tools import logutil
 from stwcs.wcsutil import HSTWCS
 
 from . import astrometric_utils
+from . import constants
 from . import photometry_tools
 from . import deconvolve_utils as decutils
 from . import processing_utils as proc_utils
@@ -1194,11 +1195,11 @@ class HAPPointCatalog(HAPCatalogBase):
 
         try:
             # Calculate and add concentration index (CI) column to table
-            if math.isclose(self.image.keyword_dict['photflam'], 0.0, abs_tol=1e-9):
+            if math.isclose(self.image.keyword_dict['photflam'], 0.0, abs_tol=constants.TOLERANCE):
                 # Fill the ci_data column with the "bad data indicator" of -9999.0
                 # where photometry_tbl["MagAp1"] was populated with -9999.0 in iraf_style_photometry
                 ci_data = photometry_tbl["MagAp1"].data
-                ci_mask = np.logical_not(ci_data > -9999.0)
+                ci_mask = np.logical_not(ci_data > constants.FLAG + 1.0)
             else:
                 ci_data = photometry_tbl["MagAp1"].data - photometry_tbl["MagAp2"].data
                 ci_mask = np.logical_and(np.abs(ci_data) > 0.0, np.abs(ci_data) < 1.0e-30)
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -1194,17 +1194,24 @@ def measure_sources(self, filter_name):
...
+            else:
+                ci_data = photometry_tbl["MagAp1"].data - photometry_tbl["MagAp2"].data
+                ci_mask = np.logical_and(np.abs(ci_data) > 0.0, np.abs(ci_data) < 1.0e-30)
+                big_bad_index = np.where(abs(ci_data) > 1.0e20)
+                ci_mask[big_bad_index] = True
+
+        except Exception as x_cept:
+            log.warning("Computation of concentration index (CI) was not successful: {} - {}.".format(self.imgname, x_cept))
+            log.warning("CI measurements may be missing from the output filter catalog.\n")
+            ci_data = -9999.0 + photometry_tbl["MagAp1"] * 0.0
```
stsci-hack, line 1211: Why not simply 'ci_data = -9999.0', since anything times zero is zero?

drizzlepac/haputils/photometry_tools.py
```diff
@@ -218,13 +223,17 @@ def convert_flux_to_abmag(in_flux, photflam, photplam):
         input flux values converted to ABMAG
     """
 
-    # convert flux from units of electrons/second to ergs/cm2/angstrom/second
-    f_lambda = in_flux * photflam
+    # Avoid taking the log10 of zero
+    if math.isclose(photflam, 0.0, abs_tol=1e-9):
```
stsci-hack, line 227: Typical values of PHOTFLAM are on the order of 1e-16 to 1e-20 or less as reported in the WFC3 Photometric Zero Point page.  In which case, the `abs_tol` value is too large and would always flag any value of PHOTFLAM as `identical` to zero.

stsci-hack (changes requested):
Please reset tolerances on 'isclose' to better match the actual values of PHOTFLAM.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: I need the result to be a column and not a scalar.   I thought this would ensure I got a column as this is what the remainder of the code is expecting.

drizzlepac/haputils/photometry_tools.py
mdlpstsci: This must be my issue.

drizzlepac/haputils/catalog_utils.py
stsci-hack: That makes sense.  A comment to the effect would prevent confusion in the future.

mdlpstsci: Addressed in 3a0707e.

drizzlepac/haputils/photometry_tools.py
mdlpstsci: Addressed in 3a0707e.

stsci-hack (approved):
This solution works nicely.  Now we have to move other constants to that module.

PR #1297, opened 2022-03-04
Remove IPPPSSOO keyword from MVM product headers (again)

stsci-hack:
This corrects an oversight in the drizzling of MVM products, where the rules file intended for generating the MVM products was not being explicitly passed to AstroDrizzle.  The changes here now insures that the correct rules file gets used so that the IPPPSSOO keyword no longer appears in ANY MVM product header.


mdlpstsci (approved):
Looks fine.

PR #1300, opened 2022-03-08
Statistics for SVM and MVM 

mdlpstsci:
This ticket addresses a majority of HLA-338 HLA-687.  

Also, fixed a bug which manifested itself by inserting incorrect statistics into each SkyCell layer.  The statistics for the last layer processed were used for all layers.  The problem was a shallow copy of the default dictionary was used to initialize the object attribute when a deepcopy was needed.  The stats are in SCI header only.

WHT statistics.


stsci-hack (approved):
Simple enough. Good catch!

PR #1306, opened 2022-03-14
Fix setup.py

mdlpstsci:
Fix setup.py to include json files which reside at the same level as [mvm|svm]_parameters.


stsci-hack (approved):
Makes sense to me.

PR #1307, opened 2022-03-14
Explicitly update boolean column in ASN tables

stsci-hack:
These changes force boolean columns of ASN tables to be explicitly copied over with the exact values as the input ASN table when creating copies for pipeline processing in `runastrodriz`.  This corrects the problem with using `stsci.tools.asnutil.readAsnTable(asnfile, prodonly=True)` and not getting any table data as seen in testing.

### DIFF
--- a/drizzlepac/processInput.py
+++ b/drizzlepac/processInput.py
@@ -768,6 +768,7 @@ def changeSuffixinASN(asnfile, suffix):
     fasn = fits.open(_new_asn, mode='update', memmap=False)
     fasn[0].header['DUPCHECK'] = "COMPLETE"
     newdata = fasn[1].data.tolist()
+
     for i in range(len(newdata)):
         if asn_decode:
             val = newdata[i][0].decode(encoding='UTF-8').strip()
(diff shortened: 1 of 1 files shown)


mdlpstsci (approved):

drizzlepac/processInput.py
```diff
@@ -793,6 +794,10 @@ def changeSuffixinASN(asnfile, suffix):
     # Assign newly created, reformatted array to extension
     newasn = np.array(newdata,dtype=new_dtype)
     fasn[1].data = newasn
+    # Explicitly set the values of the boolean column to original values since they get
+    # lost when writing out as a FITS_Rec object using the FITS interface.
+    for row, dcol in zip(fasn[1].data, newdata):
```
raswaters, line 799: Will this work for older astropy versions as well?

stsci-hack: This change does not rely on any specific feature of astropy, as it simply copies data using numpy functionality directly.  This was confirmed by interactively processing data from 2 ASNs (both which were able to reproduce the original problem) without any problems using both astropy 4.3.post1 and 5.0.2.

PR #1309, opened 2022-03-15
Ignore git information for installations in conda envs

stsci-hack:
These changes add logic to guard against trying to cross-match empty catalogs in svm_quality_analysis which would otherwise cause an Exception reporting an "Unrecognized format.  Exiting...".  

In addition, these changes look to see whether or not the package has been installed outside the source tree, like it would be when installed under a conda envs site-packages directory.  If no git information is present, it instead reports the commit id as 'master' as a default rather than raising an Exception.  

Finally, some changes were also included to clean up the git hash returned under Windows so that is is valid.

### DIFF
--- a/drizzlepac/haputils/comparison_utils.py
+++ b/drizzlepac/haputils/comparison_utils.py
@@ -352,6 +352,10 @@ def slFiles2dataTables(slNames):
             compData_in = Table.read(slNames[1], format='ascii.daophot')
         except Exception:
             compData_in = Table.read(slNames[1], format='ascii')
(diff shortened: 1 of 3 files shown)


stsci-hack:
The original problems were reproduced using data from visit 'ibvz01', and these changes allowed the data to finally process successfully.  The messages about git were also made less prominent so that they do not hold up the processing anymore.

mdlpstsci (dismissed):

mdlpstsci (approved):

PR #1311, opened 2022-03-15
Upgrading from MacOS 10.14 to 10.15 as 10.14 is no longer available

BGannon2:


stsci-hack (approved):
Simple enough update.

PR #1312, opened 2022-03-16
Synchronize output WCS specifications for SVM processing

stsci-hack:
The specification of the output WCS used in determining the backgrounds during catalog generation, and all throughout SVM processing, are now synchronized between what AstroDrizzle computes and what the SVM determines for the meta_wcs.  Hard-coded values and interpreted WCS objects were removed in favor of exact user and/or config-file specified output WCS values which eliminates a source of rounding error when trying to use the AstroDrizzle products with the SVM computations.  

These changes address the problems seen when processing some ACS/HRC datasets where an IndexError like:

```
 File "/ifs/pipeline/dms/rhel7/pkgs/miniconda3/envs/caldp_mvmbeta/lib/python3.8/site-packages/drizzlepac/haputils/catalog_utils.py", line 275, in compute_background
    num_of_zeros = np.count_nonzero(imgdata[self.footprint_mask] == 0)
IndexError: boolean index did not match indexed array along dimension 0; dimension is 1462 but corresponding boolean dimension is 1461
```

The problems were reproduced using data from visits 'j6d508' and 'j8ep07', and those same datasets (along with all HAP pytest tests) were used to verify that the problem was fixed with these changes.

### DIFF
--- a/drizzlepac/haputils/product.py
+++ b/drizzlepac/haputils/product.py
@@ -594,7 +594,6 @@ class FilterProduct(HAPProduct):
         # start by pulling out the specific parameters for the final refimage as
         # separate input parameters
         # rot and scale are specified in the config pars json files
-        # drizzle_pars["final_refimage"] = meta_wcs
         drizzle_pars["final_ra"] = meta_wcs.wcs.crval[0]
         drizzle_pars["final_dec"] = meta_wcs.wcs.crval[1]
         drizzle_pars["final_crpix1"] = meta_wcs.wcs.crpix[0]
--- a/drizzlepac/haputils/product.py
+++ b/drizzlepac/haputils/product.py
@@ -128,10 +128,14 @@ class HAPProduct:
         log.debug("The following images will be used: ")
         log.debug("{}\n".format(exposure_filenames))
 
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/product.py
```diff
@@ -587,7 +591,17 @@ def wcs_drizzle_product(self, meta_wcs):
         # Retrieve the configuration parameters for astrodrizzle
         drizzle_pars = self.configobj_pars.get_pars("astrodrizzle")
         # ...and set parameters which are computed on-the-fly
-        drizzle_pars["final_refimage"] = meta_wcs
+        # start by pulling out the specific parameters for the final refimage as
+        # separate input parameters
+        # rot and scale are specified in the config pars json files
+        # drizzle_pars["final_refimage"] = meta_wcs
```
mdlpstsci, line 597: Please remove commented out line 597.

mdlpstsci (changes requested):
Just a small change...

drizzlepac/haputils/product.py
stsci-hack: Fixed(?) in 508b432.

mdlpstsci (approved):

PR #1313, opened 2022-03-16
Refine HAP Pytests success criteria

stsci-hack:
This just cleans up a couple of HAP pytests; 

- fixed variable being referenced with the variable that actually exists in test_svm_ibqk07
- refined the success criteria for a test dataset with only 1 strong source to use atol instead of rtol, since 1 is 0.5 of 2 using rtol limits.


mdlpstsci (changes requested):
Although not part of this issue, test_svm_ibyt50.py has this same error so it may as well be fixed also (since it is so easy).

mdlpstsci (dismissed):
Yay -- more GAIA WCS...and improvements to test code!

stsci-hack:
The test `test_svm_ibyt50` also had its variables fixed in 5104ca0 as requested.

mdlpstsci (approved):

PR #1316, opened 2022-03-17
Ensure units of catalog variables comply with Astropy

mdlpstsci:
Modify the units of the catalog variables to put them into a format recognized by astropy.units.  The reason for this is when a table is read (Table.read), it is prohibitively slow when astropy cannot figure out the units.

We (Rick and Steve for MAST) are reading all the ecsv files and inserting them into database tables, and about 25% of the time was being spent reading these headers!  The problem is that astropy is trying to figure out what the unit is.

Tested with dataset ic5s09 (hst_13089_09_wfc3_uvis_total_ic5s09_point-cat.ecsv).  Verified all output catalogs, filters and total, contain the updated units.


stsci-hack (approved):
These changes make sense and seem straight-forward enough.

PR #1324, opened 2022-03-29
Add MEANWHT and MEDWHT keywords to drizzle products

mdlpstsci:
Add MEANWHT and MEDWHT keywords to drizzle products.  These keyword computations are currently working
and tested for SVM Total and Filter drizzle products where they are resident in the WHT extension.


stsci-hack (dismissed):
This looks good to me.

stsci-hack (approved):
New logic looks fine to me.

PR #1325, opened 2022-03-29
Correctly extract visit id for HAP diagnostics

stsci-hack:
These changes isolate the logic used to extract the visit ID for the quality assessment products.  This code gets used for both standard calibration products (with IPPPSSOOT filenames) as well as SVM products and needed to account for both types of filenames while still extracting the visit ID for parallel observations as well.


dulude (approved):
LGTM

PR #1326, opened 2022-03-30
Restrict version of Photutils to < 1.4.0.

mdlpstsci:
Updated setup.cfg.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ install_requires =
     pandas
     spherical_geometry>=1.2.22
     astroquery>=0.4
-    photutils>=1.0.0
+    photutils<1.4.0
     lxml
     PyPDF2
     scikit-image>=0.14.2


stsci-hack (approved):
Just what the doctor ordered...

mcara:
These kind of restrictions - setting upper limit - create compatibility issues that complicate installations and updates for end users later. Also, it would be nice to state in this PR the reason why such limitation is needed so that later anyone could understand why this was done.

mdlpstsci:
A restriction was introduced in setup.cfg forcing use of  Photoutils<1.4.0.  This version of Photoutils became public 5 days ago.  Since we are currently on a deadline to finalize a release of Drizzlepac for the HAP project, we did not want to introduce new changes to accommodate Photoutils 1.4.0 at this juncture.  We do have a ticket (JIRA HLA-728) to update the _detect_sources_ signature and will need additional tickets for thorough testing of Drizzlepac with Photoutils 1.4.0.

PR #1335, opened 2022-04-12
Update astropy min version to 5.0.4

WilliamJamieson:


stsci-hack (approved):
Fortunately, this update to v5.0.4 does not have any impact on the computations done in this package.  LGTM.

nden:
@WilliamJamieson  @stsci-hack Drizzlepac is one of the few packages which requires astropy at setup. That's why there's an astropy version in pyproject.toml which needs to be changed as well.

PR #1347, opened 2022-04-14
Update pyproject.toml astropy min version

WilliamJamieson:
As [noted](https://github.com/spacetelescope/drizzlepac/pull/1335#issuecomment-1099237782) by @nden in PR #1335, the astropy version in `pyproject.toml` needs to be updated. This PR does that too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = ["setuptools>=42.0",
             "setuptools_scm[toml]>=3.4",
             "wheel",
             "oldest-supported-numpy",
-            "astropy>=4.3",
+            "astropy>=5.0.4",
             "markupsafe<=2.0.1"]
 build-backend = "setuptools.build_meta"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = ["setuptools>=42.0",
             "setuptools_scm[toml]>=3.4",
             "wheel",
-            "oldest-supported-numpy",
+            "numpy>=1.18",
             "astropy>=5.0.4",
             "markupsafe<=2.0.1"]
 build-backend = "setuptools.build_meta"


nden:
pyproject.toml and setup.cfg define a version of numpy and these need to be in sync.

However, I am not sure numpy needs to be in the setup.cfg given that it's in pyproject.toml and is installed prior to installing drizzlepac. If it doesn't have to be included it may be better to remove it from setup.cfg so we don't have to update it in two places. Perhaps @pllim has more info on this?

WilliamJamieson:
@nden, numpy is not in the setup.cfg. The CI failures come from the fact that in the pyproject.toml, it declares `oldest-supported-numpy`, which in the python 3.8 configuration used works out to be 1.17.3. However, astropy 5.0.4 requires numpy 1.18+.

nden:
Oh, I see. Sounds like an astropy issue. Then we should set numpy >=1.18

pllim:
I don't grok why you need astropy at build time instead of install time. Is there some weird C-ext thingy you grab while building C extension?

Oldest support numpy is defined at https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg . I don't see how it conflicts with your package if you already have Python min version set to 3.8.

WilliamJamieson:
> Oldest support numpy is defined at https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg . I don't see how it conflicts with your package if you already have Python min version set to 3.8.

Oldest supported numpy was returning numpy 1.17.3 (I think its due to this https://github.com/scipy/oldest-supported-numpy/blob/ef8f28c646eefbc416408423eea35876dbadbcf5/setup.cfg#L53) during the python 3.8 CI runs. However, astropy 5.0.4 requires numpy 1.18+, so the package install was failing.

pllim:
But why do you need astropy at build time?

WilliamJamieson:
> But why do you need astropy at build time?

@nden, you referred to this in this comment: https://github.com/spacetelescope/drizzlepac/pull/1335#issuecomment-1099237782. Do you know why it is needed?

nden:
The wcslib header files are needed to build the extension.

stsci-hack (commented):
Given the dependencies for astropy, this seems reasonable to me.

PR #1349, opened 2022-04-18
Clarify comment portion of FITS card for MEANWHT and MEDWHT.

mdlpstsci:
Small change to the comment portion of the FITS cards for MEANWHT and MEDWHT to clarify the meaning of the keyword.


stsci-hack (approved):
Fine with me.

PR #1351, opened 2022-04-20
Flag and ignore bad images based on detecting linear features

stsci-hack:
This adds a check to the analyze_wrapper function in analyze.py to ignore images it detects as being affected by guiding problems.  The algorithm used depends on detecting lines in the FLT/FLC image's science array using the Hough transform as implemented in scikit-image.  If the image is 'dominated' by linear features that do not align with columns (90 degrees exactly) or diffraction spikes (+/- 45 degrees exactly), then it is deemed as 'bad' and removed from the processing list.  

This code was tested against:
    - iacs01skq:  good WFC3/IR image
    - j96a02doq: good ACS/HRC single-source coronographic image
    - jc1b01ihq : ACS/WFC Grism image
    - jc1b01ieq: good ACS/WFC F475W image (direct image for Grism exposure)
    - ie9m30wlq : WFC3/UVIS image affected by guiding problems
    - ie9m29wdq : good WFC3/UVIS image of rich stellar field


drizzlepac/haputils/analyze.py
```diff
@@ -471,3 +483,163 @@ def generate_msg(filename, msg, key, value):
...
+    data = hdu[("SCI", 1)].data.copy()
+    scale_data = hdu[("SCI",1)].header["bunit"].endswith('/S')
+    data = np.nan_to_num(data, nan=0.0)  # just to be careful
+    if scale_data:
+        # Photutils works best in units of DN
+        scale_hdr = hdu[0].header if 'exptime' in hdu[0].header else hdu[1].header
+        scale_val = scale_hdr['exptime']
+        data *= scale_val
+    bkg_stats = sigma_clipped_stats(data, maxiters=2)
+    bkg_limit = bkg_stats[1] + bkg_stats[2]
```
mdlpstsci, line 526: Please add comments for explanation. It looks like the bkg_limit is the combination of the median and stddev.

drizzlepac/haputils/analyze.py
```diff
@@ -471,3 +483,163 @@ def generate_msg(filename, msg, key, value):
...
+    data -= bkg_limit
+    imgarr = np.clip(data, 0, data.max())
+
+    # Determine rough number of probable sources
+    # Trying to ignore small sources (<= 4x4 pixels in size, or npixels < 17)
+    #   which are either noise peaks or head-on CRs.
+    segm = detect_sources(imgarr, 0, npixels=17)
+    log.debug(f'Detected {segm.nlabels} raw sources in {filename}')
+    if segm.nlabels < 2:
+        return False, None
```
mdlpstsci, line 538: Why are there two values being returned from this function at this point?  It looks like there should only be one value returned.

stsci-hack:
In addition to the problems raised in the review, this code mistakenly flagged good images as 'bad' for visit 'ib7788' (a dense star field).   This code can not be merged until the logic has been amended to recognize such fields as good.

drizzlepac/haputils/analyze.py
stsci-hack: Fixed in 95a5e59.

drizzlepac/haputils/analyze.py
stsci-hack: Fixed in 9acdf37.

drizzlepac/haputils/analyze.py
```diff
@@ -529,28 +534,51 @@ def verify_guiding(filename, min_length=33):
     data -= bkg_limit
     imgarr = np.clip(data, 0, data.max())
 
+    # Build up a mask of all bad pixels and ignore them when looking for
+    # sources and linear features
+    dqarr = None
+    for extn in hdu:
+        if 'extname' in extn.header and extn.header['extname'] == 'DQ':
+            dqarr = hdu[("DQ", 1)].data.copy()
+            break
```
mdlpstsci, line 543: Maybe it does not matter in this specific case, but could the hdu[("DQ",1)].data.copy() just be extn.data.copy() to be more generic?  There may only be "1" DQ in the file, but the source code does not necessarily know this.  Just a question really.

drizzlepac/haputils/analyze.py
```diff
@@ -529,28 +534,51 @@ def verify_guiding(filename, min_length=33):
...
+    if dqarr is not None:
+        dqmask = bitfield_to_boolean_mask(dqarr, ignore_flags=[256, 512, 1024, 2048, 4096])
+    else:
```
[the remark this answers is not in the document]
stsci-hack: I didn't use enumeration, but the values are now commented at the top of the module in be338c9.

drizzlepac/haputils/analyze.py
stsci-hack: Good point.  That change was made in 476518e.

mdlpstsci (changes requested):
Just a question and request to add a comment.

mdlpstsci (approved):

PR #1357, opened 2022-04-28
Improve algorithm for identifying and filtering large segments

mdlpstsci:
Improved the algorithm used to filter the array of big segments which
will be deblended.  Added safety catches, more comments, and clean-up to
the logic.  The clean-up clarifies the distinction between the code
used to identify "big islands" or a large "source fraction" which are
only used to set a flag to indicate another kernel or background
computation could yield improved results.  This is distinct from
an array of identified large segments which are filtered only to
contain segments of reasonable size to be deblended.  Deblending
enormous segments can be prohibitive in time.

This PR address the issue reported in JIRA HLA-757

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1761,13 +1761,14 @@ class HAPSegmentCatalog(HAPCatalogBase):
                 self._define_empty_table(g_segm_img)
                 return
 
-            # Deblend the segmentation image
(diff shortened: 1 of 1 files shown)


stsci-hack (approved):
I think I understand what you are doing here, and it looks fine to me.

mdlpstsci:
This ticket actually addresss JIRA HLA-753 and HLA-757.  All three datasets were used for testing, as well as additional datasets.

stsci-hack (approved):
LGTM

Issue #1359, opened 2022-05-03
MVM: SkyCell identify fails for data from visit iepo3k


stscijgbot-hstdp:
_Issue [HLA-765](https://jira.stsci.edu/browse/HLA-765) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

No SkyCell ID was determined for data from visit 'iepo3k' causing problems during pipeline MVM processing as noted in HSTSDP-1780.  

Interactively trying to identify what SkyCell this data overlaps resulted in the following Exception being generated:

> Traceback (most recent call last):
>  File "C:\Users\Warren\miniconda3\envs\rthst38\lib\site-packages\IPython\core\interactiveshell.py", line 3441, in run_code
>  exec(code_obj, self.user_global_ns, self.user_ns)
>  File "<ipython-input-114-b4b466ec7059>", line 1, in <module>
>  scids = cutils.get_sky_cells(flcs)
>  File "D:\Drive_D\cygwin\home\Warren Hack\github\drizzlepac\drizzlepac\haputils\cell_utils.py", line 146, in get_sky_cells
>  File "D:\Drive_D\cygwin\home\Warren Hack\github\drizzlepac\drizzlepac\haputils\cell_utils.py", line 761, in get_sky_cells
>  File "D:\Drive_D\cygwin\home\Warren Hack\github\drizzlepac\drizzlepac\haputils\cell_utils.py", line 741, in get_projection_cells
>  if ra is None:
>  File "D:\Drive_D\cygwin\home\Warren Hack\github\drizzlepac\drizzlepac\haputils\cell_utils.py", line 617, in get_edges_sky
>  if len(self.corners) == 0:
>  File "D:\Drive_D\cygwin\home\Warren Hack\github\drizzlepac\drizzlepac\haputils\cell_utils.py", line 598, in find_corners
> ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (2,) + inhomogeneous part.

If this Exception was thrown when generating the database entry for exposures from this visit, then no SkyCell ID would have been found and the database would have an entry of 'None'.  At any rate, this problem needs to be resolved so that the data can be placed in the correct SkyCell.

stscijgbot-hstdp:
[...]

Issue #1358, opened 2022-04-29
MVM failure: IndexError: index 0 is out of bounds for axis 0 with size 0


stscijgbot-hstdp:
_Issue [HLA-764](https://jira.stsci.edu/browse/HLA-764) was created on JIRA by [Matthew Burger](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mburger):_

A few skycells have failed in MVM_Repro: 

skycell-p2489x15y02

skycell-p0544x18y15

skycell-p0081x16y13

Error message:

 
```java
Traceback (most recent call last):
  File "/dms/local/hst/pipeline/pkgs/miniconda3/envs/caldp_20220406/lib/python3.8/site-packages/drizzlepac/hapmultisequencer.py", line 375, in run_mvm_processing
    filter_item.generate_footprint_mask(save_mask=False)
  File "/dms/local/hst/pipeline/pkgs/miniconda3/envs/caldp_20220406/lib/python3.8/site-packages/drizzlepac/haputils/product.py", line 105, in generate_footprint_mask
    footprint.build(exposure_names, scale=True)
  File "/dms/local/hst/pipeline/pkgs/miniconda3/envs/caldp_20220406/lib/python3.8/site-packages/drizzlepac/haputils/cell_utils.py", line 381, in build
    self.find_bounded_wcs()
  File "/dms/local/hst/pipeline/pkgs/miniconda3/envs/caldp_20220406/lib/python3.8/site-packages/drizzlepac/haputils/cell_utils.py", line 430, in find_bounded_wcs
    ymin, ymax, xmin, xmax = calc_bounding_box(self.total_mask)
  File "/dms/local/hst/pipeline/pkgs/miniconda3/envs/caldp_20220406/lib/python3.8/site-packages/drizzlepac/haputils/cell_utils.py", line 1135, in calc_bounding_box
    rmin, rmax = np.where(rows)[0][[0, -1]]
IndexError: index 0 is out of bounds for axis 0 with size 0
2022119135535 INFO src=drizzlepac.hapmultisequencer- Processing completed at 2022-04-29 13:55:35.523896
2022119135535 INFO src=drizzlepac.hapmultisequencer- Total processing time: 17.952799 sec
2022119135535 INFO src=drizzlepac.hapmultisequencer- Return code for use by calling Condor/OWL workflow code: 0 (zero) for success, non-zero for error or exit. 
[...]

PR #1360, opened 2022-05-03
Ignore layers with no overlapping exposures

stsci-hack:
This implements several changes (fixes?) to how exposures are associated with SkyCells and how the processing will respond when an unexpectedly non-overlapping exposure is specified for a given SkyCell.  The specific changes include:

- adding an attribute to the SkyCellProduct class to indicate whether or not there are overlapping exposures for that product
- adding logic to check whether or not any exposure overlaps a SkyCellProduct footprint
- checking whether or not this attribute is True when running `hapmultisequencer.run_mvm_processing`
- treating all edge pixels and corner pixels as lists of numpy arrays (addresses #1358 and #1359)
- checking each exposure separately to see what SkyCell(s) it overlaps when running `cell_utils.get_sky_cells`

These changes were all tested against exposures from visit `icvz15` and all other exposures identified for `skycell-p2489x15y02`.  It turned out that the IR exposures from visit `icvz15` do not actually overlap this SkyCell, and these changes no longer identifies them as members of that SkyCell.

### DIFF
--- a/drizzlepac/hapmultisequencer.py
+++ b/drizzlepac/hapmultisequencer.py
@@ -207,6 +207,10 @@ def create_drizzle_products(total_obj_list, custom_limits=None):
     # create the drizzle-combined filtered image, the drizzled exposure (aka single) images,
     # and finally the drizzle-combined total detection image.
     for filt_obj in total_obj_list:
+        if not filt_obj.valid_product:
+            # If this object has 'valid_product' set to False,
+            # skip processing this object
+            continue
         filt_obj.rules_file = proc_utils.get_rules_file(filt_obj.edp_list[0].full_filename,
                                                         rules_type='MVM',
                                                         rules_root=filt_obj.drizzle_filename)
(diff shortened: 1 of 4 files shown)


drizzlepac/hapmultisequencer.py
```diff
@@ -373,7 +378,10 @@ def run_mvm_processing(input_filename, skip_gaia_alignment=True, diagnostic_mode
             # Compute mask keywords early in processing for use in determining what
             # parameters need to be used for processing.
             filter_item.generate_footprint_mask(save_mask=False)
-
+            if not filter_item.mask_computed:
+                log.warning(f"Ignoring {filter_item.info} as no input exposures overlap that layer.")
+                filter_item.valid_product = False
```
mdlpstsci, line 383: Maybe I have my objects confused, but it looks like the valid_product attribute is already set in the generate_footprint_mask() method?

mdlpstsci (changes requested):
As far as I can tell, this PR looks fine.  I just have a question with respect to the setting of a variable as I want to make sure this is not done in multiple locations.  If this is actually fine, and it is my confusion, then I will approve.

drizzlepac/hapmultisequencer.py
stsci-hack: Good catch!  That is a leftover from all the iterations during development on where this flag should be set.  In fact, the if statement should have been updated to use the flag set by the generate_footprint_mask() method instead of setting that flag here.  **This has been fixed in dbd64d1.**

mdlpstsci (approved):

PR #1362, opened 2022-05-05
Update crder units

stsci-hack:
Simple changes to correctly describe the CRDER and RMS_RA keywords in headers aligned to GAIA.   This also scales the CRDER keyword values to be in units of deg (as expected by FITS and described in the comment)  instead of mas for the values from the WCS fit.


mdlpstsci (approved):
Straightforward changes.

PR #1365, opened 2022-05-06
This change addresses bugs associated with the big_segments attribute of the segmentation image

mdlpstsci:
(1) During construction of the attribute, it is necessary to check the array is not empty.  If it is, set the attribute
to None.  (2) Only perform the deblending step if the attribute in not None.


stsci-hack (approved):
Changes make sense to me.

PR #1366, opened 2022-05-06
Update the WFC3 rules files -- MERGE after PR#1365

mdlpstsci:
Update the WFC3 rules files to remove WCS-associated keywords (CRPIX, CRVAL, CTYPE, and CD) from the three column entries which govern how the output header values are computed.  Some of these entries were set to "multi" which caused the output FITS header to contain CTYPE1 = CTYPE2 = 'MULTIPLE' when these keywords should have been set to 'RA---TAN' and 'DEC--TAN', respectively.


stsci-hack:
Closes #1367.

PR #1368, opened 2022-05-10
Only allow "verify_guiding" check for MVM processing

mdlpstsci:
Added protection around the function call, verify_guiding, from the analyze_wrapper routine.  The function will only be called if the processing type is "MVM".  Also added a new parameter, do_verify_guiding, to the process function in runastrodriz.py.  This new parameter will allow a user to turn on verify_guiding for pipeline processing.  Organized the order of the parameter switches for the process function.  Added explanatory comments in the code for these changes.


stsci-hack (dismissed):
This maintains backwards compatibility with pipeline, while keeping the changes as simple as possible.  LGTM.

stsci-hack (approved):
Makes sense to me.

PR #1370, opened 2022-05-11
Pass along default WCSNAME

stsci-hack:
This simple change makes sure that the WCSNAME and WCSTYPE keywords do not get populated by blank strings if the output WCS used for drizzling did not have a `wcs.name` attribute to use in creating the WCSNAME keyword value.  This situation can happen if an output WCS gets defined and passed in to AstroDrizzle for the final drizzle step, leading to the blank string overwriting the default value  from the input images during MVM processing.

### DIFF
--- a/drizzlepac/outputimage.py
+++ b/drizzlepac/outputimage.py
@@ -721,6 +721,7 @@ def getTemplates(fnames, blend=True, rules_file=None):
 
     return newhdrs, newtab
 
+
 def addWCSKeywords(wcs, hdr, blot=False, single=False, after=None):
     """ Update input header 'hdr' with WCS keywords.
     """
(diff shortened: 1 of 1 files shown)


drizzlepac/outputimage.py
```diff
@@ -729,8 +730,9 @@ def addWCSKeywords(wcs, hdr, blot=False, single=False, after=None):
 
     # Update WCS Keywords based on PyDrizzle product's value
     # since 'drizzle' itself doesn't update that keyword.
-    hdr['WCSNAME'] = wname
-    hdr['WCSTYPE'] = wtype
+    if wname != '':
```
mdlpstsci, line 733: Please add a comment here indicating this is a temporary fix until the WCSNAME issue is resolved for MVM.

mdlpstsci (changes requested):
Just a request for an internal comment in the code.

drizzlepac/outputimage.py
stsci-hack: Additional comments were added in 3b36622.  It is not clear that this specific change is temporary, so the comments simply clarified the logic that was implemented.  The definition of the MVM WCSNAME value will not require this logic to be removed, since this logic causes the WCSNAME to revert to what was used to align the input exposures as opposed to the MVM WCS definition which is not associated with any catalog or alignment fit.

mdlpstsci (approved):
After our discussion, the current solution is as viable as any solution until the team can come up with a better idea.

PR #1371, opened 2022-05-11
Fix the size of the HAPEXPNAME column in the HDRTAB of the MVM output DRZ/DRZ file

mdlpstsci:
Ensure the HAPEXPNAME column is the size of the largest input name currently being processed OR at least 51 characters.


stsci-hack (approved):
Simple enough change to correct a silly oversight in defining max_len.

PR #1378, opened 2022-05-20
Updates to create the MVM quality analysis infrastructure

mdlpstsci:
Updates to create the MVM quality analysis infrastructure which only required a few small changes to hapmultisequencer.py as it is based upon the SVM code.  Also, created the mvm_quality_analysis.py module which contains a single function at this time, report_wcsnames.  This latter file is where additional Q&A functions will cull/create data for associated JSON files.


stsci-hack (changes requested):
Where is the new module 'mvm_quality_analysis'?

drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -0,0 +1,270 @@
...
+
+            # Obtain the value of S_REGION from all SCI extensions in the file
+            sregion = pu.interpret_sregion(edp_object.full_filename, extname='SCI')
+
+            # Convert the S_REGION into a single list containing lists of RA/Dec coordinates
+            sregion_list = []
+            for item in range(len(sregion)):
+                sregion_list.extend(sregion[item].tolist())
+
+            # Capture the RA and Dec coordinates into their own lists
```
stsci-hack, line 118: It may be more efficient and use less code to convert the "sregion_list" to a numpy array, then extract the "ra" and "dec" arrays using indexing syntax such as "ra = sregion_arr[:,0]".  Appending to a list can actually be inefficient.

stsci-hack (dismissed):
I don't see any real problems, but did have a couple of comments which could be (but don't have to be) addressed before merging.

drizzlepac/haputils/mvm_quality_analysis.py
mdlpstsci: Made the suggested changes here.

stsci-hack (approved):
LGTM!

PR #1387, opened 2022-06-01
Fixed skycell size in pixels as quoted in the documentation.

mdlpstsci:
A simple fix to documentation was done.  SkyCells are 0.2 x 0.2 degrees on a side.  For 0.04" pixels, the SkyCells are 18000 x 18000 pixels and **not** 21500 x 21500 pixels.


dulude (approved):
LGTM

PR #1395, opened 2022-06-10
Modified the docstring which defines the HAPLEVEL and its associated meaning

mdlpstsci:
The changes are that Level 3 is unused and Level 4 indicates the multi-visit mosaics.  The full documentation is now:

      1 : calibrated (FLT/FLC) input images and exposure level drizzle products with improved astrometry
      2 : filter and total products combined using the improved astrometry, consistent pixel scale, and oriented to North
      3 : unused
      4 : multi-visit mosaics aligned to a common tangent plane


stsci-hack (approved):
As simple as it gets, and looks right to me.

PR #1396, opened 2022-06-10
Created the mvm_quality_analysis.py module 

mdlpstsci:
The mvm_quality_analysis.py module has been created which generates JSON information regarding the MVM processing.  This module is a streamlined version of the svm_quality_analysis.py module in that it has been cleaned up to contain only what is necessary to run the initial data collection routine.  It is anticipated the module will be "filled out" as new data collection functions are added.


drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -103,49 +116,94 @@ def report_wcsname(total_product_list, json_timestamp=None, json_time_since_epoc
...
+                wcsname = exp[sci].header['WCSNAME'].upper()
+
+                # Each chip is assigned the 2**n value associated with its WCSNAME
+                suffix = wcsname.split('-')[1]
+                expo_wcs_value = wcs_pref_list[suffix]
+
+                # If this is a multi-chip exposure, get the CHIP number
+                chip_number = 0
+                try:
+                    chip_number = exp[sci].header['CCDCHIP']
```
stsci-hack, line 151: Just as a note regarding the impact of this on potential future development:  WFPC2 images use the 'DETECTOR' keyword in the SCI headers to record the 'chip_number'.

drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -103,49 +116,94 @@ def report_wcsname(total_product_list, json_timestamp=None, json_time_since_epoc
...
+                                   'X': xc,
+                                   'Y': yc}
+
+                # Make the PrimaryWCS unique for every exposure/chip to ensure no entry is overwritten.
+                diagnostic_obj.add_data_item(active_wcs_dict, 'PrimaryWCS' + str(counter) + "_" + str(chip_number), 
+                                             descriptions={'filename': 'Exposure filename',
+                                                           'instrument': 'Instrument',
+                                                           'detector': 'Detector',
+                                                           'primary_wcsname': 'Active WCS',
+                                                           'wcs_value': 'Value (2^n) as numeric ID of WCSNAME',
```
stsci-hack, line 185: The 'wcs_value' should be used as the 'fill' value for the footprint, not just a numeric ID and this comment should reflect that use of this value.

drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -103,49 +116,94 @@ def report_wcsname(total_product_list, json_timestamp=None, json_time_since_epoc
...
+                                                           'chip_number': 'CCD Chip Number',
+                                                           'img_value': 'Value (2^n) for fill of the footprint',
```
stsci-hack, line 187: This value could also be used as a 'fill' value for the footprint, but only in a secondary capacity.  Perhaps this comment should be updated to reflect it more as a 'numeric ID' instead?

drizzlepac/haputils/mvm_quality_analysis.py
stsci-hack: Good comment!

stsci-hack (approved):
New commetns look good!

PR #1404, opened 2022-06-21
Reorganized how information is stored in the JSON files

mdlpstsci:
WCS information has been reorganized to be better grouped according to its skycell name and, as a result, it also produces a more compact JSON file.


dulude (approved):
LGTM

Issue #1339, opened 2022-04-13
MVM: cross-match SVM catalog info in regions of MVM overlap


stscijgbot-hstdp:
_Issue [HLA-741](https://jira.stsci.edu/browse/HLA-741) was created on JIRA by [Warren Hack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=hack):_

Source catalogs generated as part of the SVM processing should be used, if available, to perform cross-matching analysis for regions of overlap between exposures in MVM layers where exposures from different visits overlap.  

The logic for transforming the source catalog positions into MVM positions and then identifying what sources (catalog rows) fall in the region of overlap will need to be developed for this analysis to be performed.  

Implementation of ways to visualize these results will be part of this ticket since it requires knowledge of how the results are generated and stored, and serves as the final product for the Q&A processing.

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-741?focusedCommentId=586986#comment-586986):_

Need overlapping observations from differnet visits to use as test particle

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-741?focusedCommentId=586987#comment-586987):_

NGC 1275 ACS/WFC F814W observations 15235_01 and 15235_02 [MAST Link](https://mast.stsci.edu/portal/Mashup/Clients/Mast/Portal.html?searchQuery=%7B%22service%22%3A%22CAOMDB%22%2C%22inputText%22%3A%22WESTERLUND1%22%2C%22paramsService%22%3A%22Mast.Caom.Cone%22%2C%22title%22%3A%22MAST%3A%20WESTERLUND1%22%2C%22columns%22%3A%22*%22%2C%22caomVersion%22%3Anull%7D)

stscijgbot-hstdp:
_Comment by [Michael Dulude](https://jira.stsci.edu/secure/ViewProfile.jspa?name=dulude) on [JIRA](https://jira.stsci.edu/browse/HLA-741?focusedCommentId=586992#comment-586992):_

Westerlund 1 WFC3/IR F160W observations 
[...]

PR #1405, opened 2022-06-21
MVM: Add quality test that cross-matches SVM catalog info in regions of MVM overlap and statistically quantifies results

dulude:
**Relevant Tickets and Issues**
- JIRA Ticket [HLA-741](https://jira.stsci.edu/browse/HLA-741)
- Github issue #1339 

**Summary of Changes**
This PR adds a new test to mvm_quality_analysis.py that locates regions were observations from different proposal/visits overlap and performs a crossmatch of the relevant SVM sourcelists to identify co-located sources. Statistical characterizations of the cross-matched sources are computed, and the results are reported and written  to a .json file.

Changes to code are as follows:
- mvm_quality_analysis.py: Added "overlap_crossmatch_analysis" test. This includes the following changes:
  - addition of 7 new subroutines (overlap_crossmatch_analysis, crossmatch_analysis, crossmatch_sources, determine_if_overlaps_exist, locate_overlap_regions, locate_svm_products, and reject_sources_not_in_overlap_region)
  - modifications to run_quality_analysis() to accommodate my new test
  - modifications to the command-line interface to accommodate my new test
- modifications to interdetector_sourcelist_crossmatch.py so that portions of the code could be utilized by my new test


drizzlepac/devutils/comparison_tools/interdetector_sourcelist_crossmatch.py
```diff
@@ -150,11 +149,11 @@ def run(sl_names, img_names, diagnostic_mode=False, log_level=logutil.logging.IN
     for diff_ra, title, units in zip(diff_list, title_list, units_list):
         log.info("Comparison - reference {} statistics ({})".format(title, units))
 
-        compute_stats(diff_ra, title)
+        foo = compute_stats(diff_ra, title)
```
stsci-hack, line 152: The function 'compute_stats' can accept the logging level as a parameter.  Is the logging level set in this calling routine?  If so, that logging level should be used to control the logging for this function as well by passing it in.

mdlpstsci, line 152: Since the variable `foo` does not seem to be used anywhere else in this routine, please replace with the Pythonic `_`.

drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -247,6 +266,740 @@ def run_quality_analysis(total_obj_list,
...
+
+    log.info("Reference catalog: {}, {} sources".format(os.path.basename(overlap_info["svm_sourcelist_{}".format(str(ref_index))]), len(svm_sourcelist_list[ref_index])))
+    log.info("Comparison catalog: {}, {} sources".format(os.path.basename(overlap_info["svm_sourcelist_{}".format(str(comp_index))]), len(svm_sourcelist_list[comp_index])))
+
+    # prepare xyxymatch inputs
+    ref_xy = np.stack((svm_sourcelist_list[ref_index]['X-Skycell'], svm_sourcelist_list[ref_index]['Y-Skycell']), axis=1)
+    comp_xy = np.stack((svm_sourcelist_list[comp_index]['X-Skycell'], svm_sourcelist_list[comp_index]['Y-Skycell']), axis=1)
+    
+    # execute crossmatch
+    matches = xyxymatch(comp_xy, ref_xy, tolerance=5.0, separation=1.0)
```
stsci-hack, line 758: You may want to wrap this call in a try...except block since this C-code can fail spectacularly under the right conditions.  Also, if the offsets are more than 5 pixels apart, then it will fail to find a cross-match.  Increasing the tolerance to a larger value to something more like 20pixels does increase the risk of cross-match confusion in crowded fields, but the RMS for the non-GAIA catalogs can actually approach a full-arcsecond compared to GAIA.  Thus, this low tolerance value would actually be blind to such large offsets which are what we are hoping to flag with this analysis.  A try...except block would at least allow this code to continue if it does fail in extremely crowded fields.

drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -247,6 +266,740 @@ def run_quality_analysis(total_obj_list,
...
+        matched_lines_comp.append(item[2])
+        matched_lines_ref.append(item[5])
+    cross_match_details = [len(ref_xy), len(comp_xy), len(matches)]
+    return ref_index, comp_index, matched_lines_ref, matched_lines_comp, cross_match_details
+
+# ------------------------------------------------------------------------------------------------------------
+
+
+def determine_if_overlaps_exist(total_obj_list, log_level=logutil.logging.NOTSET):
+    """determines if there are any regions where observations overlap.
```
stsci-hack, line 779: Very nice function that could really be useful in multiple places.

stsci-hack (changes requested):
There are a couple of comments which need to be addressed before this can be approved. The most important is the one related to the use of 'xyxymatch'.

drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -247,6 +266,740 @@ def run_quality_analysis(total_obj_list,
...
+                                         units=overlap_region_data_table_units,
+                                         title="{} crossmatched reference X, Y, RA, Dec, and crossmatched comparison - reference difference values".format(section_title_base_string))
+
+        # Write out the file
+        diagnostic_obj.write_json_file(json_filename, clobber=True)
+        log.info("Wrote crossmatch results file {}".format(json_filename))
+
+        # Clean up
+        del diagnostic_obj
+
```
mdlpstsci, line 588: I greatly appreciate that the logical steps for the main algorithm are numbered/organized for this procedural routine.  I also see that you have broken out major steps into their own functions.  Nevertheless, I find routines which are hundreds of lines long to be hard to follow, and I wonder if the code would benefit from a more modular approach.  Due to the constraints on your time, I am NOT suggesting you rework this code at this time.

drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -247,6 +266,740 @@ def run_quality_analysis(total_obj_list,
...
+    overlap_dict = locate_svm_products(overlap_dict, sourcelist_type, log_level=log_level)
+    sl_xy_column_name_dict = {"point": ["X-Center", "Y-Center"],
+                              "segment": ["X-Centroid", "Y-Centroid"]}
+    num_overlaps = len(overlap_dict.keys())
+    stats_dict = {}
+    total_obj_mapping_dict = collections.OrderedDict()
+    initial_cat_sizes = {}
+    for overlap_num, bit_value in zip(range(1, num_overlaps+1), sorted(overlap_dict)):
+        # report basic information about each overlap
+        num_overlaps = len(overlap_dict.keys())
```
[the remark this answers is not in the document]
dulude: Good catch! Thanks, Warren!
Issue resolved by commit [7e24797](https://github.com/spacetelescope/drizzlepac/pull/1405/commits/7e2479768d5b384dc9035548a076e4a0bc27d10d).

drizzlepac/haputils/mvm_quality_analysis.py
dulude: thanks!

mdlpstsci (approved):
Looks good to go to me.

stsci-hack (approved):
LGTM

PR #1408, opened 2022-06-28
Initial version of mvm_quality_graphics.py module  

mdlpstsci:
Infrastructure setup and initial implementation of vertical bar chart depicting WCSNAMEs per layer of a skycell. This is a Work In Progress.  The "wcs" routines need proper mapping of colors to WCSNAMEs to be hard set so no matter when plots are generated, they can be inter-compared.


dulude (approved):
LGTM

PR #1409, opened 2022-06-30
Delete all reference catalogs during SVM processing

stsci-hack:
Alignment of SVM products often requires multiple attempts based on multiple astrometric catalogs.  These changes keep track of what reference catalogs get written out for use in alignment, then removes all of them if not in diagnostic_mode during SVM processing.  

This prevents these ref_cat files from remaining on disk during pipeline processing and resolves [HLA-775](https://jira.stsci.edu/browse/HLA-775).

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -782,12 +782,12 @@ def run_align_to_gaia(tot_obj, log_level=logutil.logging.INFO, diagnostic_mode=F
 
     # Remove all reference catalogs written out to disk, if not in diagnostic mode
     if not diagnostic_mode:
-        try:
             log.info(f"Removing reference catalogs:\n {gaia_obj.refcat_filenames}")
             for fname in gaia_obj.refcat_filenames:
-                os.remove(fname)
-        except (OSError, TypeError):
-            pass
+                try:
+                    os.remove(fname)
+                except (OSError, TypeError):
+                    pass
 
     return headerlet_filenames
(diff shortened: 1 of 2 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -773,14 +773,22 @@ def run_align_to_gaia(tot_obj, log_level=logutil.logging.INFO, diagnostic_mode=F
...
     else:
         # Get names of all headerlet files written out to file
         headerlet_filenames = [f for f in align_table.filtered_table['headerletFile'] if f != "None"]
 
+    # Remove all reference catalogs written out to disk, if not in diagnostic mode
+    if not diagnostic_mode:
+        try:
+            log.info(f"Removing reference catalogs:\n {gaia_obj.refcat_filenames}")
+            for fname in gaia_obj.refcat_filenames:
+                os.remove(fname)
```
mdlpstsci, line 788: I think you want to put the "try" after the "for" statement.  In this way if there is some exception (e.g., fname does not exist), the loop will still continue and remove the remaining files in the list, if any.

mdlpstsci (changes requested):
Double-check my logic where I have requested a change.

drizzlepac/hapsequencer.py
stsci-hack: Good catch!  Fixed in 046e57f.

mdlpstsci (approved):
Update looks fine. The changes are "good to go".

PR #1411, opened 2022-07-06
Revert sky matching upon any error

stsci-hack:
Super simple update to the sky matching code in AstroDrizzle to always revert to a less complex sky matching algorithm if the full matching algorithm fails.


mdlpstsci (approved):
Straightforward change.

PR #1412, opened 2022-07-07
Update runastrodriz to work with WFPC2 singletons

stsci-hack:
These changes add all the support necessary to successfully process WFPC2 images (at least full readout images) with 'runastrodriz', including alignment to GAIA.  This includes adding default parameter files under 'hap_pars/svm_parameters' for WFPC2 support based on WFC3/IR values.  Changes to 'haputils/analyze.py' were also implemented to account for the difference in keywords used to determine whether to process a WFPC2 image or not.

A new function was also added to 'drizzlepac/wfpc2Data.py' that converts the multiple GEIS-based WFPC2 MEF files into a single MEF with the same set of extensions as ACS or WFC3 MEF files.  Specifically, the 'c0m.fits' files provide the SCI extensions, while the 'c1m.fits' files serve as the DQ extensions and the ERR extension starts as a Poissonian estimate of the errors based on the SCI extension.  This creates an 'flt.fits' file consistent with ACS or WFC3 MEF files that then gets used for all subsequent processing.  
 
These changes were used to successfully create flt.fits files for all the exposures in the visit 'u5b806' where all the images were aligned to GAIAeDR3 with an rms of about 20mas using ~385 sources per exposure. 

In addition, new rules files and parameter files need to be added to the package to support WFPC2 processing.


drizzlepac/haputils/align_utils.py
```diff
@@ -697,7 +703,7 @@ def find_alignment_sources(self, output=True, dqname='DQ', crclean=False,
 
             dqmask = self.build_dqmask(chip=chip)
             sciarr = self.imghdu[("SCI", chip)].data.copy()
-            detector = self.imghdu[0].header['detector']
+            # detector = self.imghdu[0].header['detector']
```
mdlpstsci, line 706: Should this line just be deleted?

stsci-hack: Removed in e9ca846.

mdlpstsci (approved):
Code updates done thus far look fine to me.  The configuration files are copies of WFC3/IR, so they are a good starting point.

stsci-hack:
This set of changes now supports full pipeline alignment to GAIA through `runastrodriz` as well as full SVM processing through `hapsequencer`.  This required the addition of rules files for WFPC2 products, parameter files for WFPC2 SVM (and MVM) processing based on WFC3/IR config files, and updates to SVM code as well.  These changes were tested on the 25 WFPC2 exposures of 47Tuc from visit **u5b806**.

drizzlepac/haputils/align_utils.py
```diff
@@ -170,7 +170,13 @@ def __init__(self, input_list, clobber=False, dqname='DQ', process_type='',
         try:
             for img in self.process_list:
                 log.info("Adding {} to HAPImage list".format(img))
-                detector = fits.getval(img, 'detector')
+                hdr0 = fits.getheader(img)
+                instrume = hdr0.get('instrume')
+                if instrume.lower() == 'wfpc2' and 'detector' not in hdr0:
```
mdlpstsci, line 175: Is it safe to search for 'detector' (lower-case) here in hdr0?

mdlpstsci (changes requested):
Just a question or two to resolve?

mdlpstsci (dismissed):

mdlpstsci (approved):
Oops.   I should have caught this issue!

PR #1413, opened 2022-07-13
MVM: Update cross-match routine to read pipeline poller file

dulude:
### Relevant JIRA Ticket
- Ticket [HLA-802](https://jira.stsci.edu/browse/HLA-802)

### Summary
- As originally written, mvm_quality_analysis.py would only work with input images that were named with the "pipeline" filename format (ipppssoot_fl[tc].fits), and would crash if presented with input images that were named with the "hap" filename format (hst_<propid>_<visit>_<instr>_<det>_<filter>_<ipppssoo>_fl[tc].fits). This PR fixes the hap filename format incompatibility so that mvm_quality_analysis.py can process input _fl[ct].fits images whose filenames use either format.
- removed unused overlap_dict element 'svm_img_'

### DIFF
--- a/drizzlepac/haputils/mvm_quality_analysis.py
+++ b/drizzlepac/haputils/mvm_quality_analysis.py
@@ -390,11 +390,6 @@ def overlap_crossmatch_analysis(total_obj_list, sourcelist_type="point", good_fl
                 log.warning("Unable to locate one or more SVM sourcelist(s) required for crossmatch!")
                 error_flag = True
                 break
-            if overlap_dict[bit_value]["svm_img_{}".format(set_num)] is None:
-                log.warning(
-                    "Unable to locate one or more SVM drizzled filter image(s) required for crossmatch!")
-                error_flag = True
-                break
             just_sl_name = os.path.basename(overlap_dict[bit_value]["svm_sourcelist_{}".format(set_num)])
 
             # 4: read in SVM-generated sourcelists and drizzled filter product images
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/mvm_quality_analysis.py
```diff
@@ -390,11 +390,6 @@ def overlap_crossmatch_analysis(total_obj_list, sourcelist_type="point", good_fl
                 log.warning("Unable to locate one or more SVM sourcelist(s) required for crossmatch!")
```
mdlpstsci, line 390: The two datasets I use for processing fail at line 389.  The problem is 
`    if overlap_dict[bit_value]["svm_sourcelist_{}".format(set_num)] is None:
KeyError: 'svm_sourcelist_0'
`
Since I pulled the SVM FLT/FLC and catalogs to my local directory, is there something else needed?

mdlpstsci (changes requested):
I am glad to see the changes were not as pervasive as I thought, and you were able to get "right onto" this problem.  I do have a problem processing my datasets which I can provide to you.

drizzlepac/haputils/mvm_quality_analysis.py
dulude: Issue _should_ have been resolved by commit [ebe4f8e](https://github.com/spacetelescope/drizzlepac/pull/1413/commits/ebe4f8ec593c3ddc1e4e56c56c0cb2043a31c8d2)

mdlpstsci (approved):
This ticket is approved as the problem of the source code needed to read an SVM-style poller file has been resolved.  The new failure mode reported will be documented in a new ticket as it introduced scope creep into this ticket.

PR #1414, opened 2022-07-15
Exit immediately if no new MVM inputs

stsci-hack:
This simple change converts an Exception to an immediate system exit so that the calling code can better understand the reason for quitting processing at this point.  Specifically, if there are no new MVM inputs to process, it should simply quit with a pre-defined exit code of 65.


mdlpstsci (approved):
This should work the same as the current sys.exit() invocations where the higher-level calling code catches the exception and the handles the fact the manifest filename has not yet been generated.

PR #1418, opened 2022-07-26
Update SVM to support WFPC2 subarray exposures

stsci-hack:
These updates revise the SVM processing code to support not only full 4-chip readout mode WFPC2 exposures, but also 'subarray' readout exposures where only some of the detectors are read out.  This adds a new 'detector' mode for WFPC2 data, 'wf', to compliment the default 'pc' which gets defined for 4-chip exposures.  

These changes were tested using data from "visit" **u2310k** which actually has a mix of readout modes for the exposures in the visit.

### DIFF
--- a/drizzlepac/haputils/hla_flag_filter.py
+++ b/drizzlepac/haputils/hla_flag_filter.py
@@ -34,20 +34,18 @@ Dependencies
 * ci_table.py
 """
 import glob
-import json
 import math
 import os
 import sys
 import time
 
 from astropy.io import fits as fits
-from astropy.table import Table
 import numpy
 import scipy
 import scipy.ndimage
 
-from drizzlepac.haputils import ci_table
 from stsci.tools import logutil
+from stsci.tools import fileutil
 from stwcs import wcsutil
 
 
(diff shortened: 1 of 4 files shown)


drizzlepac/haputils/hla_flag_filter.py
```diff
@@ -354,11 +353,11 @@ def hla_saturation_flags(drizzled_image, flt_list, catalog_name, catalog_data, p
         if channel.lower() in ['sbc', 'hrc']:
             image_ext_list = ["[sci,1]"]
         dq_sat_bit = 256
-    if channel.lower() == 'wfpc2':
```
mdlpstsci, line 357: Can you please add a comment explaining a bit the difference between instrument, detector, and channel as they are used in this code?  The wfpc2, in particular, is sometimes referenced as a channel and other times as an instrument.

stsci-hack: Some comments added in 4183f5e.

mdlpstsci (approved):
Thanks for the explicit comment.

PR #1419, opened 2022-07-28
Added try/except blocks for exceptions generated with regions do not have cross matches

mdlpstsci:
Added try/except blocks for a KeyError, generated for a dictionary, and an IndexError, generated for a 2D mask, where information may be missing as the regions do not overlap or do not overlap enough to have the data to populate the data structures.  Since this information is for quality analysis of the alignment, we want the evaluation to continue and skip over "inappropriate" regions. Tested with skycell-p1637x16y01 which has 24 input SVM images over three visits, four filters, and two detectors.


stsci-hack (approved):
These changes make sense to me.

PR #1420, opened 2022-08-02
Add Python 3.10 tests

stsci-hack:
This simply adds Python 3.10 tests to the suite run on each PR and build.

### DIFF
--- a/azure-templates.yml
+++ b/azure-templates.yml
@@ -10,12 +10,10 @@ jobs:
 
   strategy:
     matrix:
-      # Python310:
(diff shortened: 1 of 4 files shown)


stsci-hack:
The only reason for the failures turned out to be some parameters which had been deprecated in 3.9 were finally removed in versions available under 3.10.  These changes replaced the use of those deprecated parameters with ones which are backwards-compatible with 3.8/3.9.

mdlpstsci (approved):
There were more types of changes than I expected.  Everything looks good.

PR #1423, opened 2022-08-18
Support additional RADESYS options for input files

stsci-hack:
Some input data includes names of the coordinate system for the WCS from the REFFRAME keyword that are not supported by the FITS WCSLIB transformations.  These changes watch for any incompatible REFFRAME/RADESYS values and replaces it blindly with a value of 'ICRS' for the purposes of generating the sky coordinates of the segmentation catalog sources only.  Log messages are then generated to warn the user that such a change to the computation of the sky coordinates was made and that the output sky coordinates may not be entirely accurate.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1827,17 +1827,8 @@ class HAPSegmentCatalog(HAPCatalogBase):
                                                     kernel=self.kernel, wcs=self.image.imgwcs,
                                                     kron_params=[self._kron_scaling_radius, self._kron_minimum_radius])
 
-            # Guard against a non-standard coordinate system being defined in the input image headers
-            # If something non-standard is found, set the value to 'ICRS' to insure that the
-            # table conversion works.  Checking against the set of valid options recognized by WCSLIB and
-            # documented in http://ds9.si.edu/doc/ref/region.html#RegionFileFormat is necessary since the
-            # header keyword REFFRAME can be populated with anything specified by the user in the original
-            # proposal.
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -2174,6 +2188,17 @@ def measure_sources(self, filter_name):
...
+        # If something non-standard is found, set the value to 'ICRS' to insure that the
+        # table conversion works.  Checking against the set of valid options recognized by WCSLIB and
+        # documented in http://ds9.si.edu/doc/ref/region.html#RegionFileFormat is necessary since the
+        # header keyword REFFRAME can be populated with anything specified by the user in the original
+        # proposal.
+        if self.source_cat._wcs.wcs.radesys.upper() not in RADESYS_OPTIONS:
+            self.source_cat._wcs.wcs.radesys = 'ICRS'
+            log.warning(f"Assuming input coordinates are ICRS, instead of {self.source_cat._wcs.wcs.radesys}")
+            log.warning(f"Sky coordinates of source objects may not be accurate.")
+
```
mdlpstsci, line 2201: Solution looks fine, but since it has to be done twice...is it easy to make a little helper method do to this?

mdlpstsci (changes requested):
Solution is fine, but perhaps a smaller helper method would be a good idea (if possible).

drizzlepac/haputils/catalog_utils.py
stsci-hack: Implemented under a71cb83.

mdlpstsci (approved):
Looks great and streamlined!

PR #1424, opened 2022-08-22
Verify astrometry headerlets and remove bad ones

stsci-hack:
Some _a posteriori_ headerlet solutions retrieved from the astrometry database do not have the full distortion model solution attached as they should have which results in the omission of the distortion model when drizzling the affected exposure(s).  These changes look at all the headerlet solutions provided by the astrometry database to verify that those with WCSNAME containing '-FIT' also have a full distortion model and the corresponding CTYPE values.  This check gets performed before those solutions are used for alignment, so if any are found with problems, they are simply deleted from the input exposure so that the alignment can proceed from scratch.  

This allows the resulting image to be aligned correctly while also generating a properly distortion-corrected drizzled image and was tested on data from visit 'ie3501' where exposures from ASN `ie3501020` had incorrect GAIAeDR3 headerlets in the database.  

In addition, some Flake8 updates were also made to the affected modules as well.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -533,6 +533,9 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
             # have a priori solutions as extensions
             # FIX: This should probably only be done in the apriori sub-directory!
             updatewcs.updatewcs(_calfiles)
+            for _file in _calfiles:
+                confirm_aposteriori_hdrlets(_file, logfile=_trlfile)
+
             _trlmsg += "Adding apriori WCS solutions to {}\n".format(_calfiles)
             _trlmsg += verify_gaia_wcsnames(_calfiles) + '\n'
             _wnames_calfiles = [(c, fits.getval(c, 'wcsname', ext=1)) for c in _calfiles]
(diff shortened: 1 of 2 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -1726,6 +1733,44 @@ def update_active_wcs(filename, wcsname, logfile=None):
...
+    """Confirm that all the a posteriori headerlets are valid, and remove any that are invalid."""
+    update_msg = ""
+    num_sci_ext, extname = util.count_sci_extensions(filename)
+    extname_list = [(extname, x + 1) for x in range(num_sci_ext)]
+
+    hdu = fits.open(filename, mode='update')
+    # Get a list of all HDRLET extensions
+    hdrlet_extensions = wcsutil.headerlet.get_extname_extver_list(hdu, sciext='HDRLET')
+    hdrlet_wcsnames = wcsutil.headerlet.get_headerlet_kw_names(hdu, 'wcsname')
+
```
mdlpstsci, line 1747: Just making sure...is the order of results returned by get_extname_extver_list and get_headerlet_kw_names guaranteed to be the same? It seems like it would be, but I thought it was worth asking.

mdlpstsci (commented):
Just a question needs a response.

drizzlepac/runastrodriz.py
stsci-hack: The first function call, ``get_extname_extver_list``, only returns the tuples of all HDRLET extensions that were found in the FITS file by looping over all extensions in the FITS file in order using ``for extn in hdu``.  The second function also loops over the extensions in the FITS file in the same way, only appending results to the output list for HDRLET extensions.  As a result, the results should remain in lock-step with each other.    Good question.

mdlpstsci (approved):
Satisfied with the response to my query -- so ready to go!

PR #1426, opened 2022-08-24
Essentially remove restriction on PhotUtils package version

mdlpstsci:
Essentially remove restriction on PhotUtils package version, except to avoid old versions where there were many changes (avoid <= 1.2.0).


stsci-hack (approved):
I think this will be safe.

PR #1431, opened 2022-08-31
Insure catalogs can be successfully created for all sparse fields

stsci-hack:
These changes address a couple of issues related to catalog generation; specifically, 

- Fix logic for evaluating the segmentation catalogs to not try to use indices greater than the number of sources in the sparse field like the data from visit ``j8tq13`` and ``j96a02``.
- Only use a Table for point source catalogs when defaulting to DAOFindStar source finding such as for data from visit ``ich316``
- Keep formatting for ID as integer instead of reformatting as a float 

These revisions were tested using data which showed the original problems from visits:

- ich316
- j96a02
- ie3501
- j8tq13


mdlpstsci (approved):
Updates look fine.

PR #1436, opened 2022-09-12
Remove Python 3.7 tests from Jenkins builds

stsci-hack:
This simple change removes any Python 3.7 builds from the Jenkins builds used for testing our package.


mdlpstsci (approved):
Changes are straightforward and have been cleanly removed.

PR #1441, opened 2022-09-23
Prevent astroquery code from throwing an Exception for SVM requests

stsci-hack:
Some requests for SVM data using `astroquery_utils.retrieve_observation` will result in an Exception being thrown due to an attempt to try to move a file which has already been moved.  This simple change avoids that problem and allows the request to succeed normally.


mdlpstsci (approved):
Straightforward solution.

PR #1443, opened 2022-09-27
Use verify_guiding correctly for pipeline and SVM processing

stsci-hack:
These changes turn on the use of the `verify_guiding()` function from `haputils/analyze.py` in order to not try to align any exposure which appears (according to the algorithm) to have lost lock resulting in trailed stars.  The function was also upgraded to check the `QUALITY` keyword and ignore any exposure which has been marked with either 'GSFAIL' or 'TDF-DOWN'.  In addition, the logic for using this function was updated considerably in `runastrodriz` so that alignment is turned off when any exposure has been flagged as `BAD` by this function, and to not create any drizzle products if all exposures are found to be 'BAD'.  The logic, though, still allows the user to override these behaviors with the use of the `force` and `force_alignment` parameters. 

These changes were tested using data from ASN `jdtg08010` and from the singleton `j94rd3b0q`.   The ASN includes a single exposure which is trailed and flagged as `BAD` while the remaining exposures in the ASN have EXPTIME=0 due to the shutter remaining closed.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -188,7 +188,7 @@ wcs_preference = ['IDC_?????????-FIT_REL_GAIA*', 'IDC_?????????-FIT_IMG_GAIA*', 
 
 # Primary user interface
 def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
-            headerlets=True, align_to_gaia=True, force_alignment=False, do_verify_guiding=False, debug=False):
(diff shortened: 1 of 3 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -428,17 +436,32 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
     # parameter "do_verify_guiding".
```
mdlpstsci, line 436: Update the comment on Line 434 as this is no longer the case.

mdlpstsci (approved):
Changes look to be thorough.

PR #1444, opened 2022-09-27
Fix command line interface for runastrodriz

stsci-hack:
This simple change fixes the command line interface to `runastrodriz` in order to once again interpret the `-n` parameter input correctly.


mdlpstsci (approved):
Simple fix!

PR #1446, opened 2022-09-28
Implement update of WFPC2 header for best references in runastrodriz

stsci-hack:
These changes include:

- a new function `wfpc2Data.apply_bestrefs()`
- update to `runastrodriz.process()` to always call `wfpc2Data.apply_bestrefs()` for all WFPC2 input files
- only FLT files will be updated by `runastrodriz.process()` 
- only the distortion reference files will be updated by `runastrodriz.process()`

The new function uses CRDS to update the headers with the best reference files currently available for processing according to CRDS.  If a local CRDS cache is already defined, it will use that, otherwise it will create one in the current working directory then delete it when finished.

The code was tested using data from visit **u2lx05**.

### DIFF
--- a/drizzlepac/wfpc2Data.py
+++ b/drizzlepac/wfpc2Data.py
@@ -429,3 +429,28 @@ def wfpc2_to_flt(imgname):
         in_dq.close()
 
     return flt_filename
+
+
+# -----------------------------
+# Optional code for updating headers to latest
+# reference files from CRDS
+# -----------------------------
+def apply_bestrefs(uref_dir, dirname=None):
+
+    os.environ['CRDS_SERVER_URL'] = "https://hst-crds.stsci.edu"
+    os.environ['CRDS_PATH'] = "D:\data\crds_cache"
+    os.environ['CRDS_OBSERVATORY'] = "hst"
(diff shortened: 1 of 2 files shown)


drizzlepac/wfpc2Data.py
```diff
@@ -429,3 +430,129 @@ def wfpc2_to_flt(imgname):
...
+        Path for the ``CRDS_PATH`` directory on local system.
+        .. warning:: If no value is specified and ``CRDS_PATH`` is not
+        already defined locally, this function will create a temporary
+        ``CRDS_PATH`` directory tree under the directory with the files
+        to be processed, populate it with the latest reference files
+        and mappings needed by CRDS, then delete it when done.
+
+    """
+    starting_dir = os.getcwd()
+    import glob
```
mdlpstsci, line 475: This import be at the top of the module.

stsci-hack: Thanks for catching that!  Removed in 74dd1d7.

mdlpstsci (dismissed):
Code looks fine to me.

mdlpstsci (approved):
OK.  Looks good.

PR #1447, opened 2022-09-29
Add runastrodriz manifests

stsci-hack:
These updates add the following capabilities:

- `runastrodriz` now creates a manifest file listing the HLET, FLT, DRZ and trailer files updated and created during processing
- `hapsequencer` looks for any manifest files created by `runastrodriz` and merges their contents with it's own manifest file
- `hapsequencer` deletes `runastrodriz` manifest file if not run in diagnostic_mode
- adds a new parameter `-m` for `runastrodriz` to turn on creation of the manifest file
- deletes temporary D2IMFILE reference file created for WFPC2 data by `runastrodriz`

The deletion of the manifest files is necessary in order to support use in the automated pipeline.  It has been tested using data from visit 'u2lx05'.


raswaters (dismissed):

mdlpstsci (dismissed):
Changes look fine.

mdlpstsci (approved):
Initialization outside of the try block ensures the variables are defined - even upon immediate exception!

PR #1450, opened 2022-10-12
SVM: Fix logic for catalogs and cosmic ray checking

mdlpstsci:
The datasets quoted in this ticket which indicate they are good for testing a fix no longer even execute the apparent troublesome portion of the code.  The entirety of the drizzlepac code has changed a great deal in two years.  Since this has not been noted to have been an issue and no longer fails for the datasets suggested, I will make a modest change in the code which covers the case of zero sources. Note that zero sources should not be able to reach this portion of the code any longer.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -687,7 +687,7 @@ class HAPCatalogs:
                 all_sources = len(source_cat)
                 log.info("{} catalog with {} good sources out of {} total sources :  CR threshold = {}".format(cat_type, n_sources, all_sources, thresh))
                 # n_sources == 0 should never get here, but just to cover the past case of 0 < 0.0
-                if n_sources == 0 or n_sources < thresh:
+                if n_sources < thresh or n_sources == 0:
                     reject_catalogs = True
                     log.info("{} catalog FAILED CR threshold.  Rejecting both catalogs...".format(cat_type))
                     break
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -686,7 +686,8 @@ class HAPCatalogs:
                 n_sources = source_cat.sources_num_good  # len(source_cat)
                 all_sources = len(source_cat)
                 log.info("{} catalog with {} good sources out of {} total sources :  CR threshold = {}".format(cat_type, n_sources, all_sources, thresh))
-                if n_sources < thresh:
+                # n_sources == 0 should never get here, but just to cover the past case of 0 < 0.0
+                if n_sources == 0 or n_sources < thresh:
                     reject_catalogs = True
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -686,7 +686,8 @@ def verify_crthresh(self, n1_exposure_time):
                 n_sources = source_cat.sources_num_good  # len(source_cat)
                 all_sources = len(source_cat)
                 log.info("{} catalog with {} good sources out of {} total sources :  CR threshold = {}".format(cat_type, n_sources, all_sources, thresh))
-                if n_sources < thresh:
+                # n_sources == 0 should never get here, but just to cover the past case of 0 < 0.0
+                if n_sources == 0 or n_sources < thresh:
```
raswaters, line 690: Swap the two expressions, because n_sources == 0 is very unlikely to be true

raswaters (approved):

drizzlepac/haputils/catalog_utils.py
```diff
@@ -686,7 +686,8 @@ def verify_crthresh(self, n1_exposure_time):
                 n_sources = source_cat.sources_num_good  # len(source_cat)
                 all_sources = len(source_cat)
                 log.info("{} catalog with {} good sources out of {} total sources :  CR threshold = {}".format(cat_type, n_sources, all_sources, thresh))
-                if n_sources < thresh:
+                # n_sources == 0 should never get here, but just to cover the past case of 0 < 0.0
```
mcara, line 689: > but just to cover the past case of 0 < 0.0

I am not sure I understand this. How can 0 < 0.0? Is `n_sources` a floating point number? Should it be a floating point number? And if it is, do you mean that it could be a negative number?

mcara: What I see in the code is:

```python
source_cat.sources_num_good = len(np.where(crthresh_mask)[0])
```
on line 674 and on line 686 I see:

```python
n_sources = source_cat.sources_num_good # len(source_cat)
```

`n_sources` is an integer and I am not sure what case proposed changes deal with.

mdlpstsci: In the past a bug was tripped here because n_sources (with a value of 0) was NOT less than thresh (with a value of 0.0).  This issue will probably not ever be tripped again as no sources being found is trapped in other places in the code.  However, to be sure, I wanted to put in an evaluation which would evaluate properly on 0, but not change previous results.

mcara: Ah, it was `tresh` that was a floating point.

mcara (approved):

mcara:
Eh, my fault for merging this too soon. This PR needed a changelog entry

mcara:
because it is a bug fix

PR #1465, opened 2022-10-20
Cleanup SCI header (FLT/C, DRZ/C) keywords remaining from an 'a aposteriori' active WCS solution

mdlpstsci:
Check the SCI extension(s) of the output FLT/FLC and DRZ/DRC files.  If the active
WCS solution is 'a priori', there should not be any valid values for NMATCHES, RMS_RA/RMS_DEC, FITGEOM, and CRDER1/CRDER2 as these keywords are associated with 'a posteriori' solutions.  Ensure the WCSTYPE is based upon the active WCSNAME. Clarify some comments.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -1426,9 +1426,10 @@ def _verify_sci_hdrname(filename):
 def reset_keywords_apriori(filename):
     """
     Check the SCI extension(s) of the output FLT/FLC and DRZ/DRC files.  If the active
-    WCS solution is 'a priori', there should not be any valid values for NMATCHES,
-    RMS_RA/RMS_DEC, FITGEOM, and CRDER1/CRDER2 as these keywords are associated with 
-    'a posteriori' solutions.  Ensure the WCSTYPE is based upon the active WCSNAME.
+    WCS solution is 'a priori', delete the following keywords if they are associated
+    with the active WCS as they are residue from a previous 'a posteriori' solution:
+    NMATCHES, RMS_RA/RMS_DEC, FITGEOM, and CRDER1/CRDER2. Ensure the WCSTYPE is based
+    upon the active WCSNAME to clarify any confusion.
 
     Parameters
     ----------
@@ -1439,8 +1440,7 @@ def reset_keywords_apriori(filename):
     -------
     None
     """
-    wcs_keys = {'NMATCHES': 0, 'RMS_RA': 0.0, 'RMS_DEC': 0.0, 'FITGEOM': '', 
-                'CRDER1': 0.0, 'CRDER2': 0.0}
+    keys_to_delete = ['NMATCHES', 'RMS_RA', 'RMS_DEC', 'FITGEOM', 'CRDER1', 'CRDER2']
 
     # Determine the number of SCI extensions
     num_sci, extname = util.count_sci_extensions(filename)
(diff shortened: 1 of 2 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -1411,3 +1419,62 @@ def _verify_sci_hdrname(filename):
...
+        wcsname_list = active_wcsname.split('-')
+
+        # If the WCS solution is 'a priori' ensure any keywords in wcs_keys are set to default values
+        is_apriori = len(wcsname_list) == 2 and wcsname_list[1].upper().startswith(('GSC', 'HSC', 'GAIA'))
+        if is_apriori:
+            try:
+                # Make sure the WCSTYPE is consistent with the active 'a priori' WCSNAME
+                if hdu[ext].header['WCSTYPE']:
+                    hdu[ext].header['WCSTYPE'] = updatehdr.interpret_wcsname_type(active_wcsname)
+            except KeyError:
```
raswaters, line 1467: Seeing a except followed by a pass without a message of some kind makes me worried that one could hide errors. Is it okay for WCSTYPE not to exist?

drizzlepac/hapsequencer.py
```diff
@@ -1411,3 +1419,62 @@ def _verify_sci_hdrname(filename):
...
+                pass
+
+            # Loop over the keyword values for this particular solution to update
+            for wkey, wvalue in wcs_keys.items():
+                try:
+                    if hdu[ext].header[wkey]:
+                        hdu[ext].header[wkey] = wvalue
+                except KeyError:
```
raswaters, line 1475: The goal here is to set the keywords listed in wcs_keys to some value that signals that indicates that the values are not valid. If we get a KeyError, that means the keywords are missing altogether, which definitely indicates the values are not valid.

drizzlepac/hapsequencer.py
mdlpstsci: There will be some instances where WCSTYPE does not exist, and this seems to be fine.  Having said this,  
1. it would not hurt anything to be present, and
2. it is informative.
If it is missing, I will add the keyword.

drizzlepac/hapsequencer.py
mdlpstsci: In truth, the alternative to setting the keywords to defaults values is to remove the keywords entirely.  They only are relevant for 'a posteriori' situations and only will exist for 'a priori' situations if there is a roll-back from an 'a posteriori' solution back to an 'a priori' solution.  This is only done when forcing direct and Grism images to have the same 'a priori' WCS solution.

If the keywords are missing, I am deliberately am not adding these keywords to the 'a priori' solution.  I will add comments to the code to address this issue.

drizzlepac/hapsequencer.py
```diff
@@ -1411,3 +1419,62 @@ def _verify_sci_hdrname(filename):
...
+
+        # If the WCS solution is 'a priori' ensure any keywords in wcs_keys are set to default values
+        is_apriori = len(wcsname_list) == 2 and wcsname_list[1].upper().startswith(('GSC', 'HSC', 'GAIA'))
+        if is_apriori:
+            try:
+                # Make sure the WCSTYPE is consistent with the active 'a priori' WCSNAME
+                if hdu[ext].header['WCSTYPE']:
+                    hdu[ext].header['WCSTYPE'] = updatehdr.interpret_wcsname_type(active_wcsname)
+            except KeyError:
+                pass
```
mcara, line 1468: `WCSTYPE` is optional keyword not part of FITS WCS standard. `try-except` is used to catch the case when `WCSTYPE` does not exist. Don't you want to update/set `WCSTYPE` anyway if you know what it is? So, if `try-except` deals with missing `WCSTYPE`, what is the purpose of `if hdu[ext].header['WCSTYPE']`? What happens when `WCSTYPE` is present in the header but it is an empty string `''`?
```suggestion
        if is_apriori:
            # Make sure the WCSTYPE is consistent with the active 'a priori' WCSNAME
            if 'WCSTYPE' in hdu[ext].header:
                hdu[ext].header['WCSTYPE'] = updatehdr.interpret_wcsname_type(active_wcsname)
```

drizzlepac/hapsequencer.py
mdlpstsci: My thoughts exactly!  I have already made this change, but I am running a test before the push.

drizzlepac/hapsequencer.py
mcara, line 1475: I do not understand the purpose of this. It seems to me that this code does not do what the comment in the function description says:

```
If the active WCS solution is 'a priori', there should not be any valid values for NMATCHES,
RMS_RA/RMS_DEC, FITGEOM, and CRDER1/CRDER2 as these keywords are associated with 
'a posteriori' solutions.
```
So, instead of setting these keyword values to some obviously wrong values, I think tyou should delete them as the statement above says:
```suggestion
            # Loop over the keyword values for this particular solution to update
            for wkey, wvalue in wcs_keys.items():
                if wkey in hdu[ext].header:
                    del hdu[ext].header[wkey]
```

mdlpstsci: Deleting the keywords was offered as a solution, but INS requested they be set to default values instead.

mcara: Sigh... But then the comment below the function declaration is wrong as it does not describe what this function is actually doing.

mcara: And these keywords do not make any sense for the drizzled images. These values are different or each input image.  How does one combine this into a single value to be written to a drizzled image, a priori or not? It just feels wrong.

mcara: if INS wants the keywords at all times and regardless of the content, then you do not need `try` at all. This is much simpler:

```python
hdu[ext].header.update(wcs_keys)
```

mackjenn: Deleting the keywords is an acceptable solution.  The goal was simply to not confuse users with fit rms values, etc when there was in fact no catalog fit.

drizzlepac/hapsequencer.py
mdlpstsci: Updated code in commit  63934d7.

drizzlepac/hapsequencer.py
mdlpstsci: Updated code in commit  63934d7.

raswaters (dismissed):

drizzlepac/hapsequencer.py
```diff
@@ -1411,3 +1419,68 @@ def _verify_sci_hdrname(filename):
...
+
+            # Loop over the keys_to_delete list
+            for key in keys_to_delete:
+                try:
+                    if key in hdu[ext].header:
+                        del hdu[ext].header[key]
+
+                # Log a problem, but keep going as this is just a bit of cleanup
+                except KeyError:
+                    log.warning(f'Could not delete the {key} keyword from the active WCS solution.')
```
mcara, line 1482: There is no need for `try-except` here since there cannot be any exceptions since you are checking that the keyword exists in the header before deleting it. Please remove the try block.
```suggestion
            for key in keys_to_delete:
                if key in hdu[ext].header:
                    del hdu[ext].header[key]
```

mcara (approved):

PR #1469, opened 2022-11-10
Update wfpc2 shadow mask

stsci-hack:
The 'shadow mask' refers to the mask used to ignore those pixels from each chip which falls under the shadow of the WFPC2 pyramid.  This mask gets used when drizzling the data and can either be made from the DQ array or from a calibrated function that describes that shadowed region.  The changes in this code allow the code to properly work with the new ACS-style FLT format for the WFPC2 data as generated when updating the reference files early in astrodrizzle processing.  This allows the current logic to work correctly based on the DQ array as opposed to using the default functional form when the code was originally looking for the DQ array in the c0m file which was not provided as input to astrodrizzle.  

The logic was also updated to used the shadow mask defined by the calibrated functions to increase the overlap between chips by slightly shrinking the region flagged in the shadow mask region with DQ=2.  

The changes were tested using data from 'uba301', as well as several other visits.

### DIFF
--- a/drizzlepac/buildmask.py
+++ b/drizzlepac/buildmask.py
@@ -39,13 +39,19 @@ from stsci.tools.bitmask import bitfield_to_boolean_mask
 
 from astropy.io import fits
 import numpy as np
+from scipy import ndimage
 
 from . import processInput, util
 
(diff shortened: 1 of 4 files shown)


mdlpstsci (approved):
While complicated, the code follows what was described in our Slack conversation.

PR #1470, opened 2022-11-23
Turn on CR classify in align for WFPC2

stsci-hack:
These changes simply turn on the CR identification step within the alignment processing to ignore CRs when aligning WFPC2 data.   This corrects the problems with aligning WFPC2 data seen in testing, as confirmed by successfully aligning data from these visits:

- u2m301
- u27817
- u2lx05
- u55i03


mdlpstsci (approved):
Straightforward change to parameter default setting for WFPC2.

PR #1471, opened 2022-11-23
Implement photometric equalization for WFPC2 pipeline

stsci-hack:
This simple change applies photometric equalization using `drizzlepac.photeq` to all WFPC2 exposures during standard pipeline processing with `runastrodriz`.


mdlpstsci (approved):
Modification is as described.

PR #1474, opened 2022-12-05
Rename WFPC2 pipeline DRZ products

stsci-hack:
This is a simple update to reset the suffix for WFPC2 products generated using `runastrodriz` (standard pipeline-processed products) from 'ipppssoot_drz.fits' to 'ipppssoot_drw.fits' in order to prevent these new, WCS-updated products from overwriting the statically-archived WFPC2 products.  
These changes were tested to insure that this name change does not impact the output product names created from SVM processing.

### DIFF
--- a/drizzlepac/processInput.py
+++ b/drizzlepac/processInput.py
@@ -550,18 +550,13 @@ def process_input(input, output=None, ivmlist=None, updatewcs=True,
     asndict['original_file_names'] = origflist
 
     # Build output filename
-    drz_extn = '_drz.fits'
-    for img in newfilelist:
-        # special case logic to automatically recognize when _flc.fits files
-        # are provided as input and produce a _drc.fits file instead
-        if '_flc.fits' in img:
-            drz_extn = '_drc.fits'
-            break
+    # special case logic to automatically recognize when _flc.fits files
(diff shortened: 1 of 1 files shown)


drizzlepac/processInput.py
```diff
@@ -557,6 +557,11 @@ def process_input(input, output=None, ivmlist=None, updatewcs=True,
         if '_flc.fits' in img:
             drz_extn = '_drc.fits'
             break
```
mdlpstsci, line 559: Do you really want a "break" here on Line 559 and then again in Line 564?  It looks like there is a whole list of image names to process.

stsci-hack: The assumption, as you can tell, is that all the inputs will have the same suffix (flc.fits vs flt.fits) on input.  This code gets used (almost?) exclusively for either a single observation or a single association, so that assumption will hold true.  As for WFPC2 data, there is one type which makes this amount of logic overkill (come to think of it). the output from this function gets used to define the single output filename to be generated by AstroDrizzle, so we only need a single value.  Given that, it may just be easier to check the suffix of the first input member and be done with it rather than looking for an exception further down the list of inputs.

mdlpstsci (approved):
This logic flows better!

PR #1475, opened 2022-12-07
Update default WFPC2 HAP parameters

stsci-hack:
These updates to the default WFPC2 HAP parameters were submitted by INS for use in pipeline processing of WFPC2 data.


stsci-hack:
The line stating "self.conditions.append('wfpc2_wfpc2_any_n2')" should be right.  The first 'wfpc2' refers to the instrument (obvious), while the second one refers to the value of 'DETECTOR' which has been defined (at INS request) to be 'wfpc2' in #1473.  At least, that is what is intended by the naming convention for these config files and this code that interprets what config file should be used.

mdlpstsci:
(1) My concern is this string, 'wfpc2_wfpc2_any_n2', is not the actual filename, but rather it is a variable in a par file associated with the filename.  For example "wfpc2_wfpc2_index.json" contains 

        "wfpc2_wfpc_any_n2":"wfpc2/wfpc2/wfpc2_wfpc2_astrodrizzle_any_n2.json",

Is this all consistent?

(2) The tests are failing.

mdlpstsci (approved):

mdlpstsci (approved):
All is consistent.

PR #1476, opened 2022-12-12
Ignore headerlets for primary WCS solutions

stsci-hack:
This very simple fix allows the code to not try to delete invalid alternate WCS solutions if they are also the primary WCS.  Some headerlets from the astrometry database apparently do not have all the distortion keywords saved in them, and if they end up being used as the PRIMARY WCS, they should not get deleted since those keywords were already added by updatewcs during processing.  

This fixes the problem found with reprocessing data from idb645030, idh119010, idl330030, idxxa3030.

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -1804,7 +1804,6 @@ def confirm_aposteriori_hdrlets(filename, logfile=None):
     hdrlet_extensions = wcsutil.headerlet.get_extname_extver_list(hdu, sciext='HDRLET')
     hdrlet_wcsnames = wcsutil.headerlet.get_headerlet_kw_names(hdu, 'wcsname')
 
-
     invalid_extns = []
     for wname, extname in zip(hdrlet_wcsnames, hdrlet_extensions):
         # Only evaluate a posteriori headerlets -- those with '-FIT' in WCSNAME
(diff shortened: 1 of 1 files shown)


mdlpstsci (commented):
Sorry, I am not sure how this addresses the issue.  Could you explain through a simple illustration how this avoids deleting WCS solutions you do not wish to address?  Also, is there still a problem (different ticket?) as to why the WCS solutions are incomplete?

stsci-hack:
The underlying problem addressed by this PR starts when the headerlet for the a posteriori fit gets evaluated for 'valid_ctype' and 'valid_dist_kws'.  If either is False (i.e., invalid WCS), then the information for the alternate WCS associated with that headerlet is recorded here including the key for the alternate WCS.  The next block of code then checks to see whether there were any invalid WCS solutions, and if so, it loops over each one and deletes them from the header.  This PR insures that the if the WCS was found to be the PRIMARY WCS, the key value gets returned as an empty string instead of ' ' (a single space).  A value of an empty string will evaluate as False on line 1824, whereas the originally returned value of a single space would evaluate as True and try to delete the PRIMARY WCS and fail with the reported traceback.  
The cause of the problem being a problem with the headerlet would require updates to the database which would be way more difficult than this fix and may not be worth the effort.

mdlpstsci (approved):
With the explanation regarding the evaluation of the key associated with the primary WCS (null string vs a string containing one blank space), this change makes sense.

PR #1477, opened 2022-12-13
refactor for Tox 4

zacharyburnett:
- remove usage of `{homedir}` (not used in Tox 4)
- use `ruff` instead of `flake8` with extended ignored rule configuration (can fix issues later)
- copy `toxenv` structure from other repositories, and set up dependent job structure in workflow


stsci-hack (approved):
These changes look really good.  They finally bring this package up to date, better than I was able to do by myself!

pllim:
This is good progress but the coverage is 37%. This is because the actual heavy lifting is done in https://github.com/spacetelescope/drizzlepac/blob/master/JenkinsfileRT . True coverage should be calculated there. I think if you generate a token, you can get Jenkins to upload the coverage, though the way ITSD lock things down, I am not sure.

PR #1478, opened 2022-12-16
Updates to the SVM quality analysis to accommodate WFPC2

mdlpstsci:
Modified diagnostic_utils.py module to accommodate WFPC2, in particular for its filter keywords which are FILTNAM[1|2] vs FILTER[ |1|2] for ACS and WFC3, respectively.  Addeded keyword PRIMESI to the PHDU for WFPC2 data, and fixed a typo in a configuration JSON file. Finally, made graph_utils.py a bit more flexible as the interface now accommodates a "view" request.

### DIFF
--- a/drizzlepac/haputils/diagnostic_utils.py
+++ b/drizzlepac/haputils/diagnostic_utils.py
@@ -127,6 +127,9 @@ class HapDiagnostic(object):
             header_items_to_remove.append('FILTER2')
         if self.header['INSTRUME'] == 'WFC3':
             header_items_to_remove.append('FILTER')
+        if self.header['INSTRUME'] == 'WFPC2':
+            header_items_to_remove.append('FILTNAM1')
+            header_items_to_remove.append('FILTNAM2')
 
         header_patterns_to_remove = ["D\d+VER",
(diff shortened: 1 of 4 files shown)


stsci-hack (approved):
Changes make sense.  It raises the question whether or not the FILTNAM* keywords should have been renamed as FILTER* to make WFPC2 data more consistent with ACS and WFC3 data (something that can be considered later).

mdlpstsci:
Tested one of the failing tests, ibqk07.py, on my MacOS under Python 3.8, and the test passed ( 9 passed, 77 warnings in 253.29s (0:04:13) ).

mdlpstsci:
This PR has been approved, but it is still open as some PyTests are failing due to (1) unknown and (2) use of deprecated software of Photutils (devdeps).  The "unknown reason for failing" test was run on MacOS under Python 3.8 (as is the failure case), and the test ran just fine.  After discussion with Warren, we decided this should not stop the PR from merging.  The "deprecation" failure is for a function which changed signature, as well as moved into the SegmentationImage class as a method.  A fix for this has been discussed and will be covered in a different issue/PR.

PR #1479, opened 2023-01-04
Fix WFPC2 config file typos

stsci-hack:
These are simple changes to fix the syntax of a couple of WFPC2 config files based on errors reported during WFPC2 SVM testing.


mdlpstsci (dismissed):
This PR is approved pending further investigation for the failing PyTests.

PR #1480, opened 2023-01-04
Update required to compute_2d_background() in astrometric_utils.py 

mdlpstsci:
Modified compute_2d_background to accommodate API changes in PhotUtils while keeping this source code backwards compatible.  Modified the imports used in conjunction with OLD_PHOTUTILS.

Tested the "old" and "new" versions of compute_2d_background() by calling the function explicitly within a small test harness using hst_12572_52_wfc3_ir_total_ibug52_drz.fits as the input image.  The output images and float statistics were compared.  Note the specific portion of the code modified is invoked when the bkg = None and the resultant background and background_rms images are images of a single value.  The comparison between the "old" and "new" datasets is considered fine.

NEW
bkg_median: 0.5143258571624756   bkg_rms_median: 0.2762162387371063

OLD
bkg_median: 0.521289587020874   bkg_rms_median: 0.27636751532554626

### DIFF
--- a/drizzlepac/haputils/astrometric_utils.py
+++ b/drizzlepac/haputils/astrometric_utils.py
@@ -46,16 +46,14 @@ from astropy.time import Time
 from astropy.utils.decorators import deprecated
 
 import photutils  # needed to check version
+from photutils.segmentation import (detect_sources, SourceCatalog,
+                                    deblend_sources, SegmentationImage)
 if Version(photutils.__version__) < Version('1.5.0'):
     OLD_PHOTUTILS = True
-    from photutils.segmentation import (detect_sources, SourceCatalog,
-                                        deblend_sources, make_source_mask,
-                                        SegmentationImage)
+    from photutils.segmentation import make_source_mask
 else:
     OLD_PHOTUTILS = False
-    from photutils.segmentation import (detect_sources, SourceCatalog,
-                                        deblend_sources, detect_threshold,
-                                        SegmentationImage)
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/astrometric_utils.py
```diff
@@ -46,16 +46,18 @@
 from astropy.utils.decorators import deprecated
 
 import photutils  # needed to check version
-if Version(photutils.__version__) < Version('1.1.0'):
+if Version(photutils.__version__) < Version('1.5.0'):
     OLD_PHOTUTILS = True
-    from photutils.segmentation import (detect_sources,
-                                        deblend_sources, make_source_mask)
-    from photutils.segmentation import source_properties as SourceCatalog
+    from photutils.segmentation import (detect_sources, SourceCatalog,
```
stsci-hack, line 51: This may be trivial, but would it make this code a little easier to follow if the imports common to both the old and new versions were pulled out of the if statement.  You could have a statement like:

from photutils.segmentation import (detect_sources, deblend_sources, SegmentationImage) 

come before you even call the if statement, and then the imports inside the if statement would only import functions/classes specific to that version of the library.  Not critical to the code running, but more a maintenance issue perhaps.

mdlpstsci: Made the suggested change!

stsci-hack (dismissed):
Those changes look fine to me.

stsci-hack (dismissed):
Still LGTM.

PR #1481, opened 2023-01-04
Enable user specified CRDS URL for WFPC2 processing

stsci-hack:
This changes the code to use any user-specified CRDS_SERVER_URL from os.environ when applying bestrefs to WFPC2 data.  This will support use of test CRDS servers rather than always forcing the use of the public server.   In addition, the logic has been revised to also trigger a sync of the reference files to the local cache (or not) based on the 'CRDS_READONLY_CACHE' environment variable in order to support pipeline use of that variable. 

This code allows the CRDS code to crash with a status=-1 completely crashing out of any Python environment it was running in if the CRDS reference file cache or uref is not configured correctly BY THE USER if sync is turned off.

### DIFF
--- a/drizzlepac/wfpc2Data.py
+++ b/drizzlepac/wfpc2Data.py
@@ -494,10 +494,9 @@ def apply_bestrefs(filename=None, dirname=None,
 
     starting_dir = os.getcwd()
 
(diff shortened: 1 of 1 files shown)


stscirij (approved):
Looks good!

PR #1482, opened 2023-01-04
Documented how to read programmatically the catalog files

mdlpstsci:
Added documentation which indicates how to read programmatically the Point and Segmentation catalogs with Astropy or Pandas.


stsci-hack (dismissed):
That should be fine for any user wanting to use the data from these catalogs.

stsci-hack (approved):
LGTM.

PR #1483, opened 2023-01-05
Fix Numpy 1.24 str deprecation

stsci-hack:
These simple changes (as suggested by Numpy 1.20 Release Notes) should fix the problems with the pytests now that Numpy 1.24 has been released since Numpy 1.24 completely removed the deprecated use of 'np.str' in favor of 'np.str_'.


mdlpstsci (approved):
test-devdeps-xdist is still failing because the fix to the issue is in a different PR.  This is a chicken-egg problem only.

PR #1484, opened 2023-01-09
Avoid trying to delete non-existent readonly cache flag from environment

stsci-hack:
This change avoids treating the environment variable `CRDS_READONLY_CACHE` as a writable variable, since the code only needs to read it once.  Removing this from the list of variables read keeps this variable out of the list of variables that need to be restored to the original state so the code does not try to delete the non-existent variable.


mdlpstsci (approved):
Straight-forward update.

PR #1485, opened 2023-01-12
Reset crder units

stsci-hack:
These changes report the correct units for the CRDER* keywords now, using the astropy units already defined for them instead of trying to convert them manually.  Comments were also added to the keyword cards to specify the units of the term.  The units used for this keyword are now fully correct and consistent with the FITS Paper 1 standard and with the CUNIT* header keyword value.


mdlpstsci (approved):
Indeed, we should remember to use the facilities provided for unit conversion for consistency even when it seems so easy to do ourselves.

PR #1486, opened 2023-01-13
Fix logic in astroquery_utils to properly support regression testing

stsci-hack:
The logic used in `align.check_and_get_data()` failed when some of the files were already present and others were not.  The nightly regression tests rely on this function to make sure the necessary input files are present in the local directory for running the tests.  However, the ASN file was only used to get the specific FLT/FLC files needed for the tests, so they were not present for some tests while the FLT/FLC images were present.  The changes here made sure that if an ASN table was provided by the test as input, and as needed by astroquery to get the data, then that file now gets downloaded and used to make sure all members are present as requested.  If the members were already there, those local copies are still used and not downloaded again from the archive.  

In addition, more information is now provided to the log in the case that files were not found by default so that it is clear local files were eventually found and used.  

These changes allow the `tests/hap/test_apriori` tests to run successfully instead of throwing an Exception as done in the nightly regression tests.

### DIFF
--- a/drizzlepac/align.py
+++ b/drizzlepac/align.py
@@ -80,7 +80,6 @@ def check_and_get_data(input_list: list, **pars: object) -> list:
     candidate_list = []   # File names gathered from *_asn.fits file
     ipppssoot_list = []   # ipppssoot names used to avoid duplicate downloads
     total_input_list = []  # Output full filename list of data on disk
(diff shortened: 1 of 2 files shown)


mdlpstsci (approved):
Looks fine to me.

PR #1487, opened 2023-01-17
Update WFPC2 RT to work with new DRZ name

stsci-hack:
The WFPC2 nightly regression test needed to be updated to stay consistent with the change in WFPC2 drizzle product name when it changed from _drz.fits to _drw.fits.


mdlpstsci (approved):
Clear changes needed to support WFPC2 in a WFPC2-only source file.

PR #1488, opened 2023-01-18
Avoid running runastrodriz on calibration data

stsci-hack:
Only observations with the DRIZCORR header keyword value set to PERFORM should have 'updatewcs.updatewcs()' run on them, as the calibration files do not provide support for calibration or engineering modes.  This change fixes the problems caused by #1443 by moving that update and subsequent analysis of guiding inside the if statement for DRIZCORR.   
These changes were confirmed using data from visit 'u23v02'.


mdlpstsci (approved):
Straightforward changes.

PR #1489, opened 2023-01-18
Update to use photutils catalog wcs correctly

stsci-hack:
Photutils renamed the '_wcs' attribute for SourceCatalogs to simply 'wcs' as of photutils==1.6.0.  These changes not only look for that and use the correct attribute under 1.6.0, but also uses the old attribute name generated by 1.5.0 as well when creating the source catalogs for SVM products. 

In addition, the version of photutils being used for processing now gets reported along with the versions of numpy, stwcs and astropy since photutils is a major dependency for this package. 

These changes were tested using data from visit 'j8ep06' under photutils=1.5.0 and photutils=1.6.0.


mdlpstsci (approved):
Straightforward updates!

PR #1490, opened 2023-01-18
Handle empty images more gracefully

stsci-hack:
A couple of issues with empty images can still cause runastrodriz to crash.  

The first affects exposures where both filter names are empty (not listed as "BLANK") as found for '**u23v020nt**'.  The exposure was not able to get the full readout from the telescope resulting in an incomplete exposure with missing header information.  This causes `updatewcs.updatewcs()` to throw an Exception for an unsupported filter name. 

The second issue results from an empty image not returning any identified source segments when trying to verify whether or not the guiding was correct in `verify_guiding()` as triggered for data from visit '**u23105**'.  This resulted in an Exception reporting problems with " **'NoneType' object has not attribute 'nlabels'** ".  

These datasets were used to confirm the problem as well as the fact that these changes fix both problems.


mdlpstsci (approved):
Changes are targeted an appropriate.

PR #1491, opened 2023-01-18
Let photeq fail without crashing pipeline

stsci-hack:
Some WFPC2 exposures do not have their photometry properly calibrated causing the `photeq.photeq()` function to fail due to `PHOTFLAM` having a reported value of 0.0.  An example of such an exposure would be `u2l90105t` which uses an unusual polarizer filter, whereas other exposures in the visit using other filters process successfully.  These changes simply  trap the Exception from `photeq`, report the error in the log, then proceeds to process the image.   This should only affect the photometry of exposures taken with unsupported modes and will not affect the alignment or subsequent HAP processing of the data.


mdlpstsci (approved):
Straightforward change by adding a try/except.

PR #1496, opened 2023-01-24
Handle headerlet WCSNAME check correctly

stsci-hack:
The logic for identifying headerlet WCSs which are invalid did not account for the headerlet WCS from the database not being added as an alternate WCS.  In this case, an invalid WCS is a WCS that does not contain all the SIP keywords even though they are supposed to be a SIP WCS.  This simple change does not change the original logic except to handle a value of None correctly when identifying what alternate WCS the invalid WCS had been written out to, with a value of None meaning it wasn't.  

The data from ASN 'id8064020' has at least 1 invalid WCS and was used to confirm the problem and that this change fixed it.


mdlpstsci (approved):
Clean fix.

PR #1497, opened 2023-01-24
Insure source columns are not masked

stsci-hack:
Source position columns in the source catalogs come back as MaskedColumns, yet the masking can actually cause problems in some cases when the '_mask' attribute goes missing during processing.  Since we never use the mask attributes for any of our processing, we can avoid these problems by simply insuring that the source catalog never contains a MaskedColumn.  These changes convert all point source catalog columns to simple unmasked columns.  

The data from visit `idw101` originally had problems since some of the images were empty causing those catalogs to be MaskedColumns unnecessarily.  These changes were tested against that data to confirm that this allows that data to complete SVM processing.

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1111,7 +1111,9 @@ class HAPPointCatalog(HAPCatalogBase):
 
             log.info("Measured {} sources in {}".format(len(sources), self.image.imgname))
             log.info("   colnames: {}".format(sources.colnames))
-
+            # insure centroid columns are not masked
+            sources['xcentroid'] = Column(sources['xcentroid'])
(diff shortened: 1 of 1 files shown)


mdlpstsci:
There are a few places in the code where MaskedColumn is used (ci_col) for both Point and Segment catalogs.  Is this an issue?

stsci-hack:
The code only uses MaskedColumn objects to create completely new columns meant to be used as masks by the code to weed out bad sources.  The change here only insures that the 'xcentroid' and 'ycentroid' columns are never MaskedColumns, a change which did not cause problems with my test runs.

mdlpstsci (approved):
Clarification of MaskedColumns done -- the update looks fine.

PR #1498, opened 2023-01-25
Revise what HLET files get added to manifest file

stsci-hack:
Some headerlet files get written out by STWCS when they represent WCS solutions which are not present in the astrometry database, yet the filenames for those headerlets do not get passed along to the standard pipeline processing code `runastrodriz` to be included in the manifest file.  These changes look for those files and adds them to the manifest file along with the name of the headerlet file that was written out to represent the final WCS applied to the FLT/FLC file.  

These changes were tested against exposures from visit `u2l901` and resulted in 3 headerlet files being added to the manifest file that were not there before.


mdlpstsci (approved):
Changes look fine to me.

PR #1501, opened 2023-01-27
Fix default HLET filename

stsci-hack:
The filename used to write out the HLET file for the WCS solution used for the final runastrodriz product had been adding the WCSNAME to the filename without using the hash-tag when it should simply have been named using '_flt_hlet.fits'.  This corrects that mistake so that the simpler name gets used consistent with what gets written out to the manifest file.  

This change was tested using data from visit `u2311s`.


mdlpstsci (approved):
Straightforward change which looks fine.

PR #1505, opened 2023-02-06
Process all EXT WFPC2 targets

stsci-hack:
Some user-specified target names include terms like 'DARK', such as the targets for visit `u47m08` defined as "'NGC1976-DARK-STIS-EDGE".  Unfortunately, the original logic in `haputils.analyze.analyze_data()` registered any exposure with 'DARK' as any part of the TARGNAME keyword value as a calibration exposure that should be ignored and not processed.

The logic was updated to use the IMAGETYP keyword in conjunction with the TARGNAME for WFPC2 observations only to recognize such cases as true external science exposures that can be processed.  It was only implemented for WFPC2 since this problem has never been identified for any ACS or WFC3 exposures that we know of so far in order to minimize the risk for processing the ACS and WFC3 data.  

These changes were tested using data from visit `u47m08` to insure that this worked as desired.


mdlpstsci (approved):
Agree with the modest changes and ensuring only WFPC2 is impacted.

PR #1511, opened 2023-02-09
Turn off verify for WFPC2

stsci-hack:
The use of 'verify_guiding()' for WFPC2 images incorrectly recognizes diffraction spikes from saturated stars as evidence of loss of lock and flags those exposures as 'bad'.  This simple change turns off use of this function for WFPC2 data only, since loss of lock during an exposure was rarely an issue (if ever) and any such problems were reported in the QUALITY keywords which get checked as well.  

This was tested using data from visit `uba301`.


mdlpstsci (approved):
Straightforward change.

PR #1513, opened 2023-02-10
Update SkyCell docstring with examples

stsci-hack:
Simple update to the docstring for the SkyCell class to include examples of how to initialize this class interactively.


mdlpstsci (approved):
Important (!) information to know.

PR #1514, opened 2023-02-20
Delete WFPC2 CAL FLT files

stsci-hack:
These changes look for and remove any WFPC2 calibration exposures that can not be aligned or processed during standard pipeline alignment and drizzling.  This required updating the list of recognized calibration target names for WFPC2 and using that to identify those exposures to be skipped and deleted after converting the D0M images into FLT images.

The data from WFPC2 visit 'u2311s' include KSPOT and DARK exposures as well as 2 SCI exposures, and was used to confirm that these changes work as expected.


mdlpstsci (approved):
Changes look fine.

PR #1517, opened 2023-02-23
Removed the use of the ASTROMETRY_STEP_CONTROL environment variable

mdlpstsci:
Finally removed the use of the ASTROMETRY_STEP_CONTROL environment variable which had
been deprecated for some years as it had been replaced with ASTROMETRY_APPLY_APRIORI.  Only modest changes were necessary to clean up the deprecated variable.


stsci-hack (approved):
Changes look fine to me.  In fact, updating the test function really enforces this change.

PR #1518, opened 2023-03-01
Revise Jenkins tests to pass

stsci-hack:
These changes are intended to make the configuration of the CI testing more consistent with other packages while also passing all  steps implemented for testing.


mdlpstsci:
Closing this PR as the JenkinsfileRT has already been updated with the suggested change.  Also, the regression tests are now passing so there does not seem to be a need for a change to the pyproject.toml file.  For the junit_family, here is a discussion:
```
# Configures the format of the generated JUnit XML file. The possible options are:
#   xunit1 (or legacy): produces old style output, compatible with the xunit 1.0
#                       format. This is the default.
#   xunit2: produces xunit 2.0 style output, which should be more compatible
#           with latest Jenkins versions.


```

PR #1519, opened 2023-03-02
Update docstring for extract_sources and add default kernel

stsci-hack:
These are simple changes to the docstring for `astrometric_utils.extract_sources` along with an implementation for a default value for the `kernel` parameter.  Previously, `kernel` was actually a required parameter, yet it was specified with a default value of None implying it was optional.  Unfortunately, if `kernel` actually stayed as None, the code would throw an Exception.  So, some logic was added to compute a default kernel based on the `fwhm` parameter using code already called from other functions if `kernel` was None.

### DIFF
--- a/drizzlepac/haputils/astrometric_utils.py
+++ b/drizzlepac/haputils/astrometric_utils.py
@@ -835,10 +835,7 @@ def build_auto_kernel(imgarr, whtarr, fwhm=3.0, threshold=None, source_box=7,
         log.warning("Using a Gaussian 2D Kernel for source detection.")
         # Generate a default kernel using a simple 2D Gaussian
         kernel_fwhm = fwhm
-        sigma = fwhm * gaussian_fwhm_to_sigma
-        k = Gaussian2DKernel(sigma, x_size=source_box, y_size=source_box)
-        k.normalize()
-        kernel = k.array
+        kernel = build_gaussian_kernel(fwhm, npixels=source_box)
 
     return (kernel, kernel_psf), kernel_fwhm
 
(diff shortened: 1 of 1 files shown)


drizzlepac/haputils/astrometric_utils.py
```diff
@@ -948,10 +971,28 @@ def extract_sources(img, dqmask=None, fwhm=3.0, kernel=None, photmode=None,
...
+        [Deprecated] This parameter is not used.  In fact, it now gets computed
+        internally using the `sigma_clipped_bkg()` function which uses the
+        `dao_nsigma` parameter.
+    dao_nsigma : float
+        This number gets used to determine the threshold for detection of point
+        sources.  The threshold gets computed using a simple `mean + dao_nsigma * rms`,
+        where the `mean` and `rms` are computed from the background levels using
+        `astropy.stats.sigma_clipped_stats`.
     source_box : int
         Size of box (in pixels) which defines the minimum size of a valid source.
```
mdlpstsci, line 997: Just being picky ... Is this really the size of the box, is it the length of the side of a square box, or is the number of connected pixels which define a source?  OK, I know it is the last item listed.  It is just I tend to think of a box as an n x m quad.  I would clarify this parameter a bit.

mdlpstsci (changes requested):
Just  a small clarification of a parameter is requested.

drizzlepac/haputils/astrometric_utils.py
stsci-hack: See whether or not the small change in commit 30b6df6 clarifies the parameter well enough.

mdlpstsci (approved):
Code updates look fine to me.

PR #1532, opened 2023-03-14
Fixed some Deprecation and Syntax Warnings

mdlpstsci:
Fixed a small batch of DeprecationWarning and SyntaxWarning messages in a effort to begin to clean up the code.

### DIFF
--- a/drizzlepac/devutils/search_skyframes.py
+++ b/drizzlepac/devutils/search_skyframes.py
@@ -131,11 +131,11 @@ def make_search_string(arg_dict):
     """
     search_string = ""
     for item in arg_dict:
-        if not arg_dict[item] or item is "date_range":
+        if not arg_dict[item] or item == "date_range":
             continue
-        if item is "skycell":
+        if item == "skycell":
             substring = "{} == '{}'".format(item, arg_dict[item])
-        elif item is "spec":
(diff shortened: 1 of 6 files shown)


drizzlepac/devutils/search_skyframes.py
```diff
@@ -131,11 +131,11 @@ def make_search_string(arg_dict):
     """
     search_string = ""
     for item in arg_dict:
-        if not arg_dict[item] or item is "date_range":
+        if not arg_dict[item] or item == "date_range":
```
raswaters, line 134: Good catch, these incorrect "is". Makes me nervous that there might be more, so I grepped through the code base and identified some additional possible cases:
photeq.py:                if whdu is ref_hdu:
photeq.py:            if whdu is ref_hdu:
photeq.py:        if h is not _sh_log and isinstance(h, logging.FileHandler):
haputils/pipeline_graphics.py:            if self.Umask is not ma.nomask:

raswaters (approved):

drizzlepac/devutils/search_skyframes.py
mcara: those in `photeq` are correct.

mcara (approved):

PR #1533, opened 2023-03-23
Fixed the S_REGION bug - ensure the polygon is closed

mdlpstsci:
I fixed the bug where the S_REGION string which is setup for the pipeline FLT/FLC image is not a closed polygon.  The fix was to add the first set of coordinates to the end of the output string.


s-goldman (approved):
Looks good. I tried the code myself and it works as it should.

PR #1535, opened 2023-03-27
Update S_REGION keyword for standard pipeline drizzle products

s-goldman:
Added instance of compute_sregion for drizzled products. Tested for SVM and MVM products. 

Second attempt for this PR. First attempted auto-linted files, changing far more than desired.


mdlpstsci (changes requested):
I am not sure about your initial comment when you indicated this was tested for "SVM and MVM" products. The runastrodriz.py module creates the **_pipeline_** flt/flc and drz/drc products (i.e., ipppssoo* filenames).  What do you mean here?

While the s_region points overplotted on the image looked fine, it would be better to use connected lines rather than just points.  This is so one is not mislead by scattered points being connect in strange ways.  In truth, I do not think this is an issue here, but it would be good to redo the graphic from the Jira ticket with connected lines (just lines, no image).

This is just a request for a double-check before approval, otherwise the change looks fine.

s-goldman:
> 

Yeah, I think I meant multiple-exposure and multiple-visit (not MVM) drizzled products *still getting use to the nomenclature*. I added a line plot image to the Jira ticket and attached it below as well. 

Figure_1

mdlpstsci:
@s-goldman I am glad that you plotted the line graphic.  I am troubled by the line in the middle of the two chips.  Basically, we should just see the outline of the two chips with a bit of jagged corners as the two image files are offset from one another. There should only be one POLYGON IRCS in the S_REGION keyword, and it looks like this solution has two POLYGON ICRS components.  Can you check the S_REGION value for the drc file?

s-goldman:
> @s-goldman I am glad that you plotted the line graphic. I am troubled by the line in the middle of the two chips. Basically, we should just see the outline of the two chips with a bit of jagged corners as the two image files are offset from one another. There should only be one POLYGON IRCS in the S_REGION keyword, and it looks like this solution has two POLYGON ICRS components. Can you check the S_REGION value for the drc file?

Sorry, that was me being sloppy. That was for a single exposure where there was actually a gap that was being covered with the line. The multi-filter image (ida101010_drc.fits) is attached. I'll add that one to the Jira ticket. 

Figure_3

mdlpstsci:
Ah.  Much better!!!

mdlpstsci (approved):
Code update and validation has been satisfied well.

PR #1536, opened 2023-03-28
Ensure the SVM FLT/FLC files all contain a valid S_REGION

mdlpstsci:
Implemented an update to ensure the SVM FLT/FLC files all contain the S_REGION keyword AND it has valid values (i.e., closed polygons). Removed a print statement (cruft) in the compute_region() function.


mdlpstsci:
@s-goldman or @stscirij - Please review this PR.

mcara (approved):
Looks OK to me but why not have `copy_exposure()` method in `HAPProduct` and the rest of classes would inherit it?

PR #1537, opened 2023-03-29
Add the DATE keyword to the drizzle product primary header (HLA-961)

s-goldman:
I've added a header keyword to drizzled products. I use the current date (or date of creation) and added a header comment as well. The primary header of drz/drc products now looks like the following

The ipppssoot used for testing was iczg02a1q_flt.fits

Format looks consistent with the FLT file input.


mdlpstsci (changes requested):
The updates for the DATE keyword in the PHDU look fine.  However, we should not code smuggle and add a "hidden call" (from the specification of the ticket) to compute_sregion().  The call to compute_sregion() in runastrodriz.py needs to be removed from this PR.

When this is done, please remember to rebase as there may be other code updates which have been merged onto master while this issue has been "in play".

s-goldman:
The test that seems to be failing is in tests/hap/test_svm_hrcsbc.py within test_svm_samewcs. The test checks that products for both detectors are aligned to the same catalog. When looking for files with "total" and "fits" it doesn't seem to be finding any given the input data for the test. 

I'm not sure if this is related to the change i've made or another recent change.

mdlpstsci:
This is a real teaching experience!  Ugh.  It looks like the NEXTEND keyword is not present in all PHDUs so it should not be used to position the DATE keyword.  See the log around line 324.  I should have caught this during review - my bad.  It is not a mandatory/required FITS keyword, so this is not a violation of the standard.  

The WCS error could be a cascade because the previous PRs all passed.  I noticed where were also some connectivity problems, but this is not our fault.   

Put DATE after the NAXIS keyword as NAXIS is mandatory.

drizzlepac/outputimage.py
```diff
@@ -263,6 +264,7 @@ def writeFITS(self, template, sciarr, whtarr, ctxarr=None,
 
         # Start by updating PRIMARY header keywords...
         prihdu.header.set('EXTEND', value=True, after='NAXIS')
+        prihdu.header.set('DATE', time.strftime("%Y-%m-%d"), "date this file was written (yyyy-mm-dd)", after='NAXIS')
```
mcara, line 267:
```suggestion
        naxis = prihdu.header['NAXIS']
        prihdu.header.set('DATE', time.strftime("%Y-%m-%d"), "date this file was written (yyyy-mm-dd)", after='NAXIS' if naxis == 0 else f'NAXIS{naxis:d}')
```

mcara:
Actually, your latest version:

```python
prihdu.header.set('DATE', time.strftime("%Y-%m-%d"), "date this file was written (yyyy-mm-dd)", after='NAXIS')
```

would work for files created with `build=True` since those do not have data in the primary header (so there would be no `NAXIS1`, ...

mcara (approved):

mcara:
Would "date this file was created" be better than "date this file was written" or is this the same? [what if the file was "updated"/modified after it was first created?] I am just nitpicking. It's OK to leave as is.

PR #1542, opened 2023-04-14
Unit test for point kernel

s-goldman:
Simple unit test added for the c code point kernel. Several kernels are tested (e.g. gaussian, turbo, etc. ), and the test uses small csv truth files that still need to be added to either the local directory or on artifactory (open to suggestions).


tests/drizzle/test_drizzle.py
```diff
@@ -1,52 +1,211 @@
...
-def test_zero_input_weight(kernel):
+
+# testing set up
+class Get_Grid:
+    def __init__(self):
+        def get_wcs(_grid):
+            w = wcs.WCS()
+            w.wcs.ctype = ["RA---TAN", "DEC--TAN"]
+            w.wcs.crpix = [_grid[0], _grid[1]]
+            w.wcs.crval = [10, 10]
```
mdlpstsci, line 14: During our discussion with Warren, did he suggest the CRVAL values be larger (i.e., ˜100)?

s-goldman: I'm pretty sure that he was suggesting that the pixel being smoothed be brighter, so setting the selected insci pixel to 100 as opposed to 10. That being said, I don't really have a sense for how changing the crval changes the results.

stsci-hack: The CRVAL position specifies the RA/Dec value of the reference pixel in the image, where the reference pixel is the pixel in the input image where the distortion is zero.  The ref pixel is usually the center of the image.  In addition, I had originally suggested increasing the flux in the input pixel values from 10 to 100 or 1000 or even 10000 simply to make any splitting of the pixel's flux easier to recognize in the output frame.  This was only a suggestion, not a correction of any problem with the original test.

tests/drizzle/test_drizzle.py
```diff
@@ -1,52 +1,211 @@
...
+    point_function : Class
+        The Class inintialized in Get_Class which includes all of the inputs need to run cdriz.tdriz.
+    """
+    # add missing/flagged pixels in inwht
+    point_function.inwht[20:21, 22:23] = 0
+
+    # resample
+    cdriz_call(point_function, kernel)
 
     # check that no pixel with 0 weight has any counts:
```
mdlpstsci, line 208: If I have understood this correct, I would rephrase to: "Check that any pixel with 0 weight has no counts".

s-goldman: good catch, change made.

mdlpstsci (commented):
I have some questions which may be due to my being "a less as experienced Python programmer".
1) It is not clear to me that any output needs to be written out for any other reason than to ensure the test is working as designed.  Once this is established, it seems the assert can be accomplished with in-memory comparison.
2) I do not understand how point_function is a Class.
I have a few other questions along the same lines as Item 2.

tests/drizzle/test_drizzle.py
stsci-hack: That logic where a pixel with 0 weight has no counts only applies when the `fillval` parameter is set to 0.  I would actually suggest setting `fillval` to "0" instead of "INDEF" for this test for that very reason.

tests/drizzle/test_drizzle.py
s-goldman: Add CRVALS as decimals (ra/dec in decimal degrees)

s-goldman:
I've kept the previously name point_function class (now named kernel_params) and I've added a useful method to it. I then moved all the helper functions to a new file (test_setup.py) and I now have two files that contain tests, but both of them use the helper functions and kernel_params class in test_setup.py. 

The test_kernel.py file contains the tests that I made before using truth files. The test_cdriz.py tests are smaller unit tests that are looking at the output counts for small 3 x 3 grids, using different kernels. More to follow.

tests/drizzle/test_drizzle.py
mcara: I think looking at `outwht` is not that reliable (it _should_ work, in principle). I think, instead, one should test to see that:

1. `outwht` of bad pixels is ether exactly 0 or <1 for border pixels;
2. make a cluster of bad pixels with a high value (1e6). set the values of good pixels to 1 or some random value `0<pixel_val<1`, set `inwht` to 0 for these pixels, and then compute flux of the output image. it should be equal to the flux of all good input pixels.

s-goldman:
Still need to add code to automatically update truth files when tests fail.

tests/drizzle/test_drizzle.py
s-goldman: @mcara Thanks for the suggestions, these are really helpful. Ive adapted test_zero_input_weight into the second test that you mention, but I'm still working on understanding the results. 

I still don't understand why the output counts are around half of the input counts after drizzling, but I need to spend some time looking at the algorithm. 

For that first test suggestion, did you mean that bad border pixels with outwht set to zero should have outsci=0? if not, maybe you could elaborate. 

Thanks again!

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,140 @@
...
+        String associated with one of the c code point kernel options.
+    kernel_pars : Class
+        The Class inintialized in Get_Grid which includes all of the inputs need to run cdriz.tdriz.
+    new_truth : bool, optional
+        Flag, whether to save a new csv truth file from the output data, by default False
+    return_png: bool, optional
+        Flag, whether to create an output image map (png).
+    """
+    # truth filename
+    truth_filename = f"./tests/drizzle/truth_files/{kernel}_truth"
```
mdlpstsci, line 31: We probably should not hard-code specific filename paths.  Instead, construct the truth_filename by determining the installation directory (os.path.dirname(__file__) and then joining the installation directory with the specific kernel (whatever the current parameterized value is) and the literal text of the truth file ("_truth").  I think it will work out to something like I have noted here.

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,140 @@
...
+    # resample:
+    cdriz_setup.cdriz_call(kernel_pars, kernel)
+
+    if new_truth:
+        # save new truth file
+        cdriz_setup.save_array(kernel_pars.outsci, f"{truth_filename}.csv")
+
+    if return_png:
+        # save truth file as figure
+        cdriz_setup.generate_png(kernel_pars, f"{truth_filename}.png")
```
mdlpstsci, line 45: Again, construct the output directory/filename path.

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,140 @@
...
+        rtol=1e-5,
+    ), cdriz_setup.error_message(kernel_pars.outsci, f"{truth_filename}_new.csv")
+    # np.testing.assert_array_equal(kernel_pars.outsci, truth_array)
+
+
+def test_cdriz_edge(kernel_pars, kernel="gaussian", new_truth=False, return_png=False):
+    """Similar to test_point_kernel but looking at bright pixels at edge of field."""
+
+    truth_filename = f"./tests/drizzle/truth_files/edge_{kernel}_truth"
+    kernel_pars.insci[0, 21] = 100
```
mdlpstsci, line 64: Again, directory path construction to avoid hard-coding.

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,140 @@
...
+    assert np.allclose(
+        kernel_pars.outsci, truth_array, atol=1e-4
+    ), cdriz_setup.error_message(kernel_pars.outsci, f"{truth_filename}_new.csv")
+
+
+def test_cdriz_large(kernel_pars, kernel="gaussian", new_truth=False, return_png=False):
+    """Similar to test_point_kernel but looking at large pixel."""
+
+    truth_filename = f"./tests/drizzle/truth_files/large_sqaure_{kernel}_truth"
+    kernel_pars.insci[21:25, 22:26] = 100
```
mdlpstsci, line 81: Again, directory path construction to avoid hard-coding.  Also, I think you mean "square" and not "sqaure".

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,140 @@
...
+    ), cdriz_setup.error_message(kernel_pars.outsci, f"{truth_filename}_new.csv")
+
+
+def test_cdriz_non_symmetrical(
+    kernel_pars, kernel="gaussian", new_truth=False, return_png=False
+):
+    """Similar to test_point_kernel but looking at non-symmetrical pixel."""
+
+    truth_filename = f"./tests/drizzle/truth_files/nonsymmetrical_{kernel}_truth"
+    kernel_pars.insci[21:25, 22:23] = 100
```
mdlpstsci, line 100: Again, directory path construction to avoid hard-coding.

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,140 @@
...
+@pytest.mark.parametrize("kernel", ["square", "point", "turbo", "gaussian", "lanczos3"])
+def test_point_kernel(kernel, kernel_pars, new_truth=False, return_png=False):
+    """Function tests different c code point kernels (inputs already created on instantiation).
+
+    Parameters
+    ----------
+    kernel : str
+        String associated with one of the c code point kernel options.
+    kernel_pars : Class
+        The Class inintialized in Get_Grid which includes all of the inputs need to run cdriz.tdriz.
```
mdlpstsci, line 24: initialized

tests/drizzle/cdriz_setup.py
```diff
@@ -0,0 +1,141 @@
...
+        _set_kernel_pars.outsci,
+        _set_kernel_pars.outwht,
+        _set_kernel_pars.outctx,
+        1,  # uniqid
+        0,  # ystart
+        1,  # xmin pixel to start reading; fits so starts at 1
+        1,  # ymin pixel to start reading; fits so starts at 1
+        _set_kernel_pars.dny,  # _dny
+        1.0,  # pix_ratio
+        1.0,  # xscale; plate scale variatitons
```
mdlpstsci, line 77: variations

tests/drizzle/test_cdriz.py
```diff
@@ -0,0 +1,62 @@
...
+    kernel_pars.insci[3, 3] = 1e4
+
+    # resample:
+    cdriz_setup.cdriz_call(kernel_pars, kernel)
+
+    if return_png:
+        # save truth file as figure
+        cdriz_setup.generate_png(
+            kernel_pars, "./tests/drizzle/truth_files/cdriz_square.png"
+        )
```
mdlpstsci, line 38: Again, directory path construction to avoid hard-coding.

tests/drizzle/test_cdriz.py
```diff
@@ -0,0 +1,62 @@
...
+    assert np.allclose(np.sum(kernel_pars.outsci), 10000, 1e-7)
+
+
+def test_gaussian_kernel(kernel_pars, return_png=False):
+    """Same as above test but with gaussian kernel."""
+    kernel_pars.insci[3, 3] = 1e4
+    cdriz_setup.cdriz_call(kernel_pars, "gaussian")
+    if return_png:
+        cdriz_setup.generate_png(
+            kernel_pars, "./tests/drizzle/truth_files/cdriz_gaussian.png"
```
mdlpstsci, line 49: Again, directory path construction to avoid hard-coding.

mdlpstsci (changes requested):
Construct file paths and typos.

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # add missing/flagged pixels in inwht
+    kernel_pars.insci[20:22, 21:22] = 100
+
+    # resample:
+    cdriz_setup.cdriz_call(kernel_pars, kernel)
+
+    if new_truth:
+        # save new truth file
+        cdriz_setup.save_array(kernel_pars.outsci, f"{truth_filename}.csv")
```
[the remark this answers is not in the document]
s-goldman: The code doesn't actually use this new_truth flag with that name. The names for new truth files (if they differ from the old ones) is specified in the assert statements with a "_new" at the end.  

I've removed the above if statement and added try/except elements so that the code creates new truth files with the regular name (not with "_new") when the truth files are missing. This should make create new truth files easier.

tests/drizzle/test_kernel.py
s-goldman: good suggestions.

s-goldman: I've added the local path to the save_array and save_fig functions.

tests/drizzle/cdriz_setup.py
```diff
@@ -0,0 +1,138 @@
...
+    ----------
+    _set_kernel_pars : Class instance
+        an instance of the kernel_pars class.
+    _name : str
+        The name of the output png file.
+    """
+    # for generating truth files
+    import matplotlib.pyplot as plt
+
+    local_path = os.path.dirname("__file__")
```
mdlpstsci, line 117: Duplicate to line123.

s-goldman: fixed

tests/drizzle/test_cdriz.py
```diff
@@ -0,0 +1,62 @@
...
+    # add bright pixel at center
+    kernel_pars.insci[3, 3] = 1e4
+
+    # resample:
+    cdriz_setup.cdriz_call(kernel_pars, kernel)
+
+    if return_png:
+        # save truth file as figure
+        cdriz_setup.generate_png(
+            kernel_pars, "./tests/drizzle/truth_files/cdriz_square.png"
```
mdlpstsci, line 37: I had not expected for this much of a path to the file to still be needed.  I thought something more like "./truth_files/cdriz_square.png" would be needed to append to the os.path.dirname(dunderscorefiledunderscore)? If I am wrong, please ignore this comment for all the locations where generate_png() is invoked.

tests/drizzle/test_kernel.py
```diff
@@ -0,0 +1,138 @@
...
+    ----------
+    kernel : str
+        String associated with one of the c code point kernel options.
+    kernel_pars : Class
+        The Class inintialized in Get_Grid which includes all of the inputs need to run cdriz.tdriz.
+    return_png: bool, optional
+        Flag, whether to create an output image map (png).
+    """
+    # truth filename
+    truth_filename = f"./tests/drizzle/truth_files/{kernel}_truth"
```
mdlpstsci, line 28: Same comment here (and all the truth_filename construction code) with regard to still needing this much of a path name to the file.

s-goldman: I guess I'm not sure how to make this shorter using the _file_ variable without moving the truth files. There may be a way, but I might need help getting there.

tests/drizzle/test_cdriz.py
s-goldman: I've tried using the __file__ and it ends up putting the pngs in the main module directory.

tests/drizzle/test_kernel.py
mdlpstsci: For the failing tests, you need to do the following with the the truth_filename:

# truth filename
    truth_filename = f"truth_files/{kernel}_truth".     === No slash at the beginning and minimum path hardcoded.

    # Check to see if the specified file exists in the current working directory 
    print(f"path: {os.path.dirname(__file__)}")   === You must execute pytest as “pytest -s drizzle” to see this message
    if not os.path.exists(truth_filename):  === This will be false
        # Find where this module has been installed and determine the absolute path
        install_dir = os.path.dirname(__file__) === This is /Users/mdelapena/repos/drizzlepac/tests/drizzle/
        default_file = os.path.join(install_dir, truth_filename) === Constructs the full name (without the csv)
        truth_filename = default_file
        print(f"truth: {truth_filename}")

This is just so you can get the nearly full truth filename (no csv suffix) and have only the minimum filename hardcoded.  In particular, I sometimes want to run a subset of tests in the subdirectory.  I would copy, for example, the desired tests from the drizzle subdirectory to “mydrizzle” and then “pytest -s mydrizzle”.  The tests would fail because “drizzle” has been hardcoded to the truth filename.

Also, is there a reason not to have the “.csv” included in the original truth_filename?

s-goldman: Ahh okay, understood. I left the ".csv" off to pass the filename to both the save_png (.png) and save_file (.csv) functions.

s-goldman: I'm pretty sure that I understand what you were suggesting I do. 

I've created a new function in cdriz_setup that I use for the rest of the tests. The function takes in a relative path and a name, joins them both to each other and the absolute path, and then prints and returns the final path.

mdlpstsci (approved):
Cool.  Hard-coded paths are now removed!

PR #1546, opened 2023-05-02
HLA-1006: Black formatting of drizzlepac/tests

s-goldman:
Most of the changes are spacing and apostrophes, but this should make the code easier to read. 

[https://jira.stsci.edu/browse/HLA-1006](https://jira.stsci.edu/browse/HLA-1006)

### DIFF
--- a/tests/acs/test_acs_kernels.py
+++ b/tests/acs/test_acs_kernels.py
@@ -10,51 +10,52 @@ def pytest_generate_tests(metafunc):
     # called once per each test function
     funcarglist = metafunc.cls.params[metafunc.function.__name__]
     argnames = sorted(funcarglist[0])
-    metafunc.parametrize(argnames, [[funcargs[name] for name in argnames]
-            for funcargs in funcarglist])
+    metafunc.parametrize(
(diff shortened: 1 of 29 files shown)


mdlpstsci (commented):
Is there a way to configure Black to ignore single versus double quotes?  As long as a file is consistent in its use of the quotes (which I cannot guarantee here, but it looks like at least most of the files are like this), I am not keen on changing the single to double quotes in many files just because of a preference.  I am not necessarily against the conformity of it all, but I would like to know if this is configurable.

s-goldman:
> Is there a way to configure Black to ignore single versus double quotes? As long as a file is consistent in its use of the quotes (which I cannot guarantee here, but it looks like at least most of the files are like this), I am not keen on changing the single to double quotes in many files just because of a preference. I am not necessarily against the conformity of it all, but I would like to know if this is configurable.

I think the [configuration options](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html) are purposefully limited, but there are some options. I did see that there is a --skip-string-normalization, which ignores the double quotes. This could be a short-term solution. 

The only scenario in which the quoting could be an issue is double quotes within single quotes (or vice-versa).

PR #1560, opened 2023-05-16
HLA-783 [MVM]: Bad RA coordinates for skycells in the "North Pole" projection cell

s-goldman:
See discussion on Jira ticket. I have followed the suggestion made by Rick White to change the projection cell declinations of the first and last cell to be limited (from the previous values of 90 and -90) to 89.999999999997 at the highest declinations and -89.999999999997 at the lowest.  

The issue is only seen when looking at `hst_skycell-p2643x06y10' data in some versions of DS9 and when plotting the fits file using maplotlib. The issue now seems resolved when displaying the data with maplotlib.


rlwastro:
@s-goldman I compared the new `allsky_cells.fits` file to the old version and confirmed that the change is correct.  The new values at the dec = +/-90 ends of the table are correct, and nothing else in the file has changed.

I have not tried to run the code to check on the new behavior, but I do expect it to work.  Since you were able to confirm the flaky, ds9-version-dependent behavior with the previous parameters, the fact that you find it now works with software that failed before is good enough for me!  But if you want me to check the file, let me know where to find a copy of it.

s-goldman:
Thanks for reviewing that @rlwastro, and for providing such detailed descriptions of the problem and solution!

mdlpstsci (approved):
Rick suggested the initial modification and approves of the situation.

PR #1572, opened 2023-06-15
HLA-1029: Fix projection cell identification in overlapping regions

s-goldman:
I've added a +1 to the upper range so that the code appropriately indexes skycells. Previously a range(20:21) was being used to loop through cell 21, but this range yielded just [20] so skycells with y=21 were not being looped through. 

The code now identifies program iczge2* as being in skycell-p2111x08y04 and skycell-p2111x09y04, and now, skycell-p2036x01y21 and skycell-p2037x17y21 as well.


mdlpstsci (approved):

mdlpstsci:
The PyTests suffer from the mysterious problem we have often where tests fail when run on the test system, but they are fine when run locally.  We are still trying to address this issue.  In any case, both the author and I have successfully run the PyTests, do I will take the responsibility to allow the modest change (which does not even affect the tests) to merge.

PR #1573, opened 2023-06-24
Fix an incompatibility with np 1.25 in minmed

mcara:
Fixes a crash in minmed code due to incompatibility with numpy 1.25 dropping support for comparisons of arrays of different lengths. This PR fixes this.


mdlpstsci (approved):
I fixed this very same problem during Hack Day on Friday as part of a test to remove potentially astroquery from this package so I am sad my fix will not merge.  However, your fix is more succinct, so it is a better solution.

PR #1606, opened 2023-07-05
Remove use of astroquery from drizzlepac

mdlpstsci:
Use of the astroquery package via the drizzlepac astroquery_utils module has been removed.  As observed in other spacetelescope packages, a conflict between astroquery and Artifactory caused PyTests to fail in an inexplicable way.  The PyTests have been modified to use Artifactory.  The few PyTests which deliberately have their datasets defined based on a randomly selected dataset from an input file have been set to **skip**. These tests were used to derive statistics on the success rate of alignment for single visits, so they really are deprecated pieces of code.  Bits of other modules in drizzlepac had to be modified to accommodate the removal of astroquery.  

The pyproject.toml file was modified to force an upper limit on the Numpy version until incompatible issues can be resolved.


mdlpstsci:
The astroquery_utils.py module was not causing the problem with the PyTests as observed by this repo, as well as other repos. The functionality in the module is very handy for acquiring test data from MAST.

PR #1607, opened 2023-07-06
Added an upper limit to NumPy version temporarily

mdlpstsci:
Temporary change to pyproject.toml to add an upper limit to the NumPy version until deprecated functionality can be updated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,7 @@ runmultihap = 'drizzlepac.runmultihap:main'
 requires = ["setuptools>=61",
             "setuptools_scm[toml]>=3.4",
             "wheel",
-            "numpy>=1.18",
+            "numpy>=1.18, <=1.24",
             "astropy>=5.0.4",
             "markupsafe<=2.0.1"]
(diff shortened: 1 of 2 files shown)


mcara:
I do not see what is this fixing. If you plan to make a release of the `drizzlepac` (published on PyPi) with this restriction then it will work for users: they will not be able to upgrade to `numpy` 1.25. But what is the point in doing this since the incompatible code was already fixed on `master` and if you are going to make a release, `drizzlepac` will work with both `numpy` 1.24 and 1.25.

Maybe it makes sense to restrict `numpy` version in https://github.com/spacetelescope/stenv environment files? However, this will not prevent users (or `pip`) from updating `numpy`.

mdlpstsci:
Discussed Mihai's suggestions with Warren, the author and former maintainer of drizzlepac.  A release of drizzlepac is pending some decisions and timing of the HSTDP-2023.2.0 potential delivery and eventual release. In any case, it was decided adding an upper limit to the NumPy version was not a good idea.  This PR is being closed without the merging of any changes.

PR #1612, opened 2023-07-13
Allow group specifications as integers or tuples

mcara:
This PR fixes a bug reported in INC0189304 that, contrary to the documentation, setting `group` argument would result in a crash. In addition, now it is allowed to set `group` to a Python list of integers (_extension numbers_) and list of tuples of _extension name_ and _extension version_ (`(extname, extver)`).


s-goldman (dismissed):
Code looks good. If you could, however, please add explicitly in the docstring descriptions of the three inputs are and their allowed types, and what is returned.

mdlpstsci (approved):

PR #1617, opened 2023-07-17
HLA-1052: Updated WFPC2 drizzle parameters for HAP

s-goldman:
Updated JSON files for HAP WFPC2. *_any_1, _any_2, _any_3, _any_4, and _any_total added. *_any_6 removed.

### DIFF
--- a/drizzlepac/haputils/config_utils.py
+++ b/drizzlepac/haputils/config_utils.py
@@ -250,8 +250,10 @@ class HapConfig(object):
                         if n_exp > 1:
                             self.conditions.append("wfpc2_wfpc2_any_n2")
                     if self.hap_pipeline_name == 'svm':
-                        if n_exp in [2, 3]:
+                        if n_exp == 2:
                             self.conditions.append("wfpc2_wfpc2_any_n2")
+                        if n_exp == 3:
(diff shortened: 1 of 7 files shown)


mdlpstsci (dismissed):
The changes to the files here look fine.  When handling the configuration parameter files, you need to check the drizzlepac/pars/hap_pars/svm_parameters/wfpc2_wfpc2_index.json file to address any updates which may be needed.  In this case, the "*_any_n6" file should be removed.  Also, the drizzlepac/haputils/config_utils.py file should be checked.  It looks like the "*_any_n6" file was never used for wfpc2, so there are no changes required.  Finally, you indicate the "*_any_n6" file has been removed, but I do not see that Git knows about this.  Please just edit the index file and check Git knows about the removal of the file.

s-goldman:
> md

I've confirmed that *any_6 is removed from the codebase as well as the index file. I also added a line to the index file that I feel is necessary, but that I don't fully understand the current logic. 

`"wfpc2_wfpc2_any_n3":"wfpc2/wfpc2/wfpc2_wfpc2_astrodrizzle_any_n3.json",`
 
 Perhaps you can explain how these are used later today.

mdlpstsci (approved):

PR #1629, opened 2023-07-27
WFPC2 json file error due to trailing comma

s-goldman:
An unwanted trailing comma in the WFPC2 json index file is the likely culprit for a failure of the WFPC2 HAP processing. This fix was confirmed by doing the WFPC2 processing locally (make_wfpc2.py) on the data from visit "ua0t02".


mdlpstsci (approved):
Change is straightforward and maintainer has tested the update.

PR #1630, opened 2023-07-31
Updates to Segmentation catalog to increase source detection

mdlpstsci:
Changed the variable names for the general "biggest source" and "biggest source fraction" limits which were misleading, and added a new variable name to accommodate the RickerWavelet biggest source limit.
   OLD                       NEW
   _rw2d_biggest_source  ==> _biggest_source_limit
   _rw2d_source_fraction ==> _biggest_source_fraction_limit
    new                  ==> _rw_biggest_source_limit

These changes required updating the catalog parameter JSON files, the code which reads the parameter values, and parameter names in segmentation source support functions.

Clarified informational log messages when reporting the configuration limits in use for the Segmentation catalog.  Moved the computation of the ratio between the "Gaussian biggest source" and the "Rickerwavlet biggest source" to be computed, as well as applied, for Round 2. Added more informational messages to keep the user better informed.  Changed some limit values (though more changes will be needed for other detectors):

ACS/WFC ONLY
   negative_percent: 15.0 ==> 20.0
   rw2d_nsigma: 10.0  ==> 40.0


raswaters (approved):

mdlpstsci:
This ticket is obsolete and has been superseded by many other updates to the catalog generation code.

PR #1632, opened 2023-08-08
HLA-1060: Fix formerly working HAP PyTest in test_align.py

s-goldman:
As discussed in Jira, I've removed a test on "j8ura1jbq_flt.fits" which was ultimately failing due to a lack of matches. The checks that were suppose to tell us what was failing were also not working because the num_xmaches variable was not being re-instantiated every loop. Instead the previous value was being carried over allowing the checks to pass. I've made sure to set the check flags before using the continue function, in order to insure that the correct information is printed in the final summary of the run.  In the future, we should try to break up large functions like this to avoid reusing the same namespaces.


mdlpstsci (changes requested):
The code changes look great.  This looks to be a very old bug, so there might be changes seen in other regression test data.  

I am not sure why you removed the "no fit information" datafile from the test.  We will get a different answer, presumably, and have to update the output.  However, the code should now handle this case properly as is.  Including this file makes it a better test. (IMHO)

Comments?

s-goldman:
> The code changes look great. This looks to be a very old bug, so there might be changes seen in other regression test data.
> 
> I am not sure why you removed the "no fit information" datafile from the test. We will get a different answer, presumably, and have to update the output. However, the code should now handle this case properly as is. Including this file makes it a better test. (IMHO)
> 
> Comments?

I think I understand, and I think I agree that it would be better. You mean leaving the file in as an expected failure. I think I could try to do that as a stand-alone file in an additional different filelist, but I'm not sure I can test one expected failure within a set of successes. How does that sound to you?

mdlpstsci:
In real-time discussion, we decided to remove the troublesome dataset from the parameterized test entirely.  Two Jira tickets will be written: (1) new PyTest to focus on the processing of this one dataset, and (2) new PyTest to focus on just the troublesome image.

s-goldman:
@mdlpstsci I blacked the two files related to this PR and removed a couple unused imports and variables. Otherwise the logic of the code hasn't changed. 

When you find the time, let me know if it looks good to merge.

drizzlepac/align.py
```diff
@@ -398,14 +434,16 @@ def perform_align(input_list, catalog_list, num_sources, archive=False, clobber=
             table = alignment_table.extracted_sources[imgname]
 
             # Get the location of the current image in the filtered table
-            index = np.where(alignment_table.filtered_table['imageName'] == imgname)[0][0]
+            index = np.where(alignment_table.filtered_table["imageName"] == imgname)[0][
+                0
+            ]
```
mdlpstsci, line 440: Lines 437-439 represent one of the odd formatting instances which is just too silly.  Please fix this manually.

s-goldman: Ahh, agreed. Good catch.

tests/hap/test_align.py
```diff
@@ -8,148 +8,204 @@
...
+                "j8ura1j8q_flt.fits",
+                "j8ura1j9q_flt.fits",
+                "j8ura1jaq_flt.fits",
+                "j8ura1jfq_flt.fits",
+                "j8ura1jgq_flt.fits",
+                "j8ura1jhq_flt.fits",
+                "j8ura1jiq_flt.fits",
+                "j8ura1jjq_flt.fits",
+                "j8ura1jkq_flt.fits",
+            ],
```
mdlpstsci, line 168: Recall we agree to remove this dataset from the parameterized list so all the tests in this file will pass.  Then yo were going to create two new tickets: (1) process the entire dataset as a single visit, and (2) process the single "bad" file.

mdlpstsci (changes requested):
Just a few syntactic changes.  Also, remove one dataset from processing per our Drizzlepac Meeting discussion. Creating the new tickets is not part of this review.

s-goldman:
@mdlpstsci  I think I misunderstood the two new planned tests, but I understand now and fully agree. The other formatting changes have been made as well.

mdlpstsci (approved):
Changes look fine.

PR #1639, opened 2023-08-11
Drizzlepac: Updated PyTest test_svm_hrcsbc.py from causing failure on PR

mdlpstsci:
Updated PyTest test_svm_hrcsbc.py after testing on Linux, MacOS, and examining archive files.  After this work was done, it was justified to change the number of sources found in the SBC image(s).


mdlpstsci:
This PR is just a change to a single value, a clean of two lines of code, and a clarifying comment.

PR #1641, opened 2023-08-13
Remove updatewcs to make test consistent for all locales.

mdlpstsci:
Remove updatewcs to make test consistent for all locales.  This update was originially done on a different branch with many other fixes, but it needs to be applied standalone for the RT on master.

### DIFF
--- a/tests/wfc3/test_wfc3.py
+++ b/tests/wfc3/test_wfc3.py
@@ -4,8 +4,6 @@ from stsci.tools import teal
 import drizzlepac
 from drizzlepac import astrodrizzle
 
-from stwcs import updatewcs
-
 from ..resources import BaseWFC3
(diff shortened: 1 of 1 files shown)


nden:
Why is this update necessary? 
If `updatewcs` is not run where is drizzlepac getting the complete WCS from?

mdlpstsci:
I was letting the files use the activate WCS solution in the first extension which is 'IDC_2731450pi-FIT_SVM_GAIADR2'.  I noticed that sometimes (at least in some of the tests which I did) the updatewcs was turned off.  I can look for the messages for the previous tests runs.  I wanted to see if I got agreement by just making sure the same actions were happening on my local machine and on PRs.

mdlpstsci:
Work done on this PR is unnecessary and superseded by other changes.

PR #1645, opened 2023-08-15
Correctly fix the wcs preferential order REL > IMG, GAIA*3 > GAIA[2|1]

mdlpstsci:
Enforce the chosen order for the WCS preferential solution for all exposures in a dataset.  REL is better than IMG, and GAIA*3 is better than GAIA 2 or GAIA1.


drizzlepac/runastrodriz.py
```diff
@@ -187,8 +187,8 @@
 envvar_reset_idctab_name = "PIPELINE_RESET_IDCTAB"
 
 # Order of preference for common WCS solutions - brute force choice of the GAIA catalog preference
-wcs_preference = ['IDC_?????????-FIT_REL_GAIA*3', 'IDC_?????????-FIT_REL_GAIADR2', 'IDC_?????????-FIT_REL_GAIADR1',
-                  'IDC_?????????-FIT_IMG_GAIA*3', 'IDC_?????????-FIT_IMG_GAIADR2', 'IDC_?????????-FIT_IMG_GAIADR1',
+wcs_preference = ['IDC_?????????-FIT_REL_GAIA*3', 'IDC_?????????-FIT_IMG_GAIA*3', 'IDC_?????????-FIT_REL_GAIADR2',
```
raswaters, line 190: It would be useful to add a sentence describing the order, like what you wrote in the comment:
Enforce the chosen order for the WCS preferential solution for all exposures in a dataset. REL is better than IMG, and GAIA*3 is better than GAIA 2 or GAIA1.

mdlpstsci: Address in commit 94ce39d.

raswaters (approved):

PR #1646, opened 2023-08-15
Removed XFAIL marker as tests are xpassing

mdlpstsci:
Tests marked as **xfail** are actually passing (**xpass**), so removed the **xfail** marker.


mdlpstsci:
Just removing xfail markers.

PR #1648, opened 2023-08-17
replace astropy VCS checkout with link to wheel URL

zacharyburnett:
This matches the corresponding solution in `romancal`'s `requirements-dev.txt`


mdlpstsci (approved):
Looks great!

PR #1654, opened 2023-08-23
Unit test added for wcs_functions.py

s-goldman:
A unit test has been added to wcs_functions using a sample drc and flc fits header files.


drizzlepac/wcs_functions.py
```diff
@@ -1074,10 +1074,17 @@ def make_mosaic_wcs(filenames, rot=None, scale=None):
...
     # Compile list of WCSs for all chips from all input filenames
     hstwcs_list = []
+    
     for f in filenames:
-        hstwcs_list.extend([get_hstwcs(f, extnum) for extnum in get_extns(f)])
-    # Combine them into a single mosaic WCS
-    output_wcs = utils.output_wcs(hstwcs_list, undistort=True)
+        if ('flc.fits' in f) | ('flt.fits' in f):
+            hstwcs_list.extend([get_hstwcs(f, extnum) for extnum in get_extns(f)])
+            output_wcs = utils.output_wcs(hstwcs_list, undistort=True)
```
mcara, line 1081: Why is output WCS computed inside the loop? Seems unnecessary and expensive.

drizzlepac/wcs_functions.py
```diff
@@ -1074,10 +1074,17 @@ def make_mosaic_wcs(filenames, rot=None, scale=None):
...
+        elif ('drc.fits' in f) | ('drz.fits' in f):
+            hstwcs_list.extend([get_hstwcs(f, extnum) for extnum in get_extns(f)])
+            output_wcs = utils.output_wcs(hstwcs_list, undistort=False)
```
mcara, line 1084: same here

drizzlepac/wcs_functions.py
```diff
@@ -1074,10 +1074,17 @@ def make_mosaic_wcs(filenames, rot=None, scale=None):
 
     # Compile list of WCSs for all chips from all input filenames
     hstwcs_list = []
+    
     for f in filenames:
-        hstwcs_list.extend([get_hstwcs(f, extnum) for extnum in get_extns(f)])
-    # Combine them into a single mosaic WCS
-    output_wcs = utils.output_wcs(hstwcs_list, undistort=True)
+        if ('flc.fits' in f) | ('flt.fits' in f):
```
mcara, line 1079: `then` and `else` branches do the same thing. Why not combine this in the `if` clause:
```suggestion
        if ('flc.fits' in f) or ('flt.fits' in f) or ('drc.fits' in f) or ('drz.fits' in f):
```
(also, why bitwise operator instead of the logical one?)

s-goldman:
@mcara I've added your suggestions as well as a unit test that uses very small fits files, mostly just encapsulating the header information of a sample drc and flc file into fits files.

drizzlepac/wcs_functions.py
```diff
@@ -1069,15 +1069,34 @@ def make_mosaic_wcs(filenames, rot=None, scale=None):
...
         the merged composite WCS
     """
+    
+    # makes str into list of str
     if not isinstance(filenames, list):
         filenames = [filenames]
 
+    # checks if drz/drc or flc/flt identifies type with extension var
+    extension='None'
+    if np.all(['flc' in x or 'flt' in x for x in filenames]):
```
mcara, line 1079:
```suggestion
    if all(['flc' in x or 'flt' in x for x in filenames]):
```
no need to convert this to `numpy` arrays.

drizzlepac/wcs_functions.py
```diff
@@ -1069,15 +1069,34 @@ def make_mosaic_wcs(filenames, rot=None, scale=None):
...
-    # Combine them into a single mosaic WCS
-    output_wcs = utils.output_wcs(hstwcs_list, undistort=True)
+
+    # Generate output WCS based on input WCSs, and undistorts depending on extension. 
+    if extension == 'fl':
+        output_wcs = utils.output_wcs(hstwcs_list, undistort=True)
+    elif extension == 'dr':
+        output_wcs = utils.output_wcs(hstwcs_list, undistort=False)
+    else:
+        raise Exception('Invalid extension type. Must be either fl or dr.')
```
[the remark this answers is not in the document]
mcara: However, I would like to suggest a different approach that is not based on file extension
```python
hstwcs_list = []
from astropy.wcs.utils import is_proj_plane_distorted  # move to the top of the module
has_dist = False
for f in filenames:
    w = get_hstwcs(f, extnum)
    hstwcs_list.extend([w for extnum in get_extns(f)])
    has_dist = is_proj_plane_distorted(w)

output_wcs = utils.output_wcs(hstwcs_list, undistort= has_dist)
```

mcara (changes requested):

s-goldman:
@mcara I like those suggestion far better that what I had. Thanks, suggestions included.

mcara (dismissed):

drizzlepac/wcs_functions.py
```diff
@@ -1069,15 +1070,17 @@ def make_mosaic_wcs(filenames, rot=None, scale=None):
...
 
-    # Compile list of WCSs for all chips from all input filenames
-    hstwcs_list = []
-    for f in filenames:
-        hstwcs_list.extend([get_hstwcs(f, extnum) for extnum in get_extns(f)])
-    # Combine them into a single mosaic WCS
-    output_wcs = utils.output_wcs(hstwcs_list, undistort=True)
+    hstwcs_list = [get_hstwcs(f, extnum) for f in filenames for extnum in get_extns(f)]
+
+    # Generate output WCS based on input WCSs, and undistorts depending on extension.
```
nden, line 1080: I'm not sure about the meaning of `undistorts depending on extension`. Could you clarify the comment?

s-goldman: I will remove that part. 

The aim was to allow this to work for DRZ(C)s and FLT(C)s, but it turns out that much more code would have to change to make that a reality. For the moment, we are just using this as an opportunity to clean up this part of the code.

drizzlepac/wcs_functions.py
```diff
@@ -1069,15 +1070,19 @@ def make_mosaic_wcs(filenames, rot=None, scale=None):
...
-    for f in filenames:
-        hstwcs_list.extend([get_hstwcs(f, extnum) for extnum in get_extns(f)])
-    # Combine them into a single mosaic WCS
-    output_wcs = utils.output_wcs(hstwcs_list, undistort=True)
+    hstwcs_list = [get_hstwcs(f, extnum) for f in filenames for extnum in get_extns(f)]
+
+    # Generate output WCS based on input WCSs, 
+    # is_proj_plane_distorted checks for distortion correction and creates an undistorted 
+    # output WCS if that is the case
+    output_wcs = utils.output_wcs(hstwcs_list, undistort=is_proj_plane_distorted(hstwcs_list[0]))
```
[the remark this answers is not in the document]
s-goldman: I see what you mean. The change allows for the use of drz's for this function, whereas previously, that was not the case. 

Initially I was trying to add a tool for visualizing skycell and projection cell data back when I was still trying to understand that code and debug MVM issues. 

I'm okay with removing this particular change as it doesn't seem worth the risk at the moment.

PR #1667, opened 2023-09-15
SCSB-93: Changes to fix errors expected from Numpy 2.0

s-goldman:
All np.float_'s are changed to np.float64 as the former is being deprecated in Numpy2.0


mdlpstsci (approved):
Straightforward.

PR #1697, opened 2023-11-07
Initial setup for ADRs

s-goldman:
Initial setup of Architectural Design Records used to keep track of top-level thinking behind the code. These have the benefit of being stored with the codebase in a simple format, they are searchable, they don't need to be regularly altered (hopefully), and capture conversation had among developers.


mdlpstsci (approved):
This is an excellent new concept for the drizzlepac package - long overdue.

PR #1700, opened 2023-11-14
[WIP] HLA-1161: MagErrAp2 and other magnitude errors in HAP catalogs have incorrect negative values

s-goldman:
Resolves [HLA-1161](https://jira.stsci.edu/browse/HLA-1161)

This PR addresses two issues in the HAP catalogs. 

1. Magnitude errors are being calculated as negative numbers, and these are used for threshold calculations. The threshold calculations then result in bad sources with good flags. 
2. Negative fluxes result in magnitude errors that are incorrect

This PR adds two fixes: 

1. A threshold calculation using flux errors (that will never be negative) instead of magnitude errors. 
3. Magnitude errors calculated using negative fluxes are set to infinity

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/doc/ADRs.md
+++ b/doc/ADRs.md
@@ -1,4 +1,4 @@
-# Updating WCS in Headers
+# Updating WCS in Headers 11/14/23
 
 ## Context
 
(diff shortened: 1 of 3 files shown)


s-goldman:
Had to change the number of expected sources in the point source catalog for test_svm_segment_total_cat() in test_svm_ibqk07.py. Currently investigating whether that is linked to these changes.

PR #1710, opened 2023-12-05
HLA-1164: Disable code that changes HAP catalog flag values

### LINKED ISSUE #1706: Disable the code that changes HAP catalog flag values for saturated/faint sources
_Issue [HLA-1164](https://jira.stsci.edu/browse/HLA-1164) was created on JIRA by [Rick White](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rlw):_

There are sources in the HAP catalogs that are marked as saturated (`Flags` = 4 or 5) even though they are below the 5-sigma detection threshold.  Those sources are expected to have the `Flags` bit set for value 8 (source too faint).  So their flags should be 12 or 13 instead of 4 or 5.

This apparent conflict in the flag values was created deliberately by the [flag4and8_hunter_killer()](https://github.com/spacetelescope/drizzlepac/blob/61f998cf735b00d976faf37e7d4ed9bcc90d3a8f/drizzlepac/haputils/hla_flag_filter.py#L1618) function in the `haputils/hla_flag_filters.py` module.  That function is called from the [hla_saturation_flags()](https://github.com/spacetelescope/drizzlepac/blob/61f998cf735b00d976faf37e7d4ed9bcc90d3a8f/drizzlepac/haputils/hla_flag_filter.py#L583) function after the saturation flags have been added, with the explicit goal of removing the "too faint" flag 8 for sources marked as saturated.

That approach is inherited from the HLA catalog flagging, where it was added in May 2015.  For the HLA catalogs, the goal was to help with flagging spurious sources found near very bright objects using the "swarm" flag.  Sometimes those extremely bright objects are missing from the catalogs, which makes it impossible for the swarm flagging to work.

A detailed investigation (see [this innerspace page](https://innerspace.stsci.edu/x/51_MGg)) found that the vast majority of the sources this algorithm unflags are in fact spurious junk objects that do not belong in the catalog.  And in fact, the swarm flagging algorithm is not even implemented for the HAP catalogs.  The swarm flagging code is [commented out](https://github.com/spacetelescope/drizzlepac/blob/61f998cf735b00d976faf37e7d4ed9bcc90d3a8f/drizzlepac/haputils/hla_flag_filter.py#L160).

Changing the flags using the `flag4and8_hunter_killer` function is of at most limited benefit to the catalog even when the swarm filter is turned on.  It has zero benefit without the swarm filter.  And including those sources is reducing the HAP catalog quality.

The call to the `flag4and8_hunter_killer` function should be removed from the HAP code.

s-goldman:
Resolves [HLA-1164](https://jira.stsci.edu/browse/HLA-1164)

Closes #1706 

This PR removed a previous function that tried to rflag spurious sources near bright sources. After an investigation, it was found that the code was not improving the catalog as hoped (see [full discussion here](https://innerspace.stsci.edu/pages/viewpage.action?spaceKey=CSB&title=Saturation+flag+in+HAP+catalogs) or within the jira ticket). 

The result of this change should be slightly fewer catalog sources flagged as 4 and 5, and more with higher numbers (e.g. 13, 14+). 

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/drizzlepac/haputils/hla_flag_filter.py
+++ b/drizzlepac/haputils/hla_flag_filter.py
@@ -580,8 +580,6 @@ def hla_saturation_flags(drizzled_image, flt_list, catalog_name, catalog_data, p
             if saturation_flag[i]:
                 table_row["Flags"] = int(table_row["Flags"]) | 4
 
-        # phot_table_rows = flag4and8_hunter_killer(phot_table_rows, column_titles)
-
         if diagnostic_mode:
             phot_table_temp = phot_table_root + '_SATFILT.txt'
             phot_table_rows.write(phot_table_temp, delimiter=",", format='ascii')
--- a/drizzlepac/haputils/hla_flag_filter.py
+++ b/drizzlepac/haputils/hla_flag_filter.py
@@ -580,7 +580,7 @@ def hla_saturation_flags(drizzled_image, flt_list, catalog_name, catalog_data, p
             if saturation_flag[i]:
                 table_row["Flags"] = int(table_row["Flags"]) | 4
 
-        phot_table_rows = flag4and8_hunter_killer(phot_table_rows, column_titles)
+        # phot_table_rows = flag4and8_hunter_killer(phot_table_rows, column_titles)
 
         if diagnostic_mode:
             phot_table_temp = phot_table_root + '_SATFILT.txt'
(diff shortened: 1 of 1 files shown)


s-goldman:
I still need to update at least one truth files, for the regression tests to pass.

s-goldman:
> Would there be value in removing the call to flag4and8_hunter_killer but leaving the function in place?

It will always be saved in the github repository, but Rick and I agreed that it doesn't seem that it will be very helpful in the future.

PR #1729, opened 2024-02-08
HLA-993: Adding skycell keyword to pipeline and SVM products

s-goldman:
Resolves [HLA-993](https://jira.stsci.edu/browse/HLA-993)

This PR adds the skycell(s) that the image data fall within, for the pipeline products and SVM products. The keyword is added to both the FLT(C) and DRZ(C) science headers. I have added the card below the S_REGION keyword with entries separated by semi-colon and the "skycell-" strings removed to conserve space:

`SKYCELL = 'p1748x01y06; p1749x18y06' / Skycell(s) that this image occupies`

Note: In the case of FLC as opposed to FLT files, the FLC files have the added keyword, but the FLT files do not. 

**Checklist for maintainers**
- [x] added relevant label(s)

[Jenkins test](https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FDrizzlepac-Developers-Pull-Requests/detail/Drizzlepac-Developers-Pull-Requests/50/pipeline)


s-goldman:
I'm currently testing different datasets. ACS and WFC3 pipeline and SVM products run correctly. While the code also correctly runs for WFPC SVM products, there seems to be an issue with pipeline products. 


It seems that get_skycells ignores WFPC2 data because the science headers do not have the 'wcsname' keyword. This was done to ignore datasets with exptime=0, but now is hampering our ability to implement this code for WFPC2 pipeline products. 


I can either add a temporary "wcsname" to these products, or change the get_sky_cells() function.

mdlpstsci:
@s-goldman 
Recall that we cannot alter the original WFPC2 datasets in any way.  Can you please give me a WFPC2 dataset name so that I can remind myself of any other details?  Thanks.

mdlpstsci:
@s-goldman 
Thanks.  The original GEIS files associated with WFPC2 cannot be changed by our software and are not in the proper format/structure, nor do they have the keywords necessary for processing by our current software.  Recall that during SVM processing the *c0m.fits file is converted into a *flt.fits file (wfpc2Data.py) which conforms to the current software. 

I can see that the *flt.fits file for the dataset you gave me has the WCSNAME keyword.  Are you looking in the correct file?  Am I misunderstanding your question?

(caldp_20231201) bash-4.2$ fitsheader u5h1010ar_flt.fits -e 1 -k wcsname
# HDU 1 in u5h1010ar_flt.fits:
WCSNAME = 'IDC_ta81040lu-GSC240' / Coordinate system title  fitsheader

s-goldman:
> @s-goldman Thanks. The original GEIS files associated with WFPC2 cannot be changed by our software and are not in the proper format/structure, nor do they have the keywords necessary for processing by our current software. Recall that during SVM processing the *c0m.fits file is converted into a *flt.fits file (wfpc2Data.py) which conforms to the current software.
> 
> I can see that the *flt.fits file for the dataset you gave me has the WCSNAME keyword. Are you looking in the correct file? Am I misunderstanding your question?
> 
> (caldp_20231201) bash-4.2$ fitsheader u5h1010ar_flt.fits -e 1 -k wcsname
> 
> # HDU 1 in u5h1010ar_flt.fits:
> WCSNAME = 'IDC_ta81040lu-GSC240' / Coordinate system title fitsheader

It seems that the wcsname is not present at the time that the new code is called in [runastrodriz.py](https://github.com/spacetelescope/drizzlepac/blob/fc546c60ba7ad2b1f365e450cb059290ba958f79/drizzlepac/runastrodriz.py#L450). I can try to make the call after the wcsname is added, wherever that is.

s-goldman:
Yeah that seemed to do it. Thanks for the direction. I will check each case again and run a jenkins regression test, but the code should be ready to be reviewed.

drizzlepac/haputils/processing_utils.py
```diff
@@ -341,6 +344,56 @@ def compute_sregion(image, extname='SCI'):
...
 
+def add_skycell_to_header(image, total_obj_list=None, extname='SCI'):
+    """Determines the skycells for which the image falls within and adds the 
+    information to the science header.
+
+    Parameters
+    ----------
+    image : Astropy io.fits  HDUList object
+        Image to update with the skycell keyword in each extname extension.
+    extname : str, optional
```
mdlpstsci, line 355: Missing description of total_obj_list parameter.

drizzlepac/haputils/processing_utils.py
```diff
@@ -341,6 +344,56 @@ def compute_sregion(image, extname='SCI'):
...
+    # Find all extensions to be updated
+    numext = countExtn(hdu, extname=extname)
+    
+    for extnum in range(1, numext + 1):
+        sciext = (extname, extnum)
+        # only add skycell keyword if it does not already exist
+        if 'skycell' not in hdu[sciext].header:
+            if total_obj_list: # for SVMs
+                skycells = get_sky_cells([x.full_filename for x in total_obj_list[0].edp_list])
+            else: # for pipeline products
```
mdlpstsci, line 378: For line 377, if the SVM contains more than one detector, then total_obj_list will be a list of 2.  As such, it looks like the exposures belonging to the second detector will not be processed.  I think SVM HST_11665_06 would be a good test case here.

mdlpstsci (dismissed):
Please check the total_obj_list issue.

s-goldman:
@mdlpstsci I've addressed your comments. One thing that is still not clear to me is how the HST pipeline handles the FLT(C) files. Does the current setup allow for the FLT(C) files to be updated when creating the pipeline and SVM products. 

I ask because the current version of this change adds the SKYCELL keyword to the FLT(C)s during the pipeline product phase and not the SVM phase.

s-goldman:
I've moved the add_skycell_to_header() to it's own call in hapsequencer and now pass only the FLC(T) image filenames to the function. This has simplified the code and removed the need to use the total_image_list within the function. 


The code now passes the skycell_keyword to all FLT(C) and DRZ(C) files of interest. I have also tested the code for the case of multiple detectors. 


I would like to add a unit test for this code, but I would like to make sure this code is acceptable before adding it.

mdlpstsci (commented):
ipppssoot - See https://archive.stsci.edu/hlsp/ipppssoot.html

drizzlepac/hapsequencer.py
```diff
@@ -622,6 +622,11 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
                 log.warning("This Total Data Product only has Grism/Prism data and no direct images: {}".format(total_item.drizzle_filename))
                 log.warning("No SVM processing is done for the Grism/Prism data - no SVM output products are generated.")
                 product_list += [total_item.trl_filename]
+            
+            # Add skycell header keyword to SVM flt(c) files
+            # iterates for each total object, but passes full total object product list to the function
+            for image in total_item.edp_list:
```
mdlpstsci, line 628: The comment needs to be updated.  The iteration is over each exposure in this total_item, but only the actual exposure name is provided to the function.

drizzlepac/haputils/processing_utils.py
```diff
@@ -341,6 +341,53 @@ def compute_sregion(image, extname='SCI'):
...
+        hdu, closefits = _process_input(image_filename)
+    except:
+        log.error(f"Could not open {image_filename} during add_skycell_to_header. Exiting.")
+
+    # Find all extensions to be updated
+    numext = countExtn(hdu, extname=extname)
+    
+    for extnum in range(1, numext + 1):
+        sciext = (extname, extnum)
+        # only add skycell keyword if it does not already exist
```
mdlpstsci, line 371: This is just FYI and you do not have to change as I see the variable in question, _extnum_, has the same variable name in other routines in this module.  Technically, this variable would be more appropriately named _extver_.  An _extnum_ is just an extension number which is sequential (1, 2, 3, ...) in the FITS file.  An _extver_ is associated with an imset and this is what you are really updating.   The way this is referenced is [extname, extver] or ['sci', 1] or ['err', 5] where ['err', 5] could be the 15th extension.

drizzlepac/haputils/tests/test_processing_utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+@pytest.mark.parametrize(
+    "filename",
+    [
+        "sample_ipppssoot_flc.fits", # WFC3/UVIS
+        "sample_ipppssoot_flt.fits", # WFC3/IR
+        "sample_svm_flc.fits", # WFC3/UVIS SVM file
+        "sample_svm_flt.fits", # WFC3/IR SVM file
+        "sample_ipppssoot_flt_w_skycell_header_keyword.fits"
+    ],
```
mdlpstsci, line 17: These FITS files are large enough that they should be stored in Artifactory and downloaded when the test is done.

s-goldman: The data within the files is truncated to 10 x 10 arrays in each extension to limit the data file size. Do you still think that these should be put on artifactory? I wasn't sure if the goal was to only host files on artifactory even for unit tests.

drizzlepac/haputils/processing_utils.py
s-goldman: I like the suggestion and I've made the change anyway.

drizzlepac/haputils/processing_utils.py
```diff
@@ -341,6 +341,53 @@ def compute_sregion(image, extname='SCI'):
...
+    Parameters
+    ----------
+    image : str 
+        Image filename to update with the skycell keyword in each extname extension.
+    extname : str, optional
+        EXTNAME value for extension containing the header to be updated
+        
+    Returns
+    -------
+    Nothing; image header has been updated unless keyword already exists.
```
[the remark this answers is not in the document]
s-goldman: You're right, there may be a case where a small WCS change results in different skycells. I've rearanged the code to update the skycell header keyword when it already exists.

drizzlepac/hapsequencer.py
s-goldman: comment updated.

s-goldman:
After the most recent code change I've checked to make sure that the skycell keyword is created or updated for: SVM and pipeline products for each instrument, and a program with both UV/IR detectors.

drizzlepac/haputils/tests/test_processing_utils.py
s-goldman: The remaining data is primarily in the headerlets, which I could also remove.

s-goldman: Note for deleting headerlets: https://stwcs.readthedocs.io/en/latest/headerlet_tsr/source/hdrlet_tsr.html#delete-headerlet

mdlpstsci: Just an FYI.  Even though Git/GitHub can handle binary files, we need to be circumspect about how many binary (i.e., FITS files in our case) we put into the repository.  When someone downloads the repo, they will have to download potentially large binary files, and this is an inconvenience.  Having said this, I think small files may be fine as long as there are not too many of them.

drizzlepac/runastrodriz.py
```diff
@@ -463,6 +463,10 @@ def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
         if _calfiles_flc:
             updatewcs.updatewcs(_calfiles_flc, use_db=False, checkfiles=False)
 
+        # adds skycell keyword to science header of all flt(c) and drz(c,w) files. 
+        for filename in _calfiles+_calfiles_flc:
```
mdlpstsci, line 467: Please add a note here that indicates the SKYCELL value for IPPPSSOOT and SVM products may be different as the current computation is based upon the WCSNAME of the input images.

drizzlepac/haputils/processing_utils.py
```diff
@@ -341,6 +341,55 @@ def compute_sregion(image, extname='SCI'):
     if closefits:
         hdu.close()
 
+def add_skycell_to_header(image_filename, extname='SCI'):
+    """Determines the skycells for which the image falls within and adds the 
+    information to the science header.
+
```
mdlpstsci, line 347: Please add a note here that indicates the SKYCELL value for IPPPSSOOT and SVM products may be different as the current computation is based upon the WCSNAME of the input images.

This is redundant with the comment in runastrodriz.py but we need to make sure it is well-documented.

mdlpstsci (changes requested):
Please just add the requested two comments.  Even though I am out after lunch, if you do this by Monday, I will re-review so this ticket can be completed!

mdlpstsci (dismissed):

s-goldman:
> Please just add the requested two comments. Even though I am out after lunch, if you do this by Monday, I will re-review so this ticket can be completed!

Comments added!

PR #1731, opened 2024-02-12
HLA-1203: Bug fix for HCALDP Reunion 

s-goldman:
This PR should address a deprecation warning in the building of drizzlepac for the recent HCALDP Reunion branch build. The error is:

```
 writing /srv/jenkins/tmp/pip-modern-metadata-lp4ht0h1/drizzlepac.egg-info/PKG-INFO
 /srv/jenkins/tmp/pip-build-env-w6qtofwu/overlay/lib/python3.11/site-packages/setuptools/_core_metadata.py:157: SetuptoolsDeprecationWarning: Invalid config.
  !!

          ********************************************************************************
          newlines are not allowed in `summary` and will break in the future
          ********************************************************************************

  !!
```

This fix moves the description in the pyproject.toml from multiple lines to one line. The summary in the PKGINFO appears to be taken from the pyproject.toml description. 

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,11 +1,6 @@
 [project]
 name = 'drizzlepac'
-description = """
-    HST image combination using the drizzle algorithm to
-    combine astronomical images, to model image distortion,
-    to remove cosmic rays, and
-    generally to improve the fidelity of data in the final image.
-    """
+description = """ HST image combination using the drizzle algorithm to combine astronomical images, to model image distortion, to remove cosmic rays, and generally to improve the fidelity of data in the final image. """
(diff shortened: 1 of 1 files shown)


mdlpstsci (approved):
I draw the same conclusion you have in that the description text just needed to be one line with no white space.  I tried searching for "SetuptoolsDeprecationWarning: Invalid config." and while I found some complaints, no one addressed the issue directly.  Users were told to download some other file specific to the package and the issue was fixed.  I wish there were a way to test this first.  In any case, we will find out when I try making the build or doing a "dry run".  As it is a warning, we still have to fix the issue if necessary.

PR #1737, opened 2024-02-15
pyproject.toml specify python_requires >=3.10

s-goldman:
This change does not affect the default build python version, but does specify a requirement to anyone using the code.


mdlpstsci (approved):
I recall you and Nadia agreed upon the low limit of Python 3.10.

PR #1762, opened 2024-03-07
HLA-1195 Improved S_REGION Around Edges

s-goldman:
Resolves [HLA-1195](https://jira.stsci.edu/browse/HLA-1195)

This PR improves the calculation of S_REGION's in cases where too many vertices are created near noisy edges. The change uses a several  techniques with the basic idea of accentuating and then diminishing the features along the edges of the image.

1.  It starts will binary_fill_holes (as before) to fill gaps in the image. 
2. Then it uses binary_dialation which expands the overall region and curves/blurs the edges slightly.
3. Then we use binary_erosion  in a kind of reverse_dilation to get our original, yet more geometric, region.
4. We then use a np.bitwise_or to ensure that we return the hard corners that may have been curved through the process.
5. The edges are then gaussian smoothed to remove additional bad edge pixels.  

**Checklist for maintainers**
- [x] added relevant label(s)

[Jenkins test](https://plwishmaster.stsci.edu:8081/job/RT/job/Drizzlepac-Developers-Pull-Requests/61/)


mdlpstsci (approved):
The code changes look fine.  You and Warren agree on the updates needed.  Please note I re-ran the failed PyTest as there was some connection problem, and now the test succeeds.

drizzlepac/haputils/cell_utils.py
```diff
@@ -547,7 +548,12 @@ def find_corners(self, member='total'):
             fp = np.clip(self.footprint, 0, 1).astype(np.int16)
 
             # simple trick to remove noise and small regions 3x3 or less.
-            scmask = ndimage.binary_dilation(ndimage.binary_erosion(fp, iterations=3), iterations=2)
+            scmask_dilated_erroded = ndimage.binary_dilation(ndimage.binary_erosion(fp, iterations=3), iterations=2)
+            
+            # Start by smoothing out the edges of the chips/field
+            # this will remove rough edges up to 3 pixels deep along the image edge
+            scmask = ndimage.gaussian_filter(scmask_dilated_erroded.astype(np.float32) * 100, sigma=2) > 50
```
mdlpstsci, line 555: Please add a much needed comment here as to the reason for multiplying by 100 (which leads to the "> 50" criterion).  I seem to recall this is scaling needed to give the actual values some presence over the weeds, but I may not recall correctly.  Hence, the need for a comment.

mdlpstsci (changes requested):
Please add a much needed comment.

drizzlepac/haputils/cell_utils.py
s-goldman: Ahhh right. We could have removed it and used a threshold of 0.5, but warren wanted to keep it from being a decimal I think. I'll add a comment.

mdlpstsci (approved):

PR #1766, opened 2024-03-08
Fix failing Jenkins tests

s-goldman:
Tests are currently failing because we specified the required python version in the [pyproject.toml](https://github.com/s-goldman/drizzlepac/blob/18a8df11e0d8dea1a36b62325a99ac2154f0d02c/pyproject.toml#L5) as python 3.9, but we are still testing python 3.9 in Jenkins. This PR should remove python 3.9 testing.

PR #1767, opened 2024-03-12
HLA:1061 Re-enable final_fillval set to fill in values when WHT=0

### LINKED ISSUE #1613: AstroDrizzle - Investigate new option for final_fillval when WHT=0
_Issue [HLA-1061](https://jira.stsci.edu/browse/HLA-1061) was created on JIRA by [Jennifer Mack](https://jira.stsci.edu/secure/ViewProfile.jspa?name=mack):_

[Readthedocs|[https://drizzlepac.readthedocs.io/en/latest/astrodrizzle.html]] states two options for the parameter : **final_fillval**   float (Default = None)

> The value for this parameter is to be assigned to the output pixels that have zero weight or which do not receive flux from any input pixels during drizzling.
> 
> If the default of *`None`* is used, and if the weight in both the input and output images for a given pixel are zero, then the *output pixel will be set to the value it would have had if the input had a non-zero weight.* 
> 
> Otherwise, if a **numerical value** is provided (e.g. 0), then these pixels will be set to that numerical value.

Actual behavior: If the default of **None** is used, the fill value is set to {*}NaN{*}.  This is ok for some datasets, but not the desired behavior in the pipeline, e.g. especially for datasets with no dither (WFPC2) or for drizzling only a single input frame.

Currently the MDRIZTAB parameters for ACS and WFC3 use a workaround by setting all bad pixels to good, e.g. final_bits=65535 for N=1.  
 * PRO:  Fills in bad pixels in the SCI array,
 * CON: Does not map the bad pixels into the WHT image.

Desired behavior:   Add a new option to fill in the SCI array pixel value with either: a best guess of what it would have been <sub>OR</sub> the value from the last image in the input list.  

Users can then use the WHT image as a mask for the SCI array prior to doing photometry, if desired.

s-goldman:
Resolves [HLA-1061](https://jira.stsci.edu/browse/HLA-1061)

Closes #1613 

This PR reverts a previous change: PR #1222. We are re-enabling the behavior that pixels with WHT=0 are still filled in with any available value. This includes bad pixel and fuzzy edge areas. The code has been tested with ipppssoot u65q06 which successfully fills in WHT=0 areas and still produces the desired S_REGION. 

Work still to do is allowing the options of a fillvalue of nan when desired by users. 

**Checklist for maintainers**
- [x] added relevant label(s)

[Jenkins test ](https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FDrizzlepac-Developers-Pull-Requests/detail/Drizzlepac-Developers-Pull-Requests/60/pipeline)(failing: need to update truth files)


s-goldman:
A Note: The Jenkins test PR (#1766) needs to be merged before this one.

tests/drizzle/test_kernel.py
```diff
@@ -140,12 +140,12 @@ def test_zero_input_weight(kernel, kernel_pars):
...
+    # adding two additional bright "sources"
     kernel_pars.insci[6, 7] = 1000
     kernel_pars.insci[9, 6] = 1000
 
     # resample
     cdriz_setup.cdriz_call(kernel_pars, kernel)
 
-    # check that any pixel with 0 weight has any counts:
-    assert np.sum(kernel_pars.outsci) < 1e5
+    # check that orginal bad pixel flux still coming through:
```
mdlpstsci, line 150: Fix spelling: orginal -> original

mdlpstsci (changes requested):
Just a small typo to fix.  I added "Re-enable" to the title of this issue as this ticket is not creating new functionality, but re-enabling previous functionality.  Also, this PR does not seem (to me) to resolve fully HLA-1061 (or am I misunderstanding)?  It seems that ticket has a portion to add a new option.  Please let me know where I am misunderstanding.

mdlpstsci (approved):

mcara (approved):

PR #1771, opened 2024-03-14
SVM: Reconcile removal of bad segments from segmentation image with removal of corresponding rows from the "total" or white light source catalog

mdlpstsci:
To remove bad rows from the "total" or white light source catalog, as well
as to remove the corresponding segments from the segmentation image, ensure the rows are first eliminated from the catalog before the segments are
removed from the segmentation image and reordered.   This dataset
seems to reveal an awareness by the catalog as the row lists, both
good and bad, based upon the original segment ordering cause a fatal
error if the segmentation image is trimmed/re-ordered before the catalog
rows are removed.

Resolves [HLA-1238](https://jira.stsci.edu/browse/HLA-1238)

This PR addresses ...

**Checklist for maintainers**
- [x] added relevant label(s)

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1,6 +1,5 @@
 """This script contains code to support creation of photometric sourcelists using two techniques:
 aperture photometry and segmentation-map based photometry."""
-
 import copy
 import math
 import sys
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -1874,20 +1873,13 @@ def identify_sources(self, **pars):
             # the segmentation image too
             good_rows = []
             good_segm_rows_by_label = []
-            bad_segm_rows_by_label = []
             updated_table = None
             for i, old_row in enumerate(total_measurements_table):
```
s-goldman, line 1877: There is nothing wrong with this, but there is probably a more succinct way to write this for loop now that there is no else statement.

mdlpstsci: Only the minimum necessary changes were done here as it is not understood why any modification was needed here at all.

raswaters (approved):

PR #1772, opened 2024-03-14
Fix jenkins pytests

s-goldman:
This PR reverts a small change that was made to the jenkinsRT in #1766 when I was trying to fix the test. This should now work as intended. 

[Jenkins test](https://plwishmaster.stsci.edu:8081/job/RT/job/Drizzlepac-Developers-Pull-Requests/66/)

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -38,12 +38,12 @@ bc2.conda_packages = ['python=3.10']
 bc2.build_cmds = ["pip install numpy astropy codecov pytest-cov ci-watson || true",
                  "pip install -r requirements-dev.txt --upgrade -e '.[test]' || true",
                  "pip freeze || true"]
-bc2.test_cmds = ["pytest --cov=./ --basetemp=tests_output --junitxml=results.xml --bigdata || true",
+bc1.test_cmds = ["pytest --cov=./ --basetemp=tests_output --junitxml=results.xml --bigdata || true",
                 "codecov || true"]
-bc2.test_configs = [data_config]
+bc1.test_configs = [data_config]
 // Apply a large failure threshold to prevent marking the pipeline job failed
 // when xunit ingests any test results
-bc2.failedFailureThresh = 1000
+bc1.failedFailureThresh = 1000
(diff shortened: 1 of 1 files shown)


JenkinsfileRT
```diff
@@ -38,12 +38,12 @@ bc2.conda_packages = ['python=3.10']
...
                  "pip freeze || true"]
-bc2.test_cmds = ["pytest --cov=./ --basetemp=tests_output --junitxml=results.xml --bigdata || true",
+bc1.test_cmds = ["pytest --cov=./ --basetemp=tests_output --junitxml=results.xml --bigdata || true",
                 "codecov || true"]
-bc2.test_configs = [data_config]
+bc1.test_configs = [data_config]
 // Apply a large failure threshold to prevent marking the pipeline job failed
 // when xunit ingests any test results
-bc2.failedFailureThresh = 1000
+bc1.failedFailureThresh = 1000
```
mdlpstsci, line 46: I readily admit to not understanding the JenkinsfileRT and depend upon the build managers for these files.  Having said this, I do not understand why the **bc1** objects are (lines 41, 43, and 46): (1) in the **bc2** section, (2) one duplicate line from the **bc1** section, (3) why these are not **bc2** commands?

mdlpstsci (dismissed):
Questions regarding these changes.

JenkinsfileRT
s-goldman: I thought the same thing, which is why I to changed them in the previous PR to see if that was the issue. It turns out that the out of order lines were intended, and this PR is reverting those changes.

zacharyburnett: I think @mdlpstsci is right and these are duplicates lines overwriting the values assigned above. If this is the only way the tests run, then these lines should be moved upward to replace the analogous ones that assign values to `bc1`.

Alternatively, if we want both `bc1` and `bc2` to run tests, then just copy these lines to replace the `bc1` ones above, and keep these affecting `bc2`

nden (approved):

PR #1784, opened 2024-04-01
HLA-1222: Remove the restriction on the version of stsci.imagestats

mdlpstsci:
Resolves [HLA-1222](https://jira.stsci.edu/browse/HLA-1222)

Remove the restriction on the stsci.imagestats package version.  This is specifically to help with the generation of a release candidate as a build sequence with is already in-play.  We need to use an older version of stsci.imagestats (1.6.3) at this time, and the previous release candidate was manually installed to force the use of stsci.imagestats 1.6.3.  This means the next release candidate is automatically restricted to use the same version of the package (which is what we want).

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     'matplotlib',
     'stsci.tools>=4.0',
     'stsci.image>=2.3.4',
-    'stsci.imagestats>=1.8.1',
+    'stsci.imagestats',
     'stsci.skypac>=1.0.9',
     'stsci.stimage',
     'stwcs>=1.5.3',


pyproject.toml
```diff
@@ -23,7 +23,7 @@ dependencies = [
     'matplotlib',
     'stsci.tools>=4.0',
     'stsci.image>=2.3.4',
-    'stsci.imagestats>=1.8.1',
+    'stsci.imagestats',
```
raswaters, line 26: Do we need a max on the version, to avoid the version that caused the earlier problems?

mdlpstsci: No.  Because the current (RC5) has stsci.imagestats set to 1.6.3, this restricts the version of the package that RC6 is allowed to use.

raswaters (approved):

PR #1786, opened 2024-04-11
HLA-1216 Remove tophat drizzle kernel and add warning for gaussian and lanczos3

### LINKED ISSUE #1751: Remove Drizzle Kernels: gaussian, lanczos, and tophat
_Issue [HLA-1216](https://jira.stsci.edu/browse/HLA-1216) was created on JIRA by [Steve Goldman](https://jira.stsci.edu/secure/ViewProfile.jspa?name=sgoldman):_

It was found that several kernels that are available options in drizzlepac do not conserve flux and should be removed as options. This change is related to recent changes in the standalone [drizzle package](https://github.com/spacetelescope/drizzle/pull/141) and Jira ticket AL-774.

 

The kernels that should be removed are gaussian, lanczos, and tophat. What should remain is point, square, and turbo. 

 

The documentation as well as drizzle unit test also need to be updated.

s-goldman:
Resolves [HLA-1216](https://jira.stsci.edu/browse/HLA-1216)

Closes #1751 

This PR addresses the need to remove the "tophat" drizzling kernel that was found to be ineffective. When tophat is selected as a kernel for astrodrizzle, it now warns the user that it is no longer an option, and that it is switching to "square". 

Warnings are also now included in the running log for the "gaussian" and "lanczos3" kernels when they are used to tell the user that the kernels may not be conserving flux. 

These were added to processInput.py in order to avoid having the warning duplicated for every input image. 

Both the removal of the "tophat" kernel and the warnings for the other two kernels have also been documented in the readthedocs. 

Also removed milestone checkmark in PR template, which we don't use. 

**Checklist for maintainers**
- [x] added relevant label(s)

[jenkins test](https://plwishmaster.stsci.edu:8081/blue/organizations/jenkins/RT%2FDrizzlepac-Developers-Pull-Requests/detail/Drizzlepac-Developers-Pull-Requests/71/pipeline/246/)


CHANGELOG.rst
```diff
@@ -18,21 +18,27 @@ number of the code change for that issue.  These PRs can be viewed at:
...
-  source catalog before the corresponding bad segments are removed from the
-  segmentation image. [#1771]
+- Removed "tophat" as a kernel option, added warnings for "gaussian" and "lanczos3" 
+  that they may not be conserving flux. [#1786]
 
 - Bug fix for mdriztab=True option in Astrodrizzle previously overwriting user inputs. [#1774]
 
 - Reverted PR #1222 allowing pixels to be filled with available data where WHT=0. [#1767]
 
+- Force the identified bad rows to be removed from the total (aka white light)
```
mdlpstsci, line 31: FYI - We can fix up the CHANGELOG on the main branch after Drizzlepac 3.7.0 is released.  The PR#1771 was part of 3.7.0 as the change was specifically requested.

doc/source/mast_data_products/hap-parameters.rst
```diff
@@ -153,7 +153,7 @@ driz_sep_bits: int (*default="16"*)
     Integer sum of all the DQ bit values from the input image's DQ array that should be considered 'good' when building the weighting mask. This can also be used to reset pixels to good if they had been flagged as cosmic rays during a previous run of ``AstroDrizzle``, by adding the value 4096 for ACS and WFPC2 data. Please see the section on Selecting the ``Bits`` Parameter for a more detailed discussion.
 
 driz_sep_kernel: str (*default="square"*)
```
mdlpstsci, line 155: The description for _driz_sep_kernel_ needs to be changed to make the "default" be "turbo" to be consistent with **adrizzle.help** and **astrodrizzle.help**.

drizzlepac/astrodrizzle.help
```diff
@@ -776,7 +781,7 @@ final_wht_type : {'EXP', 'ERR', 'IVM'} (Default = 'EXP')
           computed dark value from the image header to automatically generate
           an ``IVM`` file specific to each exposure.
 
-final_kernel : {'square', 'point', 'gaussian', 'turbo', 'tophat', 'lanczos3'} (Default = 'square')
+final_kernel : {'square', 'point', 'turbo', 'gaussian', 'lanczos3'} (Default = 'square')
```
mdlpstsci, line 784: Default value here needs to be changed to 'turbo'.

drizzlepac/processInput.py
```diff
@@ -131,6 +131,23 @@ def setCommonInput(configObj, createOutwcs=True, overwrite_dict={}):
...
         cfgpars.mergeConfigObj(configObj, mdriztab_dict)
     
+    # add warnings for kernels that may not work as intended
+    kernels_w_warnings = ["gaussian", "lanczos3", "tophat"]
+    for kernel_param in overwrite_dict:
+        if overwrite_dict[kernel_param] in kernels_w_warnings:
+            if overwrite_dict[kernel_param] == "tophat":
+                log.warning(
+                    f'WARNING: It was found that kernel "{overwrite_dict[kernel_param]}" does not work as intended. '
+                    + f'It has been removed as a option. Using kernel "square" instead for {kernel_param}.'
```
mdlpstsci, line 141: Do you want kernel 'square' to be promoted here, or should it be kernel 'turbo'?  It seems that 'turbo' is now the default.

mdlpstsci (changes requested):
See comments in code for changes.

ALSO, have adrizzle.py, astrodrizzle.py, and any other code been updated with regard to the default kernel?  With a quick check, I see adrizzle.py still has 'square' as the default.

drizzlepac/astrodrizzle.help
```diff
@@ -449,6 +445,15 @@ driz_sep_kernel : str {'square', 'point', 'gaussian', 'turbo', 'tophat', 'lanczo
         It leads to less resolution loss than other kernels, and typically
         results in reduced correlated noise in outputs.
 
+      * ``'tophat'``: (former option) this kernel is a circular "top hat" shape of 
+        width ``pixfrac``. It effects only output pixels within a radius of
+        ``pixfrac/2`` from the output position. This kernel has been removed
+        as it was found to be ineffective at reproducing results. 
+
+        .. warning:: While the ``'gaussian'`` and ``'lanczos3'`` kernels may
+        produce reasonable results, we cannot guarantee that they will properly
```
nden, line 454:
```suggestion
        produce reasonable results and can be useful in certain cases, they do not 
```

drizzlepac/processInput.py
```diff
@@ -131,6 +131,23 @@ def setCommonInput(configObj, createOutwcs=True, overwrite_dict={}):
...
+                )
+                overwrite_dict[kernel_param] = "square"
+            else:
+                log.warning(
+                    f'WARNING: We cannot guarantee that kernel "{overwrite_dict[kernel_param]}" is properly conserving flux.'
```
nden, line 146:
```suggestion
                    f'WARNING: Kernel "{overwrite_dict[kernel_param]}" does not conserve flux.'
```

nden (commented):
I added a few comments regarding the wording of the warning. Feel free to ignore them if there's a reason to be doubtful. AFAIK those kernels do not conserve flux so I think we should be straightforward with this.

mdlpstsci (dismissed):
Perhaps the "default" kernel issue is an inconsistency from past changes.  Please review default values for the kernel as in many places it now says "turbo" is the default, but "square" is referenced in other areas.  Please clarify.

drizzlepac/astrodrizzle.help
```diff
@@ -796,9 +801,10 @@ final_kernel : {'square', 'point', 'gaussian', 'turbo', 'tophat', 'lanczos3'} (D
           and is always aligned with the ``X`` and ``Y`` axes. This may result
           in a significant speed increase.
 
-        * ``'tophat'``: this kernel is a circular "top hat" shape of width
+        * ``'tophat'``: (no longer available) this kernel is a circular "top hat" shape of width
           ``pixfrac``. It effects only output pixels within a radius of
-          ``pixfrac/2`` from the output position.
+          ``pixfrac/2`` from the output position. This kernel has been removed
+            as it was found to produce poor results.
```
mcara, line 807: why even keep this as an option?

drizzlepac/processInput.py
mcara: IMO, it is better to let the code raise an exception instead of changing the kernel: some may not read through the log file but an exception is noticeable

s-goldman: I happy to use turbo. I don't think that this will be a common scenario, and if it does, a user may see that their parameter is being ignored and set the kernel to a viable option in a follow-up run of the code.

drizzlepac/astrodrizzle.help
```diff
@@ -449,6 +445,15 @@ driz_sep_kernel : str {'square', 'point', 'gaussian', 'turbo', 'tophat', 'lanczo
         It leads to less resolution loss than other kernels, and typically
         results in reduced correlated noise in outputs.
 
+      * ``'tophat'``: (former option) this kernel is a circular "top hat" shape of 
+        width ``pixfrac``. It effects only output pixels within a radius of
+        ``pixfrac/2`` from the output position. This kernel has been removed
+        as it does not conserve flux.
```
mcara, line 451: 1. I think this should be removed as an option altogether;
2. Other kernels have been found not to conserve flux but you've kept them (i.e., gaussian). The real reason it was removed it because it is buggy beyond repair (considering the effort top fix it and most likely very poor performance, time-wise, if fixed).

drizzlepac/processInput.py
mcara: The default for the final drizzle should be 'square' and for CR-rejection it was 'turbo' for performance reasons.

drizzlepac/astrodrizzle.help
s-goldman: That works for me.

drizzlepac/processInput.py
s-goldman: removed.

mcara (approved):
LGTM

s-goldman:
> LGTM

Readthedocs build failed due to indent, should be fixed now.

tests/acs/test_unit.py
```diff
@@ -219,6 +219,7 @@ def test_lanczos_with_grid(self):
         assert(med_diff < 1.0e-6)
         assert(max_diff < 1.0e-5)
 
+    pytest.mark.skip(reason='Point kernel no longer supported')
```
mdlpstsci, line 222: The text here should be "Tophat" instead of "Point".

mdlpstsci (changes requested):
Update text in test_unit.py.

tests/acs/test_unit.py
s-goldman: fixed.

mdlpstsci (approved):

PR #1797, opened 2024-05-13
HLA-1138: Exclude single filter images in the generation of the total detection image

mdlpstsci:
Resolves [HLA-1138(https://jira.stsci.edu/browse/HLA-1138)

Closes #

This PR addresses ...
Exclude single filter images in the generation of the total detection
image to minimize cosmic ray contamination, unless there are only single filter images in the visit.

**Checklist for maintainers**
- [ ] added relevant label(s)

### DIFF
--- a/drizzlepac/haputils/product.py
+++ b/drizzlepac/haputils/product.py
@@ -632,6 +632,7 @@ class TotalProduct(HAPProduct):
 
         .. note:: Cosmic-ray identification is NOT performed when creating the total detection image.
         """
+
         # This insures that keywords related to the footprint are generated for this
         # specific object to use in updating the output drizzle product.
         self.meta_wcs = meta_wcs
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/product.py
```diff
@@ -657,7 +658,41 @@ def wcs_drizzle_product(self, meta_wcs):
...
+        # dictionary to hold the filter name and the number of exposures per filter
+        set_of_filters = set([element.filters for element in self.edp_list])
+        count_dict = {key: [] for key in set_of_filters}
+
+        # Loop over the exposure objects to collect the exposures for each filter
+        for expo in self.edp_list:
+            for key in count_dict:
+                if key == expo.filters:
+                    count_dict[key].append(expo.full_filename)
+                    break
```
rlwastro, line 675:
```suggestion
        count_dict = {}

        # Loop over the exposure objects to collect the exposures for each filter
        for expo in self.edp_list:
            count_dict.set_default(expo.filters,[]).append(expo.full_filename))
```

I think the current code is much more complicated than it needs to be.  I removed the `set_of_filters` variable because I think it is not used elsewhere.  If I'm wrong about that, it can be computed from `count_dict` after the loop.

drizzlepac/haputils/product.py
```diff
@@ -657,7 +658,41 @@ def wcs_drizzle_product(self, meta_wcs):
...
+                edp_filenames += value
+            else:
+                exclude_string.append("Excluding single exposure filter image {} for {}.".format(value, key))
+
+        # However, if all the filters only have one exposure, just use all
+        # the exposures
+        if not edp_filenames:
+            for value in count_dict.values():
+                edp_filenames += value
+        else:
```
rlwastro, line 692: I think it will be useful to have a log entry here for this unusual case.

```suggestion
            log.info("All filters are single exposures, so all are included.")
        else:
```

rlwastro (changes requested):

drizzlepac/haputils/product.py
mdlpstsci: Good idea.

rlwastro (approved):
Looks good to me!

PR #1805, opened 2024-05-31
HLA-1257: Remove the use of a custom smoothing kernel for catalog generation

mdlpstsci:
Resolves [HLA-1257](https://jira.stsci.edu/browse/HLA-1257)

Closes #

This PR addresses ...
Remove the use of a "custom" kernel for the segmentation algorithm as it has been shown to cause smearing issues. The default kernel is now a Gaussian which is produced by calling an exisiting astrometric_utils.py function, build_gaussian_kernel.  All internal comments in the catalog_utils.py regarding the "custom" kernel have been updated. This is only Step 1 of this update.

**Checklist for maintainers**
- [X] added relevant label(s)

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -138,9 +138,12 @@ class CatalogImage:
 
         # Build a Gaussian2DKernel - the use of a custom kernel based upon a PSF
         # of the input image is no longer used here as it could be asymmetric and
-        # lead to smeared segments (i.e., centroid coordinates offset from actual source)
+        # lead to smeared segments (i.e., centroid coordinates offset from actual source).
+        # The value of 11 for the Gaussian kernel is big enough for all detectors. This
+        # value is hard-coded here in contrast to being in the configuration files like 
+        # the RickerWavelet2DKernel.
         self.kernel_fwhm = fwhmpsf / self.imgwcs.pscale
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -1656,6 +1652,11 @@ def identify_sources(self, **pars):
                 # Only pass along the array to be consistent with the g2d_kernel object
                 rw2d_kernel = rw2dk.array
 
+                # Write out diagnostic data
+                if self.diagnostic_mode:
+                    outname = self.imgname.replace(".fits", "_rwkernel.fits")
+                    fits.PrimaryHDU(data=rw2d_kernel).writeto(outname)
+
```
rlwastro, line 1659: I wondered whether the diagnostic_mode should also write out the Gaussian kernel to a fits file?  Maybe that already happens and I have not encountered it.  But it seems odd to write out only the RW kernel.

rlwastro (approved):
This all looks like a simple change, and I did not see any issues.  I had one question about the diagnostic mode but that can be ignored if there is an answer.  It also looked like all the documentation and comment changes were good.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: @rlwastro 
The Gaussian kernel is written out to a file in diagnostic mode (has been for a long time):
hst*_kernel.fits vs hst*_rwkernel.fits.

PR #1806, opened 2024-05-31
[SCSB-155] build with Numpy 2.0 release candidate

zacharyburnett:
resolves [SCSB-155](https://jira.stsci.edu/browse/SCSB-155)

Numpy 2.0 is scheduled for release on June 19th. Due to the C ABI updates, Python projects that build C extensions are required to build against Numpy 2.0 in order to use Numpy 2.0 in runtime.
This is backwards-compatible with Numpy versions in runtime as far back as `1.26`

this PR pins `numpy>=2.0.0rc2` in `build-system.requires`
> [!NOTE]
> this PR was generated automatically by ``batchpr`` :robot:

**Checklist for maintainers**
- [ ] added relevant label(s)


pyproject.toml
```diff
@@ -1,89 +1,107 @@
...
-    'tables',
-    'typing_extensions>4.0',
-    'markupsafe<=2.0.1',
+    "astropy>=5.0.4",
+    "fitsblender>=0.4.2",
+    "scipy",
+    "matplotlib",
+    "stsci.tools>=4.0",
+    "stsci.image>=2.3.7",
+    "stsci.imagestats",
```
mdlpstsci, line 33: We want to make sure drizzlepac uses stsci.imagestats >= 1.8.2 (**_NOT_** 1.7.0, 1.8.0, or 1.81. though I see these are currently "yanked").

pyproject.toml
```diff
@@ -1,89 +1,107 @@
...
+    "stsci.skypac>=1.0.9",
+    "stsci.stimage",
+    "stwcs>=1.5.3",
+    "tweakwcs>=0.8.2",
```
mdlpstsci, line 37: We want to make sure drizzlepac uses tweakwcs >= 0.8.7 as there is a bug in 0.8.6.

pyproject.toml
```diff
@@ -95,90 +113,88 @@ zip-safe = false
...
-    'F521', # '...'.format(...) has invalid format string: Single '}' encountered in format string
-    'F523', # '...'.format(...) has unused arguments at position(s): 0, 1
-    'F541', # f-string without any placeholders
-    'F632', # Use `==` and `!=` to compare constant literals
-    'F633', # Use of `>>` is invalid with `print` function
-    'F811', # Redefinition of unused
-    'F821', # Undefined name
-    'F822', # Undefined name `getHelpAsString` in `__all__`
-    'F841', # Local variable is assigned to but never used
+    "E401",
```
mdlpstsci, line 179: Just a question here.  Is it against the new rules to have the commentary associated with the error codes to ignore.  The commentary is rather helpful.  Thanks.

mdlpstsci (changes requested):
Some changes need to be made to ensure drizzlepac is using "good" versions of the stsci.imagestats and tweakwcs packages.  I also had one question.

pyproject.toml
zacharyburnett: oh, sorry! That must have been the formatter; I'll add those comments back in

mdlpstsci (approved):
Thanks for addressing these issues. You could not have known about the imagestats and tweakwcs versions!

PR #1808, opened 2024-06-06
Add missing 3.11 test command

jhunkeler:
Fixes:

* HCALDP pipeline job not testing anything under Python 3.11
* Regression tests not running under Python 3.11


mdlpstsci (changes requested):
Thank you for adding the tests to be run under Python 3.11.  

However, I do have some questions regarding this file where the build configs all migrate up a version of Python.  By this I mean:
Should the Python 3.10 builds just become Python 3.11 builds?
Should there be a section added to run the Python 3.11 build under MacOS?
Should we add a section for Python 3.12?

mdlpstsci (approved):
Fantastic!  This version of the JenkinsfileRT is a model template.

s-goldman:
I previously saw that the `bc1.failedFailureThresh = 1000` line is duplicated at lines 37 and 52. I thought that it may have been a mistake, and the second mention should instead be a command for bc2. The second `bc1.failedFailureThresh = 1000` command needs to be there or the tests fail. 

You may have fixed that with these improvements, but I just wanted to make sure that someone else doesn't make my mistake.

mdlpstsci:
The Linux and MacOS tests are now both passing.  The problem which I recall causing discrepancies between the Linux and MacOS results resided in the source catalogs, specifically the number of found sources.  Currently, those particular tests are skipped which doing some basic tests on the catalogs.  When explicitly and manually testing on Linux and MacOS, I got the same results.  The problem that I could not see what was happening on the test/build machines to figure out the issue.  Since I am currently modifying the catalogs a great deal, I am expecting the new results to be very different from the current assert values.  I will be generating new values after I am done with changes, as well as running tests under Linux and MacOS.

mdlpstsci:
@s-goldman @jhunkeler 
I have been looking at the output failures and found some quick and some not so quick issues to address.  The quick fixes are:

1. Set the **crrefer** environment variable (Joe has done that in this ticket)
2. Recover erroneously deleted poller files needed for the tests (Michele has done this in Git PR#1809).
3. Tests test_svm_canary.py and test_svm_hrcsbc.py are effectively the same.  One of the tests should be deleted (test_svm_canary.py). (Michele also did this in Git PR#1809).

It should be noted that once the tests which failed due to the missing **crrefer** variable execute, the output products could show differences.

Not so quick issues:
**_As @s-goldman noted correctly earlier_** there are tests (image tests) which pass on Linux and DO NOT pass on MacOS.  These needs to be investigated in a different ticket.

1. tests/acs/test_tweak.py and tests/acs/test_unit.py fail due to results exceeding difference tolerances or some results are just very different.
2. tests/hap/test_svm_hrcsbc.py no longer achieves a FIT_SVM_GAIAeDR3 WCS solution.

mdlpstsci:
@jhunkeler Do we tempt fate by pushing this ticket?  Just curious.  Also, the drizzlepac docs seem to be failing on most PRs due to a bad path (see below). Will these changes accommodate this issues?

CondaVerificationError: The package for sphinx located at /home/docs/.asdf/installs/python/mambaforge-4.10.3-10/pkgs/sphinx-7.3.7-py311h5eee18b_0
appears to be corrupted. The path 'lib/python3.11/site-packages/sphinx/util/rst.py'
specified in the package manifest cannot be found.

s-goldman:
@jhunkeler Are these changes still relevant, or can this PR be closed?

PR #1811, opened 2024-06-10
Streamline some regression tests

mdlpstsci:
Resolves [HLA-1265](https://jira.stsci.edu/browse/HLA-1265)

Closes #

This PR addresses ...
This PR streamlines some regression tests which were designed to work on large numbers of datasets to generate statistics in order to prove the HAP project should proceed.   These tests are kept for historic reasons and only need to be kept up-to-date so they can execute as necessary.

Removed template_svm_demo.py as it was really a template which was superseded by the real test file, test_svm_je281u.py.  Changed the default number of datasets to process for the "large" statistical tests to "1" (down from "5") as the test are taking a long time to run.  Simplified how an ascii file containing filenames is read (test_run_svmpoller.py).

**Checklist for maintainers**
- [ ] added relevant label(s)


jhunkeler (approved):
The xdist worker issue persists on the runner, but still, the change from Table to plain text makes things easier to understand.

PR #1816, opened 2024-06-13
Fixes codecov

jhunkeler:
* Fixes missing python 3.11 test commands

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -7,6 +7,8 @@ if (env.ARTIFACTORY_ENV) {
     artifactory_env = env.ARTIFACTORY_ENV
 }
 
+codecov_install = "curl -Os https://uploader.codecov.io/latest/linux/codecov && chmod +x codecov"
+
 // Allow modification of the job configuration, affects all relevant
 // build configs.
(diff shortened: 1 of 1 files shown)


JenkinsfileRT
```diff
@@ -60,6 +64,9 @@ bc3.conda_packages = ['python=3.11']
 bc3.build_cmds = ["pip install numpy astropy ci-watson",
 		 "pip install --upgrade -e '.[test]'",
                  "pip freeze"]
+bc2.test_cmds = ["pytest --env=${artifactory_env} --cov=./ --basetemp=tests_output --junitxml=results.xml --bigdata",
```
mdlpstsci, line 67: Should this be "b3" instead of "b2"?

mdlpstsci (changes requested):
Chastised for not leaving a comment.  I believe the "b2" should be "b3" on or around line 67.

mdlpstsci (approved):

s-goldman:
It looks like we also need to `pip install pytest-cov` in order for the --cov command to work for the 3.11 test.

PR #1832, opened 2024-07-01
HLA-1283: Added drizzlepac animation to the docs

s-goldman:
Resolves [HLA-1283](https://jira.stsci.edu/browse/HLA-1283)

This PR adds an animation as an animated gif to the astrometry section of the readthedocs.


mdlpstsci (approved):
Since this is needed ASAP, I am approving this PR.  We can then determine if there is a more "official" place or methodology for providing videos and such that are not tied to a user account.

mdlpstsci (approved):
Excellent.  A picture (or animation) is worth a thousand words, and this animation succinctly illustrates the alignment procedure.

PR #1837, opened 2024-07-08
Renaming archival tests so that they are no longer run daily

s-goldman:
We are renaming three tests (test_alignpipe_randomlist.py, test_randomlist.py, and test_run_svmpoller) so that they are no longer run every day. 

The tests were never meant to be nightly tests but rather a check of the viability of the HAP program early on using a randomized program from a list (ACSWFC3ListDefault50.csv). 

We are renaming them as opposed to moving or deleting them. This way, they can still be easily run by specifying their names explicitly (e.g. pytest --bigdata --slow tests/hap/archival_test_run_svmpoller) These files still require several associated files that will not move (conftest.py, ACSWFC3ListDefault50.csv).


mdlpstsci (approved):
Steve has already provided the explanation for the renaming of the tests.  These tests are retained mostly for historic reasons.

PR #1840, opened 2024-07-09
Upgrade jenkins dev test to python 3.11, fix ci-watson upgrade failures

s-goldman:


mdlpstsci (approved):

mdlpstsci (commented):
@s-goldman I do not understand why you want to reintroduce these tests.  Can you provide a URL where I can see an error or message that makes you believe the rename of these tests is causing a problem?

s-goldman:
> @s-goldman I do not understand why you want to reintroduce these tests. Can you provide a URL where I can see an error or message that makes you believe the rename of these tests is causing a problem?

Hey Michele, sorry I wasn't planning on merging this. I was just testing to see if the jenkins tests would pass if I reverted those changes. I don't have a reason why they should effect the test, but I also don't have a reason why the tests should have started failing.

mdlpstsci (approved):

PR #1864, opened 2024-08-12
Eliminate Python 3.13 from the target list

mdlpstsci:
Resolves [HLA-1307](https://jira.stsci.edu/browse/HLA-1307)

Since there are very few wheel packages on pypi right now for
Python 3.13, we must do a workaround and exclude 3.13 from the target list.  This is advice from Joe H. with an example from JWST.  This is necessary to do to push out the final Drizzlepac v3.7.1.

**Checklist for maintainers**
- [X] added relevant label(s)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,12 +12,13 @@ jobs:
     with:
       upload_to_pypi: ${{ (github.event_name == 'release') && (github.event.action == 'released') }}
       targets: |
-        # Linux wheels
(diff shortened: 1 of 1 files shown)


zacharyburnett (approved):
Should we also pin Python<3.13 in `pyproject.toml`? That would keep people on 3.13 from installing with pip from source

mdlpstsci:
@zacharyburnett It looks like you did this in PR#1861 which I approved and merged for you yesterday.  Is this what you meant?

PR #1872, opened 2024-08-22
HLA-1315: Fix "finally" block handling Grism/Prism under PyTest

mdlpstsci:
Resolves [HLA-1315](https://jira.stsci.edu/browse/HLA-1315)

Closes #

Fixed the code handling log/trailer file creation for Grism/Prism data in the "finally" block of run_hap_processing, as an error under PyTest 
was being generated as the designated log file could not be found to be copied and used as the Grism/Prism log file. The Grism/Prism log files are just placeholders, so an empty file is created if the designated log file cannot be found.  Added comments.  The code executes properly when not run under PyTest.

I manually ran test_svm_j97e06.py (test which revealed the problem) and test_svm_je281u.py in my workspace to test with Pytest.  I also ran both visits without PyTest.

**Checklist for maintainers**
- [X] updated or added relevant tests
- [X] added relevant label(s)

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -751,17 +751,28 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
         # The Grism/Prism SVM FLT/FLC images which have had their WCS reconciled with the
         # corresponding direct images need trailer files.  This is also true of the Ramp images
         # which have only been processed through the "align to Gaia" stage.  At this time, just
-        # copy the total trailer file, and rename it appropriately.
+        # copy the full SVM processing log and rename it appropriately to the Grism or Ramp
+        # trailer filename.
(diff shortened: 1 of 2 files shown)


zacharyburnett:
started a regtests job here: https://github.com/spacetelescope/RegressionTests/actions/runs/10514121938

zacharyburnett:
the tests failed, but the log is gigantic because of #1873

mdlpstsci:
Actually, only one test failed.  The **test_pixsky1** from the **test_acs_tweak.py** module failed.  This test has nothing to do with the fix I made as discussed in this PR as I changed code for the HAP/SVM processing on close out.  Did you see this failure previously?  In any case, I would prefer it to be a different ticket/PR as I need to finish testing changes for a ticket I am working for this sprint.

s-goldman (approved):
lgtm.

PR #1884, opened 2024-09-23
HLA-1323: s_region fix using simplify-polyline

s-goldman:
Resolves [HLA-1323](https://jira.stsci.edu/browse/HLA-1323)

This PR fixes the s_region issues that we have been observing in the SVMs. It uses the simplify-polyline package instead of skimage.corner_harris. Results from initial testing are shown in the Jira ticket. 

These code changes were generously contributed by @rlwastro.

**Checklist for maintainers**
- [x] added relevant label(s)
      
[Jenkins test](https://plwishmaster.stsci.edu:8081/job/RT/job/Drizzlepac-Developers-Pull-Requests/134/)

### DIFF
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -579,7 +579,7 @@ class SkyFootprint(object):
                 # Perform corner detection on each region/chip separately.
                 mask_corners = corner_peaks(corner_harris(label_mask),
                                        min_distance=3,
-                                       threshold_rel=0.2)
+                                       threshold_rel=0.1)
                 xy_corners = mask_corners * 0.
                 xy_corners[:, 0] = mask_corners[:, 1]
                 xy_corners[:, 1] = mask_corners[:, 0]
--- a/drizzlepac/haputils/cell_utils.py
+++ b/drizzlepac/haputils/cell_utils.py
@@ -18,6 +18,8 @@ from spherical_geometry.polygon import SphericalPolygon
 
 from PIL import Image, ImageDraw
 
+from simplify_polyline import simplify
+
 from stwcs.wcsutil import HSTWCS
 from stsci.tools import logutil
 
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/cell_utils.py
```diff
@@ -540,70 +539,53 @@ def find_corners(self, member='total'):
...
-                    ordered_indices.append(np.where(distarr == minval)[0][0])
-                radial_order = np.argsort(ordered_indices)
-                ordered_xyc = xy_corners[radial_order].tolist()
-                ordered_xyc.append(ordered_xyc[0])  # close polygon
+                # Correct for the label_border padding
+                edge_pixels = trace_polygon(label_mask, mask_slice) - label_border
+
+                # simplify the polygon
+                # note edge_pixels and xy_corners polygons are already closed
+                xy_corners = simplify(edge_pixels, min_dist=3.0)
```
rlwastro, line 580:
```suggestion
                xy_corners = simplify(edge_pixels, min_dist=10.0)
```
Later testing shows that `min_dist=10.0` is a better simplification threshold.  It reduces the complexity of the footprints a bit.

drizzlepac/haputils/cell_utils.py
```diff
@@ -540,70 +539,53 @@ def find_corners(self, member='total'):
...
+
+                if xy_corners.shape[0] <= 3:
```
rlwastro, line 582: This is certainly an error (as the comment says).  We might never encounter it with the new code though.  At a minimum, this message ought to be included in the trailer file (I don't know whether a simple print does that or not).  Anyway, some kind of minor tweak here is probably needed.

rlwastro (dismissed):
Changing the simplify threshold to 10 is required.  Otherwise everything looks good to me.

drizzlepac/haputils/cell_utils.py
s-goldman: I used a log.info() instead of a print() which should add it to any trailer/log file.

s-goldman:
While I will continue to test programs that were previously troublesome, this change appears to have fixed the issue.

PR #1886, opened 2024-09-25
Small change to Rick's Quick Fix

s-goldman:
The code is currently shrinking the boundary by one pixel. This code change will avoid that issue. The iterations for dilating and eroding the edges will now be the same. 

Several small changes were also contributed by @rlwastro.


rlwastro (approved):
All looks good to me

PR #1888, opened 2024-09-26
Comments added to cell_utils s_region code. 

s-goldman:


drizzlepac/haputils/cell_utils.py
```diff
@@ -1354,6 +1397,7 @@ def _poly_trace(input_mask, box_size=3):
 
         indx = 0
         if len(pts[0]) > 1:
+            # if two pixels found, pixks the downward one.
```
rlwastro, line 1400:
```suggestion
            # if two pixels found, picks the one that appears to continue in
            # the counter-clockwise direction. 
```

rlwastro (approved):

PR #1894, opened 2024-10-10
HLA-1284/1287: Properly account for illuminated and non-illuminated pixels

mdlpstsci:
Resolves [HLA-1284](https://jira.stsci.edu/browse/HLA-1284)

Closes #

Use the filter mask (illuminated vs non-illuminated pixels) when processing the  Filter product.  For all products, properly account for the number of pixels which are illuminated,  the number of pixels that contain true zero in the illuminated portion of the image, and non-illuminated pixels.

**Checklist for maintainers**
- [X] added relevant label(s)
- [X] ran regression tests, post a link to the Jenkins job below.

https://plwishmaster.stsci.edu:8081/job/RT/job/Drizzlepac-Developers-Pull-Requests/146/
First run done with unchanged reference/truth files in /dev and results are in /drizzlepac-results/wfc3/11Oct2024T.  Created drizzlepac/3.7.2 which is a copy of drizzlepac/dev, but the different files (WFC3 and WFPC2) have NOT yet been copied to drizzlepac/3.7.2.  I want to process some data locally as it is curious there were no ACS failures.

Investigated the "failed" tests and they have nothing to do with this PR.  The code specifically changed in this PR have nothing to do with the WFC3 tests as the changes are for SVM processing.

Ran in my Mac (23.6.0 Darwin Kernel Version 23.6.0) under caldp_20240813 build with updates from Drizzlepac.  All the WFC3 and WFPC2 tests **_passed._**

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -155,6 +155,8 @@ class CatalogImage:
                            maxiters=3):
         """Use a sigma-clipped algorithm or Background2D to determine the background of the input image.
 
+        There is also a special case background assessment reserved for SBC (MAMA detector) images.
+
         Parameters
         ----------
         image : ndarray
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -248,38 +250,48 @@ def compute_background(self, box_size, win_size,
         bkg = None
         is_zero_background_defined = False
 
-        # Make a local copy of the data(image) being processed in order to reset any
-        # data values which equal nan (e.g., subarrays) to zero.
-        imgdata = np.nan_to_num(self.data, copy=True, nan=0.0)
```
s-goldman, line 253: Will altering this line result in nans in the image?

s-goldman (commented):
Just one lingering question, the rest looks good.

drizzlepac/haputils/catalog_utils.py
mdlpstsci: Took some time to double-check!  There will still be nans in the "image", but the masks (noting where data should or should not be used) are employed when performing important computations.

s-goldman (approved):

PR #1902, opened 2024-10-22
HLA-998: Proper handling of files which have SCI extensions completely zero in RAW

s-goldman:
Resolves [HLA-998](https://jira.stsci.edu/browse/HLA-998)

Closes #

This PR addresses the need to skip processing for exposures with science arrays of only zeros in ALL science extensions. One or multiple science extensions with all zeros will result in warnings for each extension.  

This PR does not use `no_proc_keys` or `no_proc_values`, but does set do_process=False. 

I also amended (and now use) the `util.count_sci_ext` function to return the indices of the science extensions if the `return_ind` flag is used.

I've also changed the variable related to the error of little-to-no exposure time (`is_zero` --> `exposure_time_near_zero`) in order to avoid confusion with our new error case. 

**Checklist for maintainers**
- [x] added relevant label(s)

[jenkins test ](https://plwishmaster.stsci.edu:8081/job/RT/job/Drizzlepac-Developers-Pull-Requests/152/)


drizzlepac/haputils/analyze.py
```diff
@@ -364,6 +364,8 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
     for i, input_file in enumerate(input_file_list):
         header_hdu = 0
         header_data = getheader(input_file, header_hdu)
+        science_hdu = 1
```
mdlpstsci, line 367: FLT or FLC images can have multiple SCI extensions (representing different chips) depending upon the detector and how the data was acquired.  You need to determine how many SCI extensions are in the file, and then check all of the extensions.  There would only be 1 or 2 SCI extensions. 

If for some strange observation with two SCI extensions, only one extension contains all zeros, the processing should continue.  A message should be issued indicating "Warning, (SCI, n) is all zeros, but processing will continue." or something like this.  Only if all the SCI extensions are zero  in the file should the processing be set to FALSE.

drizzlepac/haputils/analyze.py
```diff
@@ -533,6 +539,10 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
             # GYROS mode so some drift)
             generate_msg(input_file, msg_type, no_proc_key, no_proc_value)
         
+        elif science_array_of_zeros:
+            do_process=False
+            log.warning(f'Science data for {input_file} filled with zeros, not aligning.')
```
mdlpstsci, line 544: I would change the message slightly to be consistent with the current code.  log.warning(f'Science data for {input_file} filled with zeros.  Dataset cannot be aligned.')

mdlpstsci (changes requested):
Some changes need to be made to accommodate multiple chips.

mdlpstsci:
The utils.py module in drizzlepac has a function, _count_sci_extensions_.

drizzlepac/util.py
```diff
@@ -436,25 +436,31 @@ def updateNEXTENDKw(fobj):
     if 'nextend' in fobj[0].header:
         fobj[0].header['nextend'] = len(fobj) - 1
 
-def count_sci_extensions(filename):
+def count_sci_extensions(filename, return_ind=False):
     """ Return the number of SCI extensions and the EXTNAME from a input MEF file.
```
mdlpstsci, line 440: Add documentation that explains "if _return_ind=True_, only a list of the true "SCI" extension indices in the MEF will be returned."  

This is in case someone tries to use this function on files which have been manipulated.  The Primary Data Unit (PDU) of a FITS file can hold images (i.e., tables or such cannot be in the Primary), but this is not the convention used for HST data.

s-goldman: Added!

drizzlepac/haputils/analyze.py
```diff
@@ -425,13 +426,30 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
         # Determine if the image has one of these conditions.  The routine
         # will exit processing upon the first satisfied condition.
 
+        # Check if all science image arrays are filled with zero values
```
mdlpstsci, line 429: Change comment to be "...all science image arrays in the RAW file...".  SCI extensions in the processed data are float values which would require a different check against zero.

drizzlepac/haputils/analyze.py
```diff
@@ -425,13 +426,30 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
...
+        non_zero_data_in_array = False # start assuming data is zeros
+        science_ext_ind_array = count_sci_extensions(input_file, return_ind=True)
+        # make sure science extesion exists
+        if len(science_ext_ind_array)>0:
+            for sci_ext_ind in science_ext_ind_array:
+                science_data = getdata(input_file, sci_ext_ind)
+                # change flag if good data in any science extensnion array
+                if np.logical_not(np.all(science_data==0)):
```
mdlpstsci, line 437: np.all(science_data==0) will return only True or False here, so there is no need for np.logical_not

drizzlepac/haputils/analyze.py
```diff
@@ -425,13 +426,30 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
         # Determine if the image has one of these conditions.  The routine
         # will exit processing upon the first satisfied condition.
 
+        # Check if all science image arrays are filled with zero values
+        non_zero_data_in_array = False # start assuming data is zeros
+        science_ext_ind_array = count_sci_extensions(input_file, return_ind=True)
+        # make sure science extesion exists
```
mdlpstsci, line 432: Typo fix: extension

drizzlepac/haputils/analyze.py
```diff
@@ -425,13 +426,30 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
...
+        non_zero_data_in_array = False # start assuming data is zeros
+        science_ext_ind_array = count_sci_extensions(input_file, return_ind=True)
+        # make sure science extension exists
+        if len(science_ext_ind_array)>0:
+            for sci_ext_ind in science_ext_ind_array:
+                science_data = getdata(input_file, sci_ext_ind)
+                # change flag if good data in any science extensnion array
+                if not np.all(science_data==0):
+                    non_zero_data_in_array = True
+                    log.warning(
```
mdlpstsci, line 440: When this if block is True, there is no need for the warning.  Seems like there might be an else block missing for this inner if block.

s-goldman: I just saw that as well and corrected it.

drizzlepac/haputils/analyze.py
```diff
@@ -425,13 +426,32 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
...
         # will exit processing upon the first satisfied condition.
 
+        # Check if all science image arrays in the RAW file are filled with zero values
+        non_zero_data_in_array = False # start assuming data is zeros
+        science_ext_ind_array = count_sci_extensions(input_file, return_ind=True)
+        # make sure science extension exists
+        if len(science_ext_ind_array)>0:
+            for sci_ext_ind in science_ext_ind_array:
+                science_data = getdata(input_file, sci_ext_ind)
+                # change flag if good data in any science extensnion array
```
mdlpstsci, line 436: spelling: extension (Line 436)

s-goldman: Fixed. I clearly have an issue with typing the word extension.

drizzlepac/haputils/analyze.py
```diff
@@ -532,7 +552,10 @@ def analyze_data(input_file_list, log_level=logutil.logging.DEBUG, type=""):
             # processing should be allowed, but there may be some issue with the result (e.g.,
             # GYROS mode so some drift)
             generate_msg(input_file, msg_type, no_proc_key, no_proc_value)
-        
+        elif non_zero_data_in_array==False:
+            do_process=False
```
mdlpstsci, line 556: add here:
process_msg="SCI data all zeros"
so the message also is put into the output table.

s-goldman: Added.

mdlpstsci (approved):
All issues have been addressed.

PR #1905, opened 2024-10-28
HLA-1361: Pin github devdep tests to python 3.12

s-goldman:
Resolves [HLA-1361](https://jira.stsci.edu/browse/HLA-1361)

Currently the devedep [tests are failing on mac](https://github.com/spacetelescope/drizzlepac/actions/runs/11544991206/job/32131286092). The devdep tests trying to use the most up-to-date dependencies are also trying to test dependencies with python 3.13, even if they do not exist. This is resulting in dependency conflicts, particularly with setuptools. To overcome this, we will pin the dependency tests to python 3.12 until we are ready to test them with python 3.13.


mdlpstsci (approved):
Agreed!

PR #1907, opened 2024-11-04
HLA-1369: No viable data code 0 changed to code 65

s-goldman:
Resolves [HLA-1369](https://jira.stsci.edu/browse/HLA-1369)

Closes #

This PR addresses an issue where an incorrect error code is being returned in SVM processing. For the case of 

`There are no viable direct images in any Total Data Product for this visit`

The error code that was previous returned (0) has been changed to 65.

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -67,6 +67,7 @@ from drizzlepac.haputils import config_utils
 from drizzlepac.haputils import diagnostic_utils
 from drizzlepac.haputils import hla_flag_filter
 from drizzlepac.haputils import poller_utils
+from drizzlepac.haputils import analyze
 from drizzlepac.haputils import product
 from drizzlepac.haputils import processing_utils as proc_utils
 from drizzlepac.haputils import svm_quality_analysis as svm_qa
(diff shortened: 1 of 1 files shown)


drizzlepac/hapsequencer.py
```diff
@@ -579,7 +579,7 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
             log.warning("")
             log.warning("There are no viable direct images in any Total Data Product for this visit. No processing can be done.")
             log.warning("No SVM processing is done for the Grism/Prism data - no SVM output products are generated.")
-            sys.exit(0)
+            sys.exit(65)
```
mdlpstsci, line 582: If it can be avoided, "magic numbers" should not be used in the source code.  Use **analyze.Ret_code.NO_VIABLE_DATA.value**.

mdlpstsci (changes requested):
Use defined enumerated value for exit code.

drizzlepac/hapsequencer.py
s-goldman: fixed

mdlpstsci (approved):

PR #1908, opened 2024-11-04
HLA-1362: Modified the minimum RMS computation used as a discriminant for choos…

mdlpstsci:
Resolves [HLA-1362](https://jira.stsci.edu/browse/HLA-1362)

Closes #

Modified the minimum RMS computation used as a discriminant for choosing the
background computation algorithm for source detection.  The update is for the source code only as any detector-dependent configuration values have not yet been tuned.

Access to the updated source code is so statistical tests can be run for further evaluation of the change and data which can be used to tune the detector-dependent configuration values.

**Checklist for maintainers**
- [X] added relevant label(s)
- [X] ran regression tests, post a link to the Jenkins job below.
https://plwishmaster.stsci.edu:8081/job/RT/job/Drizzlepac-Developers-Pull-Requests/156/

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -343,17 +343,18 @@ class CatalogImage:
                 log.info("")
 
             # Compute a minimum rms value based upon information directly from the data
-            if self.keyword_dict["detector"].upper() != 'SBC':
-                minimum_rms = self.keyword_dict['atodgn'] * self.keyword_dict['readnse'] \
-                              * self.keyword_dict['ndrizim'] / self.keyword_dict['texpo_time']
+            if self.keyword_dict['detector'].upper() != 'SBC':
+                minimum_rms = np.sqrt(self.keyword_dict['numexp']) * self.keyword_dict['readnse'] / self.keyword_dict['texpo_time']
+            else:
+                minimum_rms = np.sqrt(np.clip(bkg_median * self.keyword_dict['texpo_time'], a_min=1.0, a_max=None)) / self.keyword_dict['texpo_time']
 
-                # Compare a minimum rms based upon input characteristics versus the one computed and use
-                # the larger of the two values.
-                if (bkg_rms < minimum_rms):
-                    bkg_rms = minimum_rms
-                    log.info(f"Mimimum RMS of input based upon the readnoise, gain, number of exposures, and total exposure time: {minimum_rms:.6f}")
-                    log.info(f"Sigma-clipped RMS has been updated - Background mean: {bkg_mean:.6f}  median: {bkg_median:.6f}  rms: {bkg_rms:.6f}")
-                    log.info("")
+            # Compare a minimum rms based upon input characteristics versus the rms computed and use
+            # the larger of the two values.
+            if (bkg_rms < minimum_rms):
+               bkg_rms = minimum_rms
+               log.info(f"Minimum RMS of input based upon the readnoise, number of exposures, and total exposure time: {minimum_rms:.6f}")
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -344,17 +344,18 @@ def compute_background(self, box_size, win_size,
...
-                # the larger of the two values.
-                if (bkg_rms < minimum_rms):
-                    bkg_rms = minimum_rms
-                    log.info(f"Mimimum RMS of input based upon the readnoise, gain, number of exposures, and total exposure time: {minimum_rms:.6f}")
-                    log.info(f"Sigma-clipped RMS has been updated - Background mean: {bkg_mean:.6f}  median: {bkg_median:.6f}  rms: {bkg_rms:.6f}")
-                    log.info("")
+            if self.keyword_dict['detector'].upper() != 'SBC':
+                minimum_rms = np.sqrt(self.keyword_dict['numexp']) * self.keyword_dict['readnse'] / self.keyword_dict['texpo_time']
+            else:
+                minimum_rms = np.sqrt(np.clip(bkg_median * self.keyword_dict['texpo_time'], a_min=1.0, a_max=None)) / self.keyword_dict['texpo_time']
```
rlwastro, line 350:
```suggestion
                minimum_rms = np.sqrt(
                    (bkg_median * self.keyword_dict['texpo_time']).clip(min=1.0)
                ) / self.keyword_dict['texpo_time']
```
I like using the `ndarray.clip()` method rather than the `np.clip` function.  But if you have some reason to prefer the `np.clip` function version, your code will also work.  In either case, you could omit the `max` value (or `a_max`, which seems the same as `max`), since it defaults to None.

rlwastro (approved):
This looks good to me.  Very clean implementation.  You can certainly ignore my suggestion if you want!

drizzlepac/haputils/catalog_utils.py
mdlpstsci: I have no preference for either function/method, but since the function is already coded I will continue to use it :).   I will note that one does have to specify both a_min and a_max (TypeError: clip() missing 1 required positional argument: 'a_max') as I discovered when I initially implemented the code!

rlwastro: OK, sounds OK to me.  (You can omit the max value with ndarray.clip.)  

I have one other suggestion.  I think the version of the rms that is used in the "excessive zero background" case should also have the minimum applied.  The code is a little simpler because the median value is known to be zero.  Here's a patch:

```patch
index e9ceb8fe..dbadf50f 100755
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -289,6 +289,12 @@ class CatalogImage:
 
                 is_zero_background_defined = True
                 log.info(f"Input image contains excessive zero values in the background. Median: {self.bkg_median:.6f} RMS: {self.bkg_rms_median:.6f}")
+                # Compute a minimum rms value based upon information directly from the data
+                minimum_rms = 1. / self.keyword_dict['texpo_time']
+                if np.nan_to_num(self.bkg_rms_median) < minimum_rms:
+                    self.bkg_rms_median = minimum_rms
+                    log.info(f"Minimum RMS of input based upon the total exposure time: {minimum_rms:.6f}")
+                    log.info(f"Median RMS has been updated - Median: {self.bkg_median:.6f} RMS: {self.bkg_rms_median:.6f}")
 
         # BACKGROUND COMPUTATION 2 (sigma_clipped_stats)
         # If the input data is not the unusual case of SBC "excessive zero background", compute
```

I included code that handles the issue you ran across where the computed rms is NaN.

PR #1909, opened 2024-11-05
Ignore nans in staticmask call to stsci.imagestats

s-goldman:
In investigating a build error ([HLA-1368](https://jira.stsci.edu/browse/HLA-1368)) we found an error where a nan in the science image array caused an error in a call to stsci.imagestats. The call to stsci.imagestats calculates a min and max (used for cuts), that results in nans for both. 

Here I am passing min and max values using np.nanmin and np.nanmax, which ignore nans in the calculation.


mdlpstsci (approved):
Changes look fine.

PR #1910, opened 2024-11-07
HLA-1370: Fixed a bug from a previous commit for computing statistics

mdlpstsci:
Resolves [HLA-1370](https://jira.stsci.edu/browse/HLA-1370)

Closes #

Fixed a bug associated with a previous commit when computing statistics on the illuminated portion only of the drizzled image.  The computation now uses the footprint_mask which is based on the DRZ CTX image which reports the number of images which contributed to the final pixel value.  If the CTX[i,j] > 0, then the pixel is illuminated.

**Checklist for maintainers**
- [X] added relevant label(s)


s-goldman (approved):
If the logic works correctly now, the code looks fine.

PR #1929, opened 2025-01-09
HLA-1402: Fix failing pytests

s-goldman:
Resolves [HLA-1402](https://jira.stsci.edu/browse/HLA-1402)

This PR addresses the failing pytests on github actions. 

test_align.py: 
- Due to recent changes to the required number of n_matches, some of the expected WCS solutions are no longer valid. One dataset "ib2u12" no longer achieves a GaiaDr2 solution, as a result this dataset has been removed (temporarily commented out) from the test's input list. 

test_asn_regress:
- The test is currently passing when triggered manually using linux or mac. A problem occurs, however, when the test is run using our new github actions tests. When the astrodrizzle call is run, a dependency call to `stsci.tools.asnutils.readASNTable` fails to read in the association (asn) file. While testing, I was able to confirm that the file is downloaded to the correct location, but it is apparently [empty](https://github.com/spacetelescope/RegressionTests/actions/runs/12680081371/job/35345371388#step:29:209): `ValueError: The truth value of an empty array is ambiguous.`

**Checklist for maintainers**
- [x] added relevant label(s)


mdlpstsci (approved):
Once we are _up and running_ on GA and not under pressure for a build, we should discuss which test to improve, add, and/or eliminate!

PR #1932, opened 2025-01-16
HLA-1405: Updated alignment parameters.

### LINKED ISSUE #1925: Update alignment parameters for pipeline and SVM processing
A fix for the alignment logic is requested for the next drizzlepac build.   These changes are important to avoid skewing the images by allowing for complex fit geometry when there only a few catalog matches.

At the moment we have the following parameters for all instruments/detectors. The code will attempt “rshift”, “rscale”, and “general” fit geometry with the number of required NMATCHES for pipeline products (mosaic_rel_geom) and SVM products (mosaic_fitgeom).   
```
"mosaic_relgeom_list": {"rshift":20, "rscale": 10, "general": 9}
"mosaic_fitgeom_list": {"rshift": 4, "rscale": 10, "general": 6},
```

This is not the desired logic which should apply a more complex fit geometry only with a large number of matches.  WFC3 and ACS have been digging into to accuracy our current distortion solutions and based on those results, we would like to change the parameters to the following:
```
"mosaic_relgeom_list": {"rscale": 12}
"mosaic_fitgeom_list": {"rscale": 12, "general": 25}
```

We find errors in the filter-dependent scale in the geometric distortion reference file, so we want to eliminate the 'rshift' option and allow the software to fit for scale terms when aligning images.  We increase the number of required matches from 10 to 12 to account for sigma-clipping.  To fit for skew errors in the distortion (e.g. general fit), we find that a large number of sources are required for a good fit.  

As Michele suggests in https://jira.stsci.edu/browse/HLA-1385, we need to discuss as a group and then ask Matt to run a large regression test so that the impact of this update can be assessed in an isolated manner.

s-goldman:
Resolves [HLA-1405](https://jira.stsci.edu/browse/HLA-1405)

Closes #1925 

This PR updates the pipeline and SVM alignment parameters to required the following number of (n)matches to the _rscale_ and _general_ fit geometries. 

```
"mosaic_fitgeom_list": {"rscale": 12, "general": 25}
"mosaic_relgeom_list": {"rscale": 12}
```

**Checklist for maintainers**
- [x] added relevant label(s)


mdlpstsci (approved):

mdlpstsci (changes requested):
While I find the current changes fine, the HLA-1405 ticket suggests an SVM test run via Matt as these changes will have a big impact on the results, and INS should have a chance to see the consequence of these changes.

Matt should be asked to run an SVM test using this branch.

PR #1939, opened 2025-01-24
HLA-1407: Implemented fixes to address uneven detection thresholds in HAP catalogs

mdlpstsci:
Resolves [HLA-407](https://jira.stsci.edu/browse/HLA-1407)

Closes #

Implemented fixes recommended by R.White to address uneven detection thresholds in the HAP catalogs due to bugs in the function, make_wht_masks, which intends to create weight masks covering the full drizzled output footprint. A somewhat related bug in the compute_threshold method associated with only the Segment catalog was also addressed. The "scale factor" which causes the RMS computation to be too small was deleted.  The RMS computation for the Point and Segment catalogs is now the same.

**Checklist for maintainers**
- [ ] added relevant label(s)


rlwastro (approved):
This all looks good to me!

PR #1953, opened 2025-02-20
HLA-1427: Improve algorithm for switching to crowded field kernel for HAP segment catalogs

mdlpstsci:
Resolves [HLA-1427](https://jira.stsci.edu/browse/HLA-1427)

Closes #

Added a third test using the size in pixels of the largest determined segment as a discriminant with regard to switching from a Gaussian to a Ricker-Wavelet kernel.  The kernel is used to convolve with the input image to increase the signal-to-noise ratio for the detection of faint sources.

**Checklist for maintainers**
- [X] updated relevant documentation
- [X] added relevant label(s)

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -2844,7 +2844,6 @@ class HAPSegmentCatalog(HAPCatalogBase):
         # Compute the biggest source identified 
         # The clip eliminates zero-divides when there are no good pixels in the image
         real_pixels = (np.isfinite(image_data) & (image_data != 0)).sum().clip(min=1)
-        biggest_source = segm_img.areas.max()/real_pixels
         biggest_source_pixels = segm_img.areas.max()
         biggest_source = biggest_source_pixels/real_pixels
 
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -2846,6 +2846,7 @@ class HAPSegmentCatalog(HAPCatalogBase):
         real_pixels = (np.isfinite(image_data) & (image_data != 0)).sum().clip(min=1)
         biggest_source = segm_img.areas.max()/real_pixels
         biggest_source_pixels = segm_img.areas.max()
+        biggest_source = biggest_source_pixels/real_pixels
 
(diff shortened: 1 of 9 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -2833,7 +2845,7 @@ def _evaluate_segmentation_image(self, segm_img, image_data, big_island_only=Fal
         # The clip eliminates zero-divides when there are no good pixels in the image
         real_pixels = (np.isfinite(image_data) & (image_data != 0)).sum().clip(min=1)
         biggest_source = segm_img.areas.max()/real_pixels
-        log.info("Biggest_source in pixels: %d", segm_img.areas.max())
+        biggest_source_pixels = segm_img.areas.max()
```
rlwastro, line 2848:
```suggestion
        biggest_source_pixels = segm_img.areas.max()
        biggest_source = biggest_source_pixels/real_pixels
```
It might be a little clearer to swap these two lines.  Then you can see that the same number is being used in the test.

rlwastro (approved):
This all looks good to me.  I suggested one minor change, but it affects only the readability of the code and not the functionality.

drizzlepac/haputils/catalog_utils.py
```diff
@@ -2846,6 +2846,7 @@ def _evaluate_segmentation_image(self, segm_img, image_data, big_island_only=Fal
         real_pixels = (np.isfinite(image_data) & (image_data != 0)).sum().clip(min=1)
         biggest_source = segm_img.areas.max()/real_pixels
```
rlwastro, line 2847: Need to delete the redundant definition of `biggest_source` on line 2847.  It is replaced by new line 2849.

mdlpstsci: This is what I get for just accepting the change via the GUI which I basically never do - except for this time!

rlwastro (approved):
All looks good now!

drizzlepac/haputils/catalog_utils.py
rlwastro: I'm sure I could have marked it up correctly in the GUI so it was OK to accept the change!  So it is my fault.

PR #1956, opened 2025-02-24
HLA-1419: Removal of warnings for photometry_utils.py log of zero of negative number

s-goldman:
Resolves [HLA-1419](https://jira.stsci.edu/browse/HLA-1419)

This PR addresses warnings that are printed too often. The warning shows up in the calculation of catalog magnitudes during the creation of the catalogs (SVMs). The calculation of the magnitudes uses the log of the fluxes. Sometimes the fluxes are negative (not real) or set to zero. The warning is from calculating the log10 of a negative number or zero. 

This solution just ignores these warnings. A situation may arise where bad values may creep into the catalog and be missed by dismissing these warnings. This error would be immediately identified as the flux values are also in the catalog. As a result, this should not cause concern. 

**Checklist for maintainers**
- [ ] added relevant label(s)


drizzlepac/haputils/photometry_tools.py
```diff
@@ -235,9 +237,13 @@ def convert_flux_to_abmag(in_flux, photflam, photplam):
...
-        stmag = -2.5 * np.log10(f_lambda) - 21.10
+        # ignores warnings from log10 of zeros and negative values
+        with np.errstate(divide='ignore', invalid='ignore'):
+            stmag = -2.5 * np.log10(f_lambda) - 21.10        
 
         # Convert STMAG to ABMAG
-        abmag = stmag - 5.0 * np.log10(photplam) + 18.6921
+        # ignores warnings from log10 of zeros and negative values
+        with np.errstate(divide='ignore', invalid='ignore'):
+            abmag = stmag - 5.0 * np.log10(photplam) + 18.6921
```
mdlpstsci, line 247: Lines 241 and 246 introduce benign changes which eliminate lots of messages in the logs without changing the current computations.  In this sense, the changes are fine.  However, the Jira ticket was meant to _investigate how these computations can be better handled_ starting from the routine which calls compute_flux_to_abmag.  Should the block of code starting on Line 231 be expanded to encompass handling the flux array?  If issues are handled in the initial block of code we never drop down into the log10 computations.

In the test dataset was photplam zero or negative?  It is a pivot wavelength, so I am curious.

drizzlepac/haputils/photometry_tools.py
```diff
@@ -150,7 +150,9 @@ def iraf_style_photometry(phot_apertures, bg_apertures, data, photflam, photplam
         if math.isclose(photflam, 0.0, abs_tol=constants.TOLERANCE):
             mag_err = mag
         else:
-            mag_err = 1.0857 * flux_error / flux
+            # ignores warnings from log10 of zeros and negative values
+            with np.errstate(divide='ignore', invalid='ignore'):
+                mag_err = 1.0857 * flux_error / flux
```
mdlpstsci, line 155: It seems like the flux here is the troublesome variable as there is no log computation.  What is the value of mag_err here after the divide by zero?

mdlpstsci (changes requested):
This ticket was meant to be an investigation to understand (beyond zero or negative values) how to handle better these issues.  This requires some tracing back in the code to the calling functions.  If this is expanding into too much of a research ticket, then it should be put back on the backlog.  I do not want to mask the issues just yet, as I think the issues should actually be handled better than they are now.  What I am suggesting is potentially a more invasive change.

PR #1957, opened 2025-02-25
HLA-1433: Remove parameter deprecated in Photutils 2.0

mdlpstsci:
Resolves [HLA-1433](https://jira.stsci.edu/browse/HLA-1433)

Closes #

Removed a parameter deprecated in Photutils 2.0, edge_method, from the instantiation of a Background2D.  The default for this parameter is now always equal to "pad" which was the setting in use in our code.

**Checklist for maintainers**
- [X] added relevant label(s)


s-goldman (approved):
Looks good to me.

PR #1958, opened 2025-02-25
HLA-1417: Validation check for WCS solutions from astrometry database

s-goldman:
Resolves [HLA-1417](https://jira.stsci.edu/browse/HLA-1417)

This PR adds a validation check to adopted WCS solution taken from the astrometry database. The check makes sure that the CRVAL1 and CRVAL2 values (reference pixel coordinates in degrees) have a separation less than a certain threshold; otherwise a warning is issued.

Our current working warning separation threshold is 1 degrees, but this should be refined. 

**Checklist for maintainers**
- [x] added relevant label(s)


s-goldman:
Code behavior verified using "fresh" datasets (jb2o04), altering the header target position (RA_TARG), and seeing the warning only for the altered dataset.

mdlpstsci (changes requested):
I suggest making the warning message slightly stronger.

"...degrees from target position!  The astrometry database solution is suspect."

mdlpstsci (approved):

PR #1959, opened 2025-02-26
HLA-1419: Removal of warnings for photometry_utils.py log of zero of negative number

s-goldman:
Resolves [HLA-1419](https://jira.stsci.edu/browse/HLA-1419)

This PR addresses warnings that are printed too often. The warning shows up in the calculation of catalog magnitudes during the creation of the catalogs (SVMs). The calculation of the magnitudes uses the log of the fluxes. Sometimes the fluxes are negative (not real) or set to zero. The warning is from calculating the log10 of a negative number or zero. 

This solution results in -9999.0's for catalog Mag and MagErr values when the corresponding fluxes are negative or zero. These were previously set to infinity (1.7976931348623157e+308). The negative or zero fluxes are left unchanged, as these are useful in stacking catalogs (see Jira ticket). 

This was tested using program _ie0o39_.

**Checklist for maintainers**
- [x] added relevant label(s)


mdlpstsci (changes requested):
If you are available, let us go over this ticket together next week on Monday or Tuesday (or some other time good for you).

mdlpstsci (approved):
I am approving this ticket as it stops the numerous error messages generated when there is a divide by zero or log10(negative number or zero), by essentially allowing the "bad" value to be set as NaNs.  The NaN values are cleaned up much farther downstream when the filter catalogs are written.  Having said this, this particular handling (by setting to NaN) should be revisited after the follow on computations done in the catalogs are cleaned up.

PR #1969, opened 2025-03-09
HLA-1448: Fix the crfactor designation for the WFPC2 detector to avoid failure

mdlpstsci:
Resolves [HLA-1448](https://jira.stsci.edu/browse/HLA-1448)

Closes #

Fixed the crfactor designation for the WFPC2 detector (PC) which caused the the computation for rejecting catalog creation based on expected cosmic ray detections to fail ONLY for WFPC2.  Also, updated the WFPC2 cr_residual factor from 0.0 to 0.05 as it had never be set correctly.  Created a PyTest for WFPC2 SVM processing.

**Checklist for maintainers**
- [X] updated or added relevant tests
- [X] added relevant label(s)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,12 @@ number of the code change for that issue.  These PRs can be viewed at:
 3.9.2 (unreleased)
 ==================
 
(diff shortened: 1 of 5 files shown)


s-goldman (approved):
Looks good to me.

PR #1977, opened 2025-03-13
Removing rogue backslashes that show up as runtime warnings

s-goldman:
I've tried to remove some of the rogue backslashes that raise warnings during pipeline processing.


mdlpstsci (changes requested):
I am actually just requesting some information.  I have not seen this as an issue myself in general.  The "\" backslash should be a valid character, so I am confused about the circumstances where you have seen this as an issue.  Do you have an output log file for me to see?  I am wondering if there is some configuration setup, in particular, where this is an issue.

PR #1987, opened 2025-03-24
Dropping support for python 3.10

s-goldman:
We are removing support for python 3.10. 

We have specified that drizzlepac now requires python >=3.11 in the pyproject.toml and this removes the last (CI) test that we do under python 3.10.


mdlpstsci (changes requested):
Please update the CHANGELOG.rst file.

mdlpstsci (approved):

PR #1992, opened 2025-03-31
Fix for I/O astrometry database bug

s-goldman:
This issue is resolved by using _contextlib_ instead of a standalone function for silencing the teal import line from stsci.skypac.


mdlpstsci (approved):
Indeed, this fixes the bug when trying to communicate to the AstrometryDB.

PR #2007, opened 2025-05-01
HLA-1484: Resolve issue of duplicate IDs in the Total Point Catalog product

mdlpstsci:
Resolves [HLA-1484](https://jira.stsci.edu/browse/HLA-1484)

Closes #

Resolved the issue of duplicate "ID"s in the rows of the Total Point catalog. For "point"
source identification, looping is done over a list of weight masks, tp_masks, when invoking the "finder" algorithms. These algorithms return tables with a unique "id" number for each row/source in the table. When tp_masks > 1, the returned tables are "vstacked" to generate a final table. Unfortunately, the "id" number was not updated to reflect the stacking of the tables which created rows with the same "id".  This error did not cause the code to fail, but it did generate a garbled table.

It is believed this bug was introduced in HLA-1407/PR#1939 which is an update for this latest build (v3.10x) when the computation of the tp_masks was fixed as previously only one mask was ever generated (or so it is believed).  In essence, this was a bug waiting to happen.

**Checklist for maintainers**
- [X] added relevant label(s)

### DIFF
--- a/drizzlepac/haputils/catalog_utils.py
+++ b/drizzlepac/haputils/catalog_utils.py
@@ -1080,6 +1080,7 @@ class HAPPointCatalog(HAPCatalogBase):
             log.info("{}".format("=" * 80))
 
             sources = None
+            last_used_id = 0
             for masknum, mask in enumerate(self.tp_masks):
                 # apply mask for each separate range of WHT values
                 region = image * mask['mask']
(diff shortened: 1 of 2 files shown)


drizzlepac/haputils/catalog_utils.py
```diff
@@ -1180,9 +1183,12 @@ def identify_sources(self, **pars):
                         reg_sources.remove_column("daofind_mag")
 
                     if sources is None:
+                        last_used_id = reg_sources['id'][-1]
                         sources = reg_sources
                     else:
+                        reg_sources['id'][:] +=  last_used_id
```
s-goldman, line 1189: In the future, I think you can use reg_sources['id'] instead of reg_sources['id'][:], but everything else looks fines to me. I was able to confirm that the code works using program "ictj64".

s-goldman (approved):

PR #2027, opened 2025-05-22
HLA-1503: Report correct sigma value for computing threshold for source detection

mdlpstsci:
Resolves [HLA-1503](https://jira.stsci.edu/browse/HLA-1503)

Closes #

Clarified the sigma value used to compute the threshold above which sources are detected for the segmentation catalog.  The sigma value is different when using the Gaussian or RickerWavelet smoothing kernel for Round 1 processing, as well as both being doubled for Round 2 processing if and only if switching from a sigma-clipped background to a Background2D. The value has been corrected in the output Segmentation
catalogs, and given greater visibility in the trailer log files.

**Checklist for maintainers**
- [X] added relevant label(s)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,12 @@ number of the code change for that issue.  These PRs can be viewed at:
 3.10.0 (21-May-2025)
 ====================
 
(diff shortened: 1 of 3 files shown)


s-goldman (approved):
approved.

PR #2028, opened 2025-05-22
HLA-1479: Overwrote the MVM "alignment" configuration files with the SMV files

mdlpstsci:
Resolves [HLA-1479](https://jira.stsci.edu/browse/HLA-1479)

Closes #

Overwrote only the MVM "alignment" configuration files with the files stored in the SVM subdirectories.  It is understood these files are probably obsolete.  The files were out-of-sync, and they previously had been synchronized.

**Checklist for maintainers**
- [X] added relevant label(s)


s-goldman (approved):
approved

PR #2033, opened 2025-06-04
Drizzlepac/HAP: Special case: Ignore the RMS comparison of Background2D and sigma-clipped algorithm

mdlpstsci:
Resolves [HLA-1509](https://jira.stsci.edu/browse/HLA-1509)

Closes #

Ignore the RMS comparison between a Background2D and the sigma-clipped algorithm when the background is being "forced" to be a "Background2D" as indicated by the bkg_skew_threshold=0.0 and the negative_percent=100.0. This is done for the special case of Round 2 for identification of sources for the HAPSegmentCatalog.

**Checklist for maintainers**
- [X] added relevant label(s)


mdlpstsci:
Log statement regarding "Round 2" is already in code.

PR #2044, opened 2025-07-01
Updated Readme badges and RTD author list

s-goldman:
This PR only updates the readme so that the badges looks more like those for JWST as well as adding to the author list of the readthedocs page.


nden (approved):
LGTM

PR #2047, opened 2025-07-17
HLA-1464: Missing IDCSCALE crashes processing

nden:
Resolves [HLA-1464](https://jira.stsci.edu/browse/HLA-1464)

Closes #

This PR addresses a crash in OPS for HRC data "j9pb20010" during aposteriori processing. The issue was tracked down to update of the WCS with a headerlet which is missing IDCSCALE.

**Checklist for maintainers**
- [ ] added relevant label(s)

Regression tests passing: https://github.com/spacetelescope/RegressionTests/actions/runs/16349387771/job/46200642685

### DIFF
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -1575,11 +1575,6 @@ def update_wcs_in_list(exp_list, logfile=None):
     print(msg)
     update_msg = msg
     primary_wcsnames = set([fits.getval(fname, 'wcsname', ext=('SCI',1)) for fname in exp_list])
-    fyq_ind = exp_list.index('j9pb20fyq_flt.fits')
-    idcscale = fits.getval(exp_list[fyq_ind], ext=1, keyword="IDCSCALE")
-    msg = "\nfyq idcscale is %s " % idcscale
-    print(msg)
-    update_msg = msg
     if len(primary_wcsnames) == 1:
         msg = "\nAll Primary WCS's confirmed as consistent as {}.".format(primary_wcsnames)
         print(msg)
--- a/drizzlepac/runastrodriz.py
+++ b/drizzlepac/runastrodriz.py
@@ -203,8 +203,32 @@ wcs_preference = ['IDC_?????????-FIT_REL_GAIA*3', 'IDC_?????????-FIT_IMG_GAIA*3'
 def process(inFile, force=False, newpath=None, num_cores=None, inmemory=True,
             headerlets=True, align_to_gaia=True, force_alignment=False,
             do_verify_guiding=False, debug=False, make_manifest=False):
-    """ Run astrodrizzle on input file/ASN table
-        using default values for astrodrizzle parameters.
+    """
+    Run astrodrizzle on input file/ASN table using default values for astrodrizzle parameters.
+
+    Parameters
+    ----------
+    inFile : str
+    force : bool
+        If True make a product even if verification fails.
+    newpath : str
+        If provided processing will be done in ``newpath`` directory.
+    num_cores : int (?)
+        Number of cores to use.
+    inmemory : bool
+    headerlets : bool
+        Whether to generate headerlets or not.
+    align_to_gaia : bool
+        Whether to align to an external catalog.
+    force_alignment : bool
+    do_verify_guiding : bool
+        Whether to check for guiding problems. If True and force=True,
+        a product will be generated even if issues with guiding are found.
+    debug : bool
+        Debug logging on/off. Additional debug logging is saved in a json file.
+    make_manifest : bool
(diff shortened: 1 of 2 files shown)


drizzlepac/runastrodriz.py
```diff
@@ -1593,10 +1653,14 @@ def update_wcs_in_list(exp_list, logfile=None):
                 print(msg)
                 update_msg += msg
                 update_active_wcs(filename, final_wcsname, logfile=logfile)
+                # msg = "filename %s , idcscale is %s" % (filename, idcscale)
+                # print(msg)
+                # update_msg = msg
```
mdlpstsci, line 1658: Did you intend to remove Lines 1656-1658?

drizzlepac/runastrodriz.py
```diff
@@ -2133,6 +2205,52 @@ def _analyze_exposure(filename):
...
+
+    if isinstance(filename, str):
+        hdul = fits.open(filename, mode='update')
+    else:
+        # Assume it's an HDUList object in "update" mode.
+        hdul = filename
+    num_sci = fileutil.countExtn(hdul)
+    # get IDCTAB name
+    itabroot = hdul[0].header['idctab'].split('$')[1].split('_')[0]
+    hdu_idscale = None
```
mdlpstsci, line 2228: It does not matter, but you initialize _hdu_idscale_ here, but then it is never used.  The variable f_hdu_idscale_ is actually is use.

mdlpstsci (changes requested):
Nothing of major consequence.  There are some questions/clean-up.

drizzlepac/runastrodriz.py
```diff
@@ -2133,6 +2201,52 @@ def _analyze_exposure(filename):
...
+        for extn in extvers:
+            if itabroot in hdul[('HDRLET', extn)].header['wcsname']:
+                _hdrlet = hdul[('HDRLET', extn)].headerlet
+                if 'idcscale' in _hdrlet[('SIPWCS', sciext + 1)].header:
+                    fhdu_idscale = _hdrlet[('SIPWCS', sciext + 1)].header['idcscale']
+                    break
+        if fhdu_idscale is None:
+            cd11 = hdul[('sci', sciext + 1)].header['CD1_1']
+            cd21 = hdul[('sci', sciext + 1)].header['CD2_1']
+            fhdu_idscale = round(np.sqrt(np.power(cd11, 2) + np.power(cd21, 2)) * 3600., 3)
```
mcara, line 2237: What is the reason for rounding  idcscale? I don't know if this image will be used in sky subtraction, but if it does - it looks like idcscale is used there to scale computed sky values. Also, wouldn't `np.sqrt(cd11**2 + cd21**2)` do the same? It seems that using `numpy`'s `power` will just add overhead.

drizzlepac/runastrodriz.py
```diff
@@ -1848,7 +1909,14 @@ def update_active_wcs(filename, wcsname, logfile=None):
...
         print(msg)
         update_msg += msg
-
+    try:
+        idcscale = fits.getval(filename, ext=1, keyword='IDCSCALE')
+    except KeyError:
+        msg += "Headerlet {} was missing IDCSCALE keyword".format(wcsname)
+        print(msg)
+        update_msg += msg
+        _update_idcscale(filename)
```
mcara, line 1918: Wouldn't you need to try getting `idcscale` again using the same code as on line 1912? (`idcscale = fits.getval(filename, ext=1, keyword='IDCSCALE')`)

mcara (approved):

drizzlepac/runastrodriz.py
nden: Agreed on using `np.power`. I basically copied the expression that was used in the apriori fix to avoid getting any differences in results. But I agree it can be cleaned up.

drizzlepac/runastrodriz.py
nden: I'm not sure I understand. Line 1913 checks the science header which was just updated by a headerlet. `_update_idcscale` iterates over all headerlets in the file to retrieve an `idcscale` for the same `idctab` and if it doesn't find any, computes one from the CD matrix.
What am I missing?

drizzlepac/runastrodriz.py
mcara: _If_ these files will ever be run through `astrodrizzle`, the `idcscale` value may be used in computations and accuracy matters. There is no reason to use `round` on this value anywhere. Maybe it shouldn't be done in this PR but these calls to `round` should be removed.

nden: FIled #2051

PR #2062, opened 2025-08-19
HLA-1521: Added documentation for regression tests. 

s-goldman:
Resolves [HLA-1521](https://jira.stsci.edu/browse/HLA-1521)

This PR adds documentation for our online Github actions testing. It also adds some details that are necessary for running regression tests locally on STScI machines. More specifically it describes that environmental variables need to be specified. 

I have not listed the locations of the environmental variables for security reasons.


s-goldman:
documentation move to internal innerspace page.

PR #2073, opened 2025-09-08
HLA-1536: Remove Temporary metawcs ecsv file

s-goldman:
Resolves [HLA-1536](https://jira.stsci.edu/browse/HLA-1536)

This PR addresses an issue where ecsv files are not being removed after processing. These should only be kept if diagnostic or debug mode is on. 

This was a consequence of an more substantial bug where a string within the output filename was not being updated before using it for naming. This resulted in reference catalog files with incorrect names for all metawcs catalog ecsv files. 

The output name is something like:

hst_9401_66_wfpc2_pc_u8f_metawcs_all_*reference_catalog_name*_ref_cat.ecsv

However since the *reference_catalog_name* wasn't being assigned until after the file was created, the initial reference_catalog_name string for the first GAIAeDR3 run was empty. The subsequent runs for GAIADR2 and 2MASS also had names that were one off from where they should have been (e.g. the GAIADR2 reference catalog files had 2MASS in the name of the file). 

The only file that was not being regularly deleted as it should have been was the file with an empty "reference_catalog_name" as it should not have existed. The issue should now be resolved. 

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: [Link](https://github.com/spacetelescope/RegressionTests/actions/runs/17563250507)


CHANGELOG.rst
```diff
@@ -20,6 +20,8 @@ number of the code change for that issue.  These PRs can be viewed at:
 3.xx.x (unreleased)
 ===================
 
+- Fixed metawcs ecsv filename issue and lingering files. [#2073]
```
nden, line 23: Is it more accurate to say `and removed lingering files`?

drizzlepac/haputils/product.py
```diff
@@ -261,7 +261,7 @@ def align_to_gaia(
                     refname = (
                         self.product_basename + "_" + catalog_item + "_ref_cat.ecsv"
                     )
-
+                    self.refname = refname
```
nden, line 264: Trying to understand the code: the comment above says `Override the default self.refname` (and I understand this is not a new comment). But it looks like that's the first assignment to a new attribute `self.refname`. Am I misreading the code?
If not, I think the attribute should be defined in the `__init__` with some default value and the reassigned here,

CHANGELOG.rst
s-goldman: fixed with [142af7b61fe98c6f0b0e9315d79672ba03001ddd](https://github.com/spacetelescope/drizzlepac/pull/2073/commits/142af7b61fe98c6f0b0e9315d79672ba03001ddd)

drizzlepac/haputils/product.py
```diff
@@ -261,7 +264,7 @@ def align_to_gaia(
                     refname = (
                         self.product_basename + "_" + catalog_item + "_ref_cat.ecsv"
                     )
```
nden, line 266: The parenthesis here don't do anythong. `refname` ends up being a string. If this is the intention (which makes sense) can you remove the parenthesis. It's a bit confusing seeing them.

PR #2077, opened 2025-09-30
HLA-1520: Remove Help Files (1 of 3)

### LINKED ISSUE #2038: Drizzlepac/HAP: Remove help function files
_Issue [HLA-1520](https://jira.stsci.edu/browse/HLA-1520) was created on JIRA by [Steve Goldman](https://jira.stsci.edu/secure/ViewProfile.jspa?name=sgoldman):_

The help functions associated with functions like astrodrizzle are obsolete. They were created to mimic IRAF help functions and are no longer useful. The ticket is to move the information in the help functions to either the docstrings of those functions/modules or to the relevant ReadTheDocs page.

s-goldman:
Resolves [HLA-1520](https://jira.stsci.edu/browse/HLA-1520)

Closes #2038 

This PR is one of three that transfer the ".help" files to the docstrings of the relevant functions. This behavior was a relic of IRAF and we are bringing the code closer to the common standard for python. The previous help files also avoided raising warnings and errors as they were not compiled as doc strings. We are using this PR to also remedy all of these issues. 

This PR is one of three aiming to accomplish this. It transfers the help files to docstrings for the following modules. 

- ablot
- adrizzle
- astrodrizzle
- createMedian

I've adopted the automodapi sphinx entension for each of these modules so that the functions with automatically generate docs are specified in the function `__all__`. I have included most of the functions for these modules, but I've decided to remove some internal functions for `adrizle.py` in the `__all__`, as these do not seem relevant for the typical user. 

I've also added a couple doc strings where they were previously missing. 

There is a new issue as a result of me adding the project_root directory to the path. When RTD imports drizzlepac it cannot import cdriz as it isn't compiled. It raises an error about circular imports. To overcome this, I've found a hacky way of mocking the cdriz module when compile the doc with RTD. Please let me know if you have a better solution or understand the initial issue a little better. 

I would also like to format (black) these files once the PR has been approved. 

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: [Link](https://github.com/spacetelescope/RegressionTests/actions/runs/18541833169)

### DIFF
--- a/drizzlepac/ablot.py
+++ b/drizzlepac/ablot.py
@@ -28,9 +28,9 @@ except ImportError:
 
 from . import __version__
 
-__all__ = ['blot', 'runBlot']
+__all__ = ["blot", "runBlot"]
 
-__taskname__ = 'ablot'
+__taskname__ = "ablot"
 STEP_NUM = 5
 PROCSTEPS_NAME = "Blot"
 
@@ -212,9 +212,9 @@ def blot(
 
     if input_dict is None:
         input_dict = {}
-    input_dict['data'] = data
-    input_dict['reference'] = reference
-    input_dict['outdata'] = outdata
+    input_dict["data"] = data
+    input_dict["reference"] = reference
+    input_dict["outdata"] = outdata
 
     # gets configObj defaults using EPAR/TEAL.
     #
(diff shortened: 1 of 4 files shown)


pllim (approved):
I think you have gotten the hang of things. A spot check of the doc looks good. There are some double blank lines in the docstring when one is enough, but they don't hurt so I am not going to nitpick over them. Good luck with the rest!

PR #2078, opened 2025-09-30
HLA-1520: Remove Help Files (2 of 3)

s-goldman:
A continuation of PR #2077. 

This PR moves the IRAF style help files to python docstrings for the following modules:

- refimagefindpars
- pixtopix
- mapreg
- imagefindpars
- drizCR

The documentation that was previously in imagefindpars and refimagefindpars (which are nearly identical) are now apart of the ImageCatalog Class docs in the catalogs.py module. 

Regression test: [Link](https://github.com/spacetelescope/RegressionTests/actions/runs/18565599621)


drizzlepac/mapreg.py
```diff
@@ -51,6 +62,226 @@ def wcs_pix2sky(self, *args, **kwargs):
...
+
+        * filename of a single image
+        * comma-separated list of filenames
+        * ``@-file`` filelist containing list of desired input region filenames
+
+        The ``@-file`` filelist needs to be provided as an ASCII text file
+        containing a list of filenames for all input region files with one
+        filename on each line of the file.
+
+        images : string or list of strings (Default = ``*.fits``)
```
pllim: Is the extra indent for `images` intentional?

drizzlepac/mapreg.py
```diff
@@ -51,6 +62,226 @@ def wcs_pix2sky(self, *args, **kwargs):
...
+    To do this we run:
+
+        >>> mapreg(input_reg = 'master.reg', images='img*.fits',
+        ...        img_wcs_ext='sci;dq,2;8', chip_reg='None,
+        ...        img1_chip_sci2.reg,None,None,None,None')
+
+    This will produce six region files in the ./regions subdirectory for *each*
+    input image::
+
+        ``img1_sci1_twreg.reg``,    ``img1_sci2_twreg.reg``,    ``img1_sci3_twreg.reg``,
```
pllim: You don't need double backticks if the text is already in a literal block (`::`). Looks weird when you use both, please see https://drizzlepac--2078.org.readthedocs.build/en/2078/api/drizzlepac.mapreg.MapReg.html#drizzlepac.mapreg.MapReg

drizzlepac/mapreg.py
```diff
@@ -125,6 +356,85 @@ def map_region_files(input_reg, images, img_wcs_ext='sci',
...
+    verbose : bool, optional
+        When ``True`` print additional progress information and associations
+        between extensions and chip regions. Default is ``True``.
+
+    Returns
+    -------
+    None
+        All results are persisted to ``outpath`` and optionally to the catalog
+        file.
+
```
pllim: No need to say "return None" but you will have to move the note about all results elsewhere, maybe the top.

```suggestion
```

s-goldman:
@pllim I've addressed your comments. I also removed an `interactive` argument for mapreg that never has or will be used. In some places it is `iteractive`.

drizzlepac/mapreg.py
```diff
@@ -49,18 +60,226 @@ def wcs_pix2sky(self, *args, **kwargs):
...
+        Specify whether or not to remove the regions in the transformed region files
+        that are outside the image array. With the 'fast' mehod only intersection
+        of the bounding boxes is being checked.
+
+        .. note::
+
+            The ``'precise'`` method is not
+            implemented in this release and, if specified, defaults to 'fast'.
+            The ``'precise'`` option is not available through the ``TEAL``
+            interface.
```
pllim: Same comment as `refimg`.

```suggestion
```

drizzlepac/mapreg.py
```diff
@@ -124,7 +343,78 @@ def _simple_parse_teal_extn(extnstr):
 def map_region_files(input_reg, images, img_wcs_ext='sci',
                      refimg=None, ref_wcs_ext='sci', chip_reg=None,
                      outpath='./regions', filter=None, catfname=None,
-                     iteractive=False, append=False, verbose=True):
```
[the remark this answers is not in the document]
mcara: I need to look into this.

> The argument does nothing and has a default of False. If a user did set it as True, then their code would probably raise an error.

This is a contradiction. If the argument "does nothing" then setting it to True should **not** result in an exception. If this argument is no longer needed, deprecate it (adding a changelog entry is not enough). Or ass `**kwargs` or something that is backwards compatible.

drizzlepac/mapreg.py
```diff
@@ -47,20 +59,226 @@ def wcs_pix2sky(self, *args, **kwargs):
...
+    catfname : string (Default = ``exclusions_cat.txt``)
+        The file name of the output exclusions catalog file to be created from the
+        supplied image and region file names. This file can be passed as an input
+        to TweakReg task. Verify that the created file is correct!
+
+    append : bool (Default = False)
+        Specify whether or not to append the transformed regions to the existing
+        region files with the same name.
+
+    interactive : bool (Default = False)
```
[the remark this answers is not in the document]
mcara: That is, I do not recommend https://github.com/spacetelescope/drizzlepac/pull/2078#discussion_r2467533931 change

CHANGELOG.rst
```diff
@@ -17,10 +17,11 @@ number of the code change for that issue.  These PRs can be viewed at:
 
     https://github.com/spacetelescope/drizzlepac/pulls
 
-3.xx.x (unreleased)
+3.11.0 (unreleased)
 ===================
 
-- Removed help files and moved text to docstrings. [#2077,#2078,#2079]
+- Removed help files and moved text to docstrings. Also deprecated the 
+  ``interactive`` argument in ``MapReg`` function. [#2077,#2078,#2079]
```
pllim:
```suggestion
  ``iteractive`` argument in ``MapReg`` function. [#2077,#2078,#2079]
```

PR #2080, opened 2025-09-30
Removing dev_utils

s-goldman:
These were used to validate the HAP products, but have not been used for some time. The files are no longer useful to host within each release of the code and will always be retained on github.


nden (approved):
LGTM

PR #2093, opened 2025-10-09
HLA-1551: Remove WFC3 Quadrant filter data from total SVM image and catalogs

### LINKED ISSUE #2067: Quadrant filters should not be used for HAP catalogs
_Issue [HLA-1551](https://jira.stsci.edu/browse/HLA-1551) was created on JIRA by [Rick White](https://jira.stsci.edu/secure/ViewProfile.jspa?name=rlw):_

The quadrant filters on WFC3/UVIS (names starting with `fq`) should not be used for HAP-SVM catalog generation.  Two changes are needed in the logic:

# Do not include `fq` images in the `total` image.
# Do not make catalogs for the `fq` images.

The `fq` filters have different bandpasses in different parts of the image.  That means that the photometry and detection thresholds need to be different in the different quadrants.  That would require major changes in the cataloging software, which are too complex to consider in the HAP processing.

The correct thing to do is simply not to use the `fq` filters for the catalogs.  That means they should also be omitted from the detection image. Currently the `fq` filters apparently are included in the `total` image, and catalogs are created.  This should be fixed.

An example is visit `hst_14393_18_wfc3_uvis_id1x18`, which includes two filters, `f606w` and `fq492n`.  In the current DrizzlePac 3.10.0 processing, that visit has both catalogs rejected with zero sources.  The rejection is very likely due to inclusion of the `fq492n` image in the `total` image.  The differences in the 4 quadrants is clearly visible in the total image.

!hst_14393_18_wfc3_uvis_total_id1x18.jpg|width=512!

s-goldman:
Resolves [HLA-1551](https://jira.stsci.edu/browse/HLA-1551)

Closes #2067 

This PR alters the already-in-place delete_ramp_exposures function and utilizes it to also remove exposures that use the WFC3/UVIS Quadrant filters. 

I've used the dataset "id1x18" for testing. While there still seems to be a backgrounding issue for this dataset, this fix has effectively removed the quad filter data from the total image and catalog. 

I have also change the name of the function, as well as some of the variable names, to better describe their new dual function. 

The key change is on [line 1422](https://github.com/spacetelescope/drizzlepac/pull/2093/files#diff-791f1f011a6f5e041fda7cada35e13d11d46dd86863c87f434599cad9c0b37b5R1422) of `hapsequencer.py`. 

Screenshot 2025-10-09 at 1 59 47 PM
Screenshot 2025-10-09 at 2 04 05 PM

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: [Link](https://github.com/spacetelescope/RegressionTests/actions/runs/19335150843)

### DIFF
--- a/drizzlepac/hapsequencer.py
+++ b/drizzlepac/hapsequencer.py
@@ -631,7 +631,6 @@ def run_hap_processing(input_filename, diagnostic_mode=False, input_custom_pars_
 
             # Need to delete the Ramp filter Exposure objects from the *Product lists as
             # these images should not be processed beyond the alignment to Gaia (run_align_to_gaia).
-            # import ipdb; ipdb.set_trace()
             _ = delete_ramp_and_quadrant_exposures(total_item.fdp_list, 'FILTER')
             ramp_product_list = delete_ramp_and_quadrant_exposures(total_item.edp_list, 'EXPOSURE')
             product_list += ramp_product_list
@@ -1388,8 +1387,9 @@ def delete_ramp_and_quadrant_exposures(obj_list, type_of_list):
     excluded_filenames_list = []
     while obj_list:
         obj = obj_list.pop()
-        # If the filter name does not contain 'fr' or 'fq', then it is okay to keep.
-        if (obj.filters.lower().find('fr') == -1) and (obj.filters.lower().find('fq') == -1):
+        # If the filter name does not contain 'fr' or 'fq', then keep it.
+        filt_lower = obj.filters.lower()
+        if not filt_lower.startswith(('fr', 'fq')):
             temp.append(obj)
         else:
             # Add the Ramp images to the manifest as well as the Ramp trailer filename here.


drizzlepac/hapsequencer.py
```diff
@@ -1377,32 +1379,33 @@ def delete_ramp_exposures(obj_list, type_of_list):
...
     """
     reported = False
     temp = []
-    ramp_filenames_list = []
+    excluded_filenames_list = []
     while obj_list:
         obj = obj_list.pop()
-        if obj.filters.lower().find('fr') == -1:
+        # If the filter name does not contain 'fr' or 'fq', then it is okay to keep.
+        if (obj.filters.lower().find('fr') == -1) and (obj.filters.lower().find('fq') == -1):
```
nden: I realize the previous code used `find` but I wonder if `startswith` is a better choice given that we expect the filter name to start with these characters. Then I don't know how HST filters are named and if this is relevant.

PR #2104, opened 2025-11-25
HLA-1562: Add bigdata mark to HAP/SVM tests

s-goldman:
Resolves [HLA-1562](https://jira.stsci.edu/browse/HLA-1562)

This PR addresses an ongoing issue with failing regression tests. It is suspected that the many calls to astroquery are spamming the system and causing intermittent failures. This is likely to be the case when updating PRs after a merge and many runners are calling astroquery at the same time, and for many datasets. 

To alleviate this, I have added the bigdata flag to the longer SVM regression tests. With this change, these tests are only run manually upon calls to [STScI Regression Tests](https://github.com/spacetelescope/RegressionTests/actions/workflows/drizzlepac.yml) when creating a PR. So now we have the short unit tests (around 2 minutes) and the slower regression tests (around an hour). 

I've also moved some of the common code in these files (they are all altered copies of each other) to a svm_test_utilities file.

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: https://github.com/spacetelescope/RegressionTests/actions/runs/19681231278

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@d9b81a07e789d1b1921ab5fe33de2ddcbbd02c3f  # v2.3.0
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish.yml@9f1fedda61294df4c004c05519a3fbf3b8e1f32f  # v2.3.1
     with:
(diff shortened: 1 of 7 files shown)


s-goldman:
@pllim I've addressed the comments that you left in your review. I now use the _jail fixture, I've updated the link in the docs, and most importantly, we now use artifactory instead of astroquery for getting the input data for these tests.

pllim:
`+249 −744` 👏 

Hard for me to grok the internals but I like it in principle. Did you try a real RT run and make sure it is green before merge?

PR #2122, opened 2026-01-13
Update supported Python versions in classifiers

pllim:
This PR addresses outdated Python versions in classifiers. 

**Checklist for maintainers**
- [ ] added relevant label(s)

Regression test:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,6 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
-    "Programming Language :: Python :: 3.14",
 ]
 dependencies = [
     "astropy>=6.1",


s-goldman:
I like this, although we don't test for python 3.14, so I'm not sure if we technically support it? Maybe we should start testing it first.

pllim:
py3 picks up the latest Python, so whether you know it or not, you already support it: https://github.com/spacetelescope/drizzlepac/actions/runs/20961153794/job/60238525962

s-goldman:
> py3 picks up the latest Python, so whether you know it or not, you already support it: https://github.com/spacetelescope/drizzlepac/actions/runs/20961153794/job/60238525962

True, but the github action in this repository skips two thirds of the regression tests including the big/long ones. On RT where we do run the full test suite we don't use a py3. Maybe we should.

https://github.com/spacetelescope/RegressionTests/actions/workflows/drizzlepac.yml

PR #2123, opened 2026-01-14
Warning about missing import in svm_quality_analysis

s-goldman:
There is a broken import in svm_quality_analysis that is raising an warning. The module hasn't been tested in a while and it's unclear if it still works as intended (see [HLA-330](https://jira.stsci.edu/browse/HLA-330?jql=text%20~%20%22svm_quality_analysis%22)). It is not implemented in the code unless specified with an environmental variable. 

I'm re-introducing a file which houses functions that are still imported by the svm_quality_analysis module. This fix allows the module to run without failure. 

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,8 +19,9 @@ classifiers = [
     "Operating System :: POSIX",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.12",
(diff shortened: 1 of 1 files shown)


s-goldman:
> Not sure of your change log policy and I am not familiar with HAP stuff but nothing shocking at a glance.

In the past the policy has been, if the users might benefit from knowing the change, then add a line in the changelog. 

Thanks!

PR #2125, opened 2026-01-14
HLA-1162: Replacing stregion with astropy.regions

### LINKED ISSUE #1742: Remove stregion from drizzlepac
_Issue [HLA-1212](https://jira.stsci.edu/browse/HLA-1212) was created on JIRA by [Steve Goldman](https://jira.stsci.edu/secure/ViewProfile.jspa?name=sgoldman):_

STREGION will not be supported in python 3.12. Menitons need to be removed or replaced with astropy regions.

s-goldman:
Resolves [HLA-1162](https://jira.stsci.edu/browse/HLA-1162)

Closes #1742 

This PR replaces the deprecated stregion package with astropy regions. The only major changes are to Mapreg.py, which has mostly been re-written. This code is used to generate region files used by tweakreg. Unit tests have been added to test the inputs to the function, and small changes have been made to catalogs.py.

A change in functionality is that I have removed/deprecated the `chip_reg` parameter. In the past you could specify with a pseudo mask-like string (None, None, region_file.reg, None, None)  which region files applies to which extension. I think since we have moved to a mainly python-based functionality, we can ask users to make multiple individual calls for each combination of image file and region. We can add examples on how to programmatically make multiple calls to MapReg to satisfy this need, it just seems like an unnecessary option that complicates the code.

I've also simplified the img_wcs_ext (extension specifier) to accept "SCI" or a list of the extension (e.g. [1,4]). I found the previous formatting that combined strings and extension pairs ('SCI;DQ,1;0') a little confusing. This may be another case where an example with multiple calls to MapReg can be used. 

I'm hoping that we can also ask the folks who make the drizzlepac notebooks to create a mapreg notebook, or add to a current notebook (e.g. [use_ds9_regions_in_tweakreg/use_ds9_regions_in_tweakreg](https://github.com/spacetelescope/hst_notebooks/blob/main/notebooks/DrizzlePac/use_ds9_regions_in_tweakreg/use_ds9_regions_in_tweakreg.ipynb)).

I would still like to add more tests including a test to verify that tweakreg accepts our output exclusion files as inputs, but I have verified that manually. That may be in another PR. 

The new module is as follows. There is the original MapReg function that validates the inputs and then passes the vetted parameters to a worker function, map_region_files. There are a couple functions called in map_region_files where we have to parse individual lines in ds9 region files. In these files we specify "exclusion" regions with a minus sign before the shape line. DS9 and astropy.regions don't recognize this format so we have to parse it manually. 

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: 
https://github.com/spacetelescope/RegressionTests/actions/runs/26114834107/job/76801265820


mcara:
This replacement seems to work well for polygon regions but not very well on rectangles and ellipsis. Here is an example that I simulated using data from [use_ds9_regions_in_tweakreg/use_ds9_regions_in_tweakreg](https://github.com/spacetelescope/hst_notebooks/blob/main/notebooks/DrizzlePac/use_ds9_regions_in_tweakreg/use_ds9_regions_in_tweakreg.ipynb)

```python
# step 1: create a resampled image:
astrodrizzle.AstroDrizzle('jcdua*flt.fits', final_rot=0)

# step 2: create a region file with multiple regions across the image in DS9.
#         Use various region types. Save regions in world coordinates.

# step 3: Use MapReg to map this region file to input images ('jcdua*flt.fits')
mapreg.MapReg(input_reg="drz_region.reg", images="jcdua3f4q_flc.fits", outpath='')

# step 4: Load jcdua3f4q_flc.fits into DS9 and load the region created in step 2
#         and the one created in step 3. Do regions match? They should!
```
Here is the image in DS9 (on left - resampled image; on right jcdua3f4q_flc.fits[1] with both regions):
Screenshot 2026-02-13 at 3 30 31 AM

Here is a zoomed version of the image on right:
Screenshot 2026-02-13 at 3 32 30 AM

Screenshot 2026-02-13 at 3 33 31 AM

The magenta regions are the ones created in `step 2` loaded in DS9 in the `step 4` above. The green regions are the ones created in `step 3` (using new - from this PR - mapreg). It is invisible, but below the magenta regions are hidden yellow regions created with the existing version of `MapReg` that uses `stregions`. Yellow regions are invisible because they precisely coincide with the original region (`step 2` in world coordinates) as rendered by DS9.

Something is not working with `astropy.regions` as well as it was working in `stregions` and the trick with `AuxWCS`. It is likely (but I am not sure) that `astropy.regions` either do not use distortions and, very likely, do not compute rotations with regard to the Y axis as `stregions` AND DS9 do but rather to the X-axis -- you can see this from the misalignment of the sides of rectangles. In theory - who cares what axis is used? But in practice we want the regions in which we find sources or do sky matching, etc. to be the regions that we **see** in DS9 and not something else.

CC: @larrybradley

mcara (approved):

s-goldman:
I've reached out to drizzling-inclined INS folks on the drizzle-updates slack channel, but no one responded. 

To my knowledge, this tool is still used by people, and there are jupyter notebooks that demonstrate how its use. 

I don't have strong feelings either way. If I was using the function, I would verify that the sources I was trying to omit were properly omitted. If the border is slightly off from what I selected, I would redraw the region. I think most, if not all users would use DS9, so having the DS9 region represent what is selected seems more important to me than the true sky coordinates. 

I appreciate the work that you've done Larry in trying to improve the way this function works, but I'm okay with a slight offset if you two are.

s-goldman:
@larrybradley @mcara I'm going to merge this tomorrow unless there are any objections. Thank you both for your insights on this.

mcara:
merge at will

larrybradley (approved):
LGTM, thanks!

PR #2152, opened 2026-05-20
HLA-1569: Remove lingering TEAL text, and deprecate TEAL functions and arguments

### LINKED ISSUE #2100: Drizzlepac/HAP: Remove all mentions of the TEAL GUI
_Issue [HLA-1569](https://jira.stsci.edu/browse/HLA-1569) was created on JIRA by [Steve Goldman](https://jira.stsci.edu/secure/ViewProfile.jspa?name=sgoldman):_

There are some lingering comments and documentation that references the use of the TEAL GUI. As the GUI was disable in the last release, the documentation and comments need to be updated. 

Wait until after the Jan 2026 CAL build.

s-goldman:
Resolves [HLA-1569](https://jira.stsci.edu/browse/HLA-1569)

Closes #2100

This PR removed lingering text from the TEAL interface. We still use teal to load our default configuration parameters, but the GUI has not been supported for the last two versions. 

This PR also adds deprecation warnings when using the `run` TEAL helper functions (e.g. `createMedian.run()` as opposed to `createMedian.median()`) that no one seems to call directly. 

It also adds deprecation warnings when using the `editpars` argument for the main drizzlepac functions. It is unlikely that people have been using this argument as it invokes the TEAL GUI. We are using a deprecation warning for this build as opposed to removing it because removing it would cause a call that includes `editpars=False` to return an error.

We will remove the teal helper functions and the editpar argument in the next version of the build. 

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: 
https://github.com/spacetelescope/RegressionTests/actions/runs/26170684336/job/76987182659

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,7 +23,7 @@ number of the code change for that issue.  These PRs can be viewed at:
 
 - Added code to handle deprecations in photutils v3.0.0. [#2089]
 
-- Deprecated the teal helper "run" functions as well as the editpars 
+- Deprecated the TEAL helper "run" functions as well as the editpars 
   parameters in base drizzlepac functions. [#2152]
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,9 @@ number of the code change for that issue.  These PRs can be viewed at:
 
 - Added code to handle deprecations in photutils v3.0.0. [#2089]
 
+- Deprecated the teal helper "run" functions as well as the editpars 
+  parameters in base drizzlepac functions. [#2152]
+
 
 3.11.0 (28-Apr-2026)
 ====================
--- a/drizzlepac/ablot.py
+++ b/drizzlepac/ablot.py
@@ -39,7 +39,7 @@ __taskname__ = "ablot"
 STEP_NUM = 5
 PROCSTEPS_NAME = "Blot"
 
-@deprecated_renamed_argument(new_name='editpars', old_name='', since='3.12.0')
+@deprecated_renamed_argument('editpars', None, '3.12.0')
 def blot(
     data,
     reference,
--- a/drizzlepac/ablot.py
+++ b/drizzlepac/ablot.py
@@ -229,7 +229,7 @@ def blot(
         run(configObj, wcsmap=wcsmap)
 
 
-@deprecated(since='3.12.0', warning_type=Warning)
+@deprecated(since='3.12.0')
 def run(configObj, wcsmap=None):
     """
     Run the blot task based on parameters provided interactively by the user.
--- a/drizzlepac/ablot.py
+++ b/drizzlepac/ablot.py
@@ -13,6 +13,7 @@ import os
 import logging
 import numpy as np
 from stsci.tools import fileutil
+from astropy.utils import deprecated
 from . import outputimage
 from . import wcs_functions
 from . import util
@@ -37,7 +38,7 @@ __taskname__ = "ablot"
 STEP_NUM = 5
 PROCSTEPS_NAME = "Blot"
 
-
+@deprecated(since='3.12.0', name='editpars', warning_type=Warning)
 def blot(
     data,
     reference,
@@ -217,10 +218,7 @@ def blot(
     input_dict["reference"] = reference
     input_dict["outdata"] = outdata
 
-    # gets configObj defaults using EPAR/TEAL.
-    #
-    # Also insure that the input_dict (user-specified values) are folded in
-    # with a fully populated configObj instance.
+    # get default configuration parameters using teal.load
     configObj = util.getDefaultConfigObj(__taskname__, configObj,
                                          input_dict, loadOnly=(not editpars))
     if configObj is None:
@@ -230,6 +228,7 @@ def blot(
         run(configObj, wcsmap=wcsmap)
 
 
+@deprecated(since='3.12.0', warning_type=Warning)
 def run(configObj, wcsmap=None):
     """
     Run the blot task based on parameters provided interactively by the user.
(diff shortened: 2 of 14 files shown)


s-goldman:
I've got the thumbs up from the drizzling folks in INS (https://stsci.slack.com/archives/C04SERR4TML/p1779219394588329)

drizzlepac/ablot.py
```diff
@@ -37,7 +38,7 @@
 STEP_NUM = 5
 PROCSTEPS_NAME = "Blot"
 
-
+@deprecated(since='3.12.0', name='editpars', warning_type=Warning)
 def blot(
```
pllim, line 43: I think you want to deprecate `editpars` and not the entire function `blot`? If so, I think you need to use this instead:

https://docs.astropy.org/en/stable/api/astropy.utils.decorators.deprecated_renamed_argument.html

drizzlepac/ablot.py
```diff
@@ -230,6 +228,7 @@ def blot(
         run(configObj, wcsmap=wcsmap)
 
 
+@deprecated(since='3.12.0', warning_type=Warning)
```
pllim: Why override the warning type?

drizzlepac/adrizzle.py
```diff
@@ -99,7 +101,7 @@ def drizzle(input, outdata, wcsmap=None, editpars=False, configObj=None, **input
     wcsmap : callable, optional
         Mapping factory that converts between input and output WCS frames.
         Defaults to ``None`` which triggers the package's standard mapping.
-    editpars : bool, optional
+    editpars : bool, (Default = False; deprecated)
```
pllim: It is clearer to use this Sphinx directive:

https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-version-deprecated

pllim (commented):
I left some comments. I support this PR in principle. I cannot wait for the actual code removal in the future. Yeehaw.

CHANGELOG.rst
```diff
@@ -23,6 +23,9 @@ number of the code change for that issue.  These PRs can be viewed at:
 
 - Added code to handle deprecations in photutils v3.0.0. [#2089]
 
+- Deprecated the teal helper "run" functions as well as the editpars
```
pllim:
```suggestion
- Deprecated the TEAL helper "run" functions as well as the editpars 
```

drizzlepac/ablot.py
s-goldman: Agreed, and added!

drizzlepac/ablot.py
s-goldman: I misunderstood the functionality, and I've updated the code without the explicit warning type.

drizzlepac/adrizzle.py
s-goldman: agreed and added to PR

drizzlepac/astrodrizzle.py
```diff
@@ -94,25 +97,24 @@ def AstroDrizzle(
...
-        This button will immediately update the parameter values in the ``TEAL``
-        GUI based on those provided by the ``MDRIZTAB`` reference table referenced
-        in the first input image. This requires that the ``MDRIZTAB`` reference
-        file be available locally.
+        This button will immediately update the parameter values provided by 
+        the ``MDRIZTAB`` reference table referenced in the first input image. 
+        This requires that the ``MDRIZTAB`` reference file be available locally.
 
-    editpars : bool (Default = False)
+    editpars : bool, optional
```
s-goldman, line 104: Some descriptions say `editpars : bool, (Default = False)` and others say `editpars : bool, optional`. I've elected to use the latter description for all of them.

drizzlepac/ablot.py
```diff
@@ -37,7 +39,7 @@
 STEP_NUM = 5
 PROCSTEPS_NAME = "Blot"
 
-
+@deprecated_renamed_argument(new_name='editpars', old_name='', since='3.12.0')
```
pllim: Same comment applies to all similar usage. See https://github.com/astropy/astropy/pull/16343 for example on how it was done.

```suggestion
@deprecated_renamed_argument('editpars', None, since='3.12.0')
```

drizzlepac/ablot.py
```diff
@@ -230,6 +229,7 @@ def blot(
         run(configObj, wcsmap=wcsmap)
 
 
+@deprecated(since='3.12.0', warning_type=Warning)
```
pllim: You replied you removed the warning_type override but I don't see it in the diff here.
```suggestion
@deprecated(since='3.12.0')
```

s-goldman: good catch, I'm not sure how I missed that one.

drizzlepac/ablot.py
s-goldman: done. thanks!

drizzlepac/buildmask.py
```diff
@@ -54,10 +54,7 @@
 def run(configObj=None, input_dict={}, loadOnly=False):
```
pllim, line 55: Is this one not deprecated?

drizzlepac/astrodrizzle.py
```diff
@@ -1111,8 +1104,8 @@ def AstroDrizzle(
     Examples
     ---------
 
-    The ``AstroDrizzle`` task can be run from either the ``TEAL`` GUI or from Python. These examples illustrate the various
-    syntax options available.
+    The ``AstroDrizzle`` command can be run from the command line or from Python.
```
pllim: Is there really a command line option without PyRAF or TEAL?

```suggestion
    The ``AstroDrizzle`` class can be run from Python. 
```

drizzlepac/buildmask.py
s-goldman: I'm not sure how much use that file gets, but you're right, we can deprecate it.

pllim (approved):
Good luck! Hopefully your users pay attention to warnings and don't get surprised when these are actually removed in the future.

PR #2159, opened 2026-06-23
EXP: Test against astropy PR 19971

pllim:
This PR is to test against astropy/astropy#19971 (Deprecate most members of the astropy.wcs C API) by Tom R. Do not merge!

Regression tests:
* https://github.com/spacetelescope/RegressionTests/actions/runs/28032450491

PR #2161, opened 2026-06-23
EXP: Test against astropy PR 19971 (first commit only)

pllim:
This PR is to test against https://github.com/astropy/astropy/pull/19971 (Deprecate most members of the astropy.wcs C API) by Tom R. Do not merge!

Unlike https://github.com/spacetelescope/drizzlepac/pull/2159 , this one only grabs the first commit in the PR as requested by Tom R in https://github.com/astropy/astropy/pull/19971#issuecomment-4780980732

Regression test: https://github.com/spacetelescope/RegressionTests/actions/runs/28038929309


pllim:
Given https://github.com/spacetelescope/drizzlepac/pull/2162 that also fails using astropy main, I don't think we need this one open anymore.

PR #2163, opened 2026-06-23
[SCSB-237] build Stable ABI3 wheels with `Py_LIMITED_API`

zacharyburnett:
Resolves [SCSB-237](https://jira.stsci.edu/browse/SCSB-237)

building against the Limited API means we can build just ONE wheel (in this case for Python 3.11) and it should work on all future Python versions: https://docs.python.org/3/c-api/stable.html#limited-c-api

status of this PR: **needs changes to C code to align with Limited API / Stable ABI3**

**Checklist for maintainers**
- [ ] added relevant label(s)

Regression test:


s-goldman:
@zacharyburnett This one still in the works?

zacharyburnett:
> @zacharyburnett This one still in the works?

yeah I was hoping it would just work 😵‍💫 but this will probably need some work on the C side to make sure it ony calls the Limited Stable API (if we can do that)

zacharyburnett:
no worries if not, this would be a Nice to Have as far as building goes but if not it's fine

PR #2162, opened 2026-06-23
EXP: Test against astropy main

pllim:
This is for comparison against #2159 and #2161 for https://github.com/astropy/astropy/pull/19971

Regression test: https://github.com/spacetelescope/RegressionTests/actions/runs/28379920748

PR #2164, opened 2026-06-23
Make test more robust to avoid failures with astropy main

astrofrog:
Match input and output grid parity in the kernel tests so point sources fall on output pixel centers rather than on output pixel boundaries.

This is just an idea for a solution, but it should make the test more robust.

**Checklist for maintainers**
- [x] added relevant label(s)


astrofrog:
I'm going to try and make the changes more minimal, we shouldn't have to change the ground truth for all kernels

mcara:
I was wondering whether adding an offset to `crpix` in https://github.com/spacetelescope/drizzlepac/blob/52d25476d68bf202dbb901e08a28c64f442c5bea/tests/drizzle/cdriz_setup.py#L10
could be enough. Maybe add an `offset` parameter to this function and then setting `offset` to, let's say 0.1234567 (for the point kernel only).

I have no idea whether this is going to work. Just thinking out loud. It's too late here (or too early).

mcara:
Hm, maybe what I proposed is not going to work since the final mapping is essentially an identity transformation. Will look into this more tomorrow.

astrofrog:
@mcara - sounds good, just FYI I have amended this to make the patch as minimal as possible

pllim:
I cherry picked this PR commit into https://github.com/spacetelescope/drizzlepac/pull/2159 . Let's see... 🤞

pllim:
Regression tests (internal access only):

* https://github.com/spacetelescope/RegressionTests/actions/runs/28111551835 (astropy stable)
* https://github.com/spacetelescope/RegressionTests/actions/runs/28111641657 (astropy dev)

pllim:
Plot twists:

* This does not seem to address the other failure that only pops up in OSX (https://github.com/spacetelescope/drizzlepac/pull/2162)
* Mihai opened a competing PR at https://github.com/spacetelescope/drizzlepac/pull/2165

pllim:
Regression testing results:

* All passed with astropy stable
* 17 failures with astropy dev . Since we only store one set of truth files, we will either have to let them fail for a while for this job, or adjust the tests so they pass for both astropy stable and dev. Internal log at https://github.com/spacetelescope/RegressionTests/actions/runs/28111641657/job/83240348287

mcara (approved):
LGTM. This is one of solutions.

mcara:
I will try to reconcile this with my PR later - my PR solves the problem in an equivalent way but it also simplifies the code for unit test functions.

PR #2165, opened 2026-06-24
Make cdriz kernel tests more stable to rounding errors

mcara:
~Alternative to #2164~
Minor improvement to test code. #2164 had a valid fix.

**Checklist for maintainers**
- [x] added entry in `CHANGELOG.rst` within the relevant release section
- [x] updated or added relevant tests
- [x] updated relevant documentation
- [ ] added relevant label(s)

Regression test:

PR #2166, opened 2026-06-24
EXP: Test against astropy main + PR 2165

pllim:
Like #2162 but with #2165


mcara:
OK, let's abandon #2165

PR #2184, opened 2026-08-31
HLA-1617: Reduce number unhelpful warnings

### LINKED ISSUE #2178: Drizzlepac/HAP: Reduce number of unhelpful warnings
_Issue [HLA-1617](https://jira.stsci.edu/browse/HLA-1617) was created on JIRA by [Steve Goldman](https://jira.stsci.edu/secure/ViewProfile.jspa?name=sgoldman):_

In some runs there are way to many warnings that look like:

`WARNING src=drizzlepac.haputils.astrometric_utils- The PHOT_RESULTS table has no good rows. Trying again.`

It was be more useful if this only happened once.

s-goldman:
Resolves [HLA-1617](https://jira.stsci.edu/browse/HLA-1617)

Closes #2178 

This PR silences some very common, yet unhelpful warnings that appear in logs and printed outputs for runastrodriz. I've changed the log level for some and added a logging filter for two other astropy logged messages. 

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test: 
https://github.com/spacetelescope/RegressionTests/actions/runs/34877831938

### DIFF
--- a/drizzlepac/haputils/astrometric_utils.py
+++ b/drizzlepac/haputils/astrometric_utils.py
@@ -887,7 +887,7 @@ def build_auto_kernel(imgarr, whtarr, fwhm=3.0, threshold=None, source_box=7,
     return (kernel, kernel_psf), kernel_fwhm
 
 
-def find_fwhm(psf, default_fwhm, log_level=logutil.logging.INFO):
+def find_fwhm(psf, default_fwhm):
     """Determine FWHM for auto-kernel PSF
 
     This function iteratively fits a Gaussian model to the extracted PSF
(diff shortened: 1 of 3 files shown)


drizzlepac/haputils/astrometric_utils.py
```diff
@@ -887,7 +887,7 @@ def build_auto_kernel(imgarr, whtarr, fwhm=3.0, threshold=None, source_box=7,
     return (kernel, kernel_psf), kernel_fwhm
 
 
-def find_fwhm(psf, default_fwhm, log_level=logutil.logging.INFO):
```
s-goldman, line 890: The code doesn't currently use log_level, so I'm removing this.

pllim: Is this considered "public API"? If not, this is fine; otherwise might need to be deprecated first.

Regardless, does not hurt to also mention this change in the change log?

drizzlepac/util.py
```diff
@@ -187,6 +187,22 @@ def end_logging(filename=None):
     root_logger = logging.getLogger()
     root_logger.removeHandler(_log_file_handler)
 
+
+class SuppressNoisyAstropyWCS(logging.Filter):
```
pllim, line 191: This is two-edge sword though. If the data is actually bad and those warnings can help debug, now you won't see it. 🤷‍♀️

drizzlepac/haputils/astrometric_utils.py
s-goldman: I guess technically it's public API, but log_level was never listed as an argument in the docstring. I don't think anyone would be using this functions as it just calls photutils, but i'll add a deprecation and a changelog entry just to be safe.

s-goldman: fixed in [cb715bf](https://github.com/spacetelescope/drizzlepac/pull/2184/commits/cb715bf4ec433a40cba381a5fc81e7c23e792af9)

drizzlepac/util.py
s-goldman: Since we get these for every run of every dataset, I don't think we would notice anyway 🫤

drizzlepac/haputils/astrometric_utils.py
```diff
@@ -887,6 +887,7 @@ def build_auto_kernel(imgarr, whtarr, fwhm=3.0, threshold=None, source_box=7,
     return (kernel, kernel_psf), kernel_fwhm
 
 
+@deprecated_renamed_argument('log_level', since='3.12.0')
```
pllim:
```suggestion
@deprecated_renamed_argument('log_level', None, since='3.12.0')
```

drizzlepac/util.py
pllim: I'll defer to your expertise here. Marking as resolved.

s-goldman:
RT tests pass, the local ones can be ignored for now.

pllim (approved):
:shipit:

PR #2188, opened 2026-09-16
Change build status badge in README

s-goldman:
Switches build status badge to RT. Local actions have been disabled.


pllim:
Oh, really? Not even some smoke tests that don't need Artifactory?

What about "check-style"?

The branch protection rules need to be updated also, it seems. And would be nice to update PR template to remind people to run RT.

Is RTD failure related?

PR #2189, opened 2026-09-17
HLA-1188: Deprecate match_2dhist_fit in align_utils.py

s-goldman:
Resolves [HLA-1188](https://jira.stsci.edu/browse/HLA-1188)

This PR deprecates the the match_2dhist_fit function as it is identical to match_default_fit. The two functions exist for historical reasons. The relevant argument is use2Dhist which fits the initial guess of the offset between the catalog and image. If the argument use2dhist is set to False, the code will use an input x and y offset values (sert to zero by default) instead of fitting them. 

As the name of match_2dhist_fit could be misleading (as use2dhist could be set to false) I have elected to deprecate that function instead of match_default_fit. 

Warnings will show up in the HAP pipeline until the next Calibration build. The function will be removed in a subsequent release. 

**Checklist for maintainers**
- [x] added relevant label(s)

Regression test:


drizzlepac/haputils/align_utils.py
```diff
@@ -1014,6 +1014,10 @@ def match_default_fit(imglist, reference_catalog, **fit_pars):
 
     log.info("{} (match_default_fit) Cross matching and fitting "
              "{}".format("-" * 20, "-" * 27))
+    if fit_pars['use2dhist']==True:
```
pllim:
```suggestion
    if fit_pars['use2dhist'] is True:
```

drizzlepac/haputils/align_utils.py
```diff
@@ -1032,68 +1036,14 @@ def match_default_fit(imglist, reference_catalog, **fit_pars):
...
-    del matched_cat
-
-    return imglist
+    """Deprecated alias for match_default_fit."""
+    warnings.warn(
+        "match_2dhist_fit is deprecated and will be removed in version 3.13.0; use match_default_fit instead.",
+        DeprecationWarning,
+        stacklevel=2,
+    )
+    return match_default_fit(imglist, reference_catalog, **fit_pars)
```
pllim: I don't understand. If you already plan to get rid of it, why even bother refactoring it? Also, why not use `astropy.utils.decorators.deprecated` decorator?

pllim: There is also https://docs.python.org/3/library/warnings.html#warnings.deprecated but you cannot use it until Python 3.13

